PR #11, opened 2017-02-01
Excalibur prototype demonstrator

timcnicholls:
This PR merges in several threads of development in both frameReceiver and fileWriter that were successfully demonstrated with the EXCALIBUR 1M system in Jan 17. This includes readout by dual and single nodes with parallel FR/FW chains running.

This is intended to provide a clean state reflecting latest work on master so that the necessary repo reorganisation can take place prior to future development.


test/unittest/FrameReceiverConfigUnitTest.cpp
```diff
@@ -55,13 +55,13 @@ BOOST_AUTO_TEST_CASE( ValidSensorNameToTypeMapping )
     FrameReceiver::FrameReceiverConfig theConfig;
     std::string p2mName         = "percival2m";
     std::string p13mName        = "percival13m";
-    std::string excalibur3mName = "excalibur3m";
+    std::string excalibur3mName = "excalibur";
```
ulrikpedersen: Should this not just be "excaliburSensorName" instead of "excalibur3mName"? I think elsewhere you have gone with just one type of Excalibur sensor, presumably because the framereceiver/filewriter only ever see one Excalibur module which is always the same(?)

include/FrameReceiverDefaults.h
```diff
@@ -22,7 +22,8 @@ namespace FrameReceiver
 			SensorTypePercivalEmulator,
 			SensorTypePercival2M,
 			SensorTypePercival13M,
-			SensorTypeExcalibur3M,
+			SensorTypeExcalibur1M,
```
GDYendell: Is the 1M supposed to be added to the sensor_name_map?

src/FrameReceiverApp.cpp
```diff
@@ -336,8 +360,14 @@ void FrameReceiverApp::initialise_ipc_channels(void)
...
     // Bind the frame ready and release channels
-    frame_ready_channel_.bind(config_.frame_ready_endpoint_);
-    frame_release_channel_.bind(config_.frame_release_endpoint_);
+    try {
+    	frame_ready_channel_.bind(config_.frame_ready_endpoint_);
+    	frame_release_channel_.bind(config_.frame_release_endpoint_);
+    }
+    catch (std::exception& e)
+    {
+    	LOG4CXX_ERROR(logger_, "Got exception biding channels: " << e.what());
```
GDYendell: Typo here.

test/unittest/FrameReceiverConfigUnitTest.cpp
timcnicholls: It should. I have already dropped the 3M in the core code as there will only be one decoder with configuration parameters. This is a hangover in the test cases only, but will go away once the config model changes to ClassLoader + config over control channel.

include/FrameReceiverDefaults.h
timcnicholls: Nope! I'll remove it, thanks!

src/FrameReceiverApp.cpp
timcnicholls: Will fix, thanks!

ulrikpedersen:
| Do we need to update write_excalibur_buffer in this PR or can that be done another time?

I think that's up to the two of you...

timcnicholls:
> | Do we need to update write_excalibur_buffer in this PR or can that be done another time?

> I think that's up to the two of you... 

I'm happy to merge the PR without it, can be done later at  @GDYendell's convenience, but am also happy to hang on if he has time to fix it.

timcnicholls:
Going to postpone write_excalibur_buffer work for now and merge PR. @GDYendell is opening an issue to track that required change.

PR #13, opened 2017-02-28
Add option to shutdown after given number of datasets received

GDYendell:
So I have now got this to work properly, on the surface at least, without needing a sleep. I still need to add the parsing of the command line args so it doesn't need the file_writer_client.

I have requested a review from you guys just as the easiest way to show it to you; don't feel you need to give a thorough review. As a summary:

- The FileWriter now registers a callback to the FileWriterController to pass any frames it receives
- The FWC callback function simply counts the number of frames it gets and when it receives the number of frames you define, it triggers the exit condition, which allows waitForShutdown to continue and shutdown everything.
- The FWC then stops and erases all the plugins, stops it's own worker thread and IPC reactor, and then closes the control and FrameReceiver interfaces

This seems to allow it to shutdown cleanly, but it's entirely possible it's not the correct way to do it.


tools/filewriter/DataBlockPool.cpp
```diff
@@ -262,4 +262,16 @@ namespace filewriter
     return memoryAllocated_;
   }
 
+  /**
+   * Delete DataBlockPool instances stored in static class attribute instanceMap_
+   */
+  void DataBlockPool::tearDownClass()
```
ulrikpedersen: As a general C++ view (I don't fully understand where it fits in the application): shouldn't this be a destructor instead?

tools/filewriter/FileWriterController.cpp
```diff
@@ -340,6 +377,29 @@ namespace filewriter
...
+    LOG4CXX_DEBUG(logger_, "Stopping FileWriterController worker thread and IPCReactor.\n")
+    stop();
+    reactor_->stop();
+    
+    // Close control IPC channel
+    closeControlInterface();
+    // Close FrameReceiver interface IPC channels
+    closeFrameReceiverInterface();
+    
+    // Destroy any allocated DataBlocks
```
ulrikpedersen: The comment even says "Destroy"... See my comment above about tearDownClass() :-)

tools/filewriter/FileWriterController.cpp
```diff
@@ -384,6 +444,39 @@ namespace filewriter
 
   }
 
+  /** Close the frame receiver interface.
+   *
+   * This method creates new SharedMemoryController and SharedMemoryParser objects,
```
ulrikpedersen: Comment wrong. Presumably a copy'n paste error.

tools/filewriter/FileWriterController.cpp
```diff
@@ -409,6 +502,23 @@ namespace filewriter
     reactor_->register_channel(ctrlChannel_, boost::bind(&FileWriterController::handleCtrlChannel, this));
   }
 
+  /** Close the control interface.
+   *
+   */
+  void FileWriterController::closeControlInterface()
```
ulrikpedersen: This function also sound a bit like a destructor: closing and releasing a shared resource.

tools/filewriter/SharedMemoryController.cpp
```diff
@@ -69,7 +69,14 @@ namespace filewriter
    */
   SharedMemoryController::~SharedMemoryController()
```
ulrikpedersen: Great this **is** actually a destructor! :+1: 

Make sure that none of the methods called in here can throw any exceptions - that's a big issue with destructors.

tools/filewriter/DataBlockPool.cpp
GDYendell: DataBlockPool has pool of allocated memory as a Class attribute, so each instance doesn't have to allocate it's own memory.

ulrikpedersen: oh ok. Who (which class) is responsible for allocating the memory in the first place? You'll want allocation and deletion of resources to happen in the same class.

Again: I haven't reviewed the code in full so I don't know the details of what is already there...

ajgdls: Hi actually DataBlockPool is a static class singleton.  Objects are allocated when they are first called by index (std::string name) through a static method and then persist so that if they are called again memory does not need to be re-allocated.  At least that is how I remember it :)  Perhaps this pattern is not appropriate here and the pool should itself be just an object.

ulrikpedersen: It probably is appropriate. If the DataBlockPool is a singleton it should delete the allocated memory objects - but only when the last instance of it is deleted.

GDYendell:
@ulrikpedersen @ajgdls I have made some updates to try to copy the way that Tim shuts things down in the FrameReceiver. The app now runs the FWC::run() function in a try catch block, the run function calls waitForShutdown and when that returns does the shutdown procedure.

I have also completed the changes required to run without a client, however it will only write 1 file, if you want more you have to send a start writing command from the client in between datasets. I have added the option to register a blocking callback in IFrameCallback and used that to register the FWC with the FileWriter (plugin), so that when the expected frames are received the FW doesn't process more frames while the FWC is shutting down (This hasn't been a problem, but it seemed like it could be).

GDYendell:
Have added some more to the PR. It now has support for Percival and Excalibur single-shot mode. I also believe it now blocks the plugin queue properly when shutting down; before I think it sometimes wrote an extra frame to file.

GDYendell:
Hi all, 

I have rebased this PR off of repo_reorg to allow that to be merged into master first. I think this is ready to be merged once that happens. A lot has changed since the PR was made initially. @timcnicholls just to let you know, this commit now touches IPCMessage, so you may want to take a look at that.

This will allow Alan, Matt and Myself to work from this branch as a base for Eiger developments and then we should be able to merge back to master cleanly.

PR #14, opened 2017-03-02
Replaced all trace of SharedMemoryParser with SharedBufferManager.

ajgdls:
Hi @timcnicholls this pull request cleans up the frame processor by making use of the available common shared memory class instead of using its own copy.


timcnicholls:
Thanks @ajgdls , happy to merge this PR.

PR #15, opened 2017-03-07
Repo reorg classloader

ajgdls:
Updated odin-data frame receiver to work with the classloader.  Updated percival and excalibur classes to build into their own shared libraries and then updated the application to load the appropriate one.  There is still some specific code in the receiver application class, mostly to do with reading configuration but these changes will allow a new third party project to build its own decoder library and in the future work with odin-data.

Please review and ask for any clarification...


ajgdls:
Hi Tim, no problems.  Agreed regarding the namespace.

PR #16, opened 2017-04-20
Repo reorg frontend

ajgdls:
Front end updates to support UDP or ZMQ interface to detector.  Rework of RxThread into an interface, followed by creation of specific interfaces.  Updated command line parameters.
What is missing here is testing updates, I had to comment out some tests and new tests are required.  I can work on those next but wanted to get this pull request created.

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -1,1 +1,2 @@
+add_subdirectory(include)
 add_subdirectory(src)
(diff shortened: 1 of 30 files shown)


timcnicholls:
@ajgdls Just checked the current state of this PR against the EXCALIBUR test system in my lab and it works fine.

ajgdls:
Hi @timcnicholls I've made the additional changes to split the decoders and also fixed the existing tests to work again.  I've also updated the ClassLoader namespace.  If you want to merge this PR before I strip out the Percival and Excalibur decoders go ahead because I need to create the CMake structure for those projects and so will not finish that today I expect.  I can easily raise another PR when I'm ready to remove those files from this repo.

timcnicholls:
Thanks @ajgdls I'm happy to merge this as is. Separating out the plugins can come afterwards as you suggest. It is sufficiently different to warrant its own PR and I also need chance to recreate the plugin in e.g. odin-excalibur. Can you point me at the right EIGER and/or LATRD repos (and branches) for an example of the CMake setup please?

PR #20, opened 2017-05-09
Frame processor renaming files and classes.

ajgdls:
Also updated namespace and reworked make system to produce a shared library for external plugins.


timcnicholls:
@GDYendell Are you happy for me to merge this Gary?

ajgdls:
The Excalibur and Percival plugins are to be removed from this repository so I left them as they were for now.

PR #21, opened 2017-05-12
Repository reorgansiation

timcnicholls:
First pass at repository reorganisation is reflected in this PR. The following steps have been undertaken:

- application, namespace and class names made consistent (frameReceiver, frameProcessor, odinData)
- removal of detector-specific FR and FP plugins, which are now in separate repos
- rationalisation of test code structure to mimic standard source tree layout
- fix for unnecessary null-termination handling on ZeroMQ IpcChannels


ajgdls:
Cloned, built OK.  Ran tests OK.  Ran applications as expected.  Checked libraries and they look OK.  So I think I'm giving this the thumbs up.

GDYendell:
The build looks a lot cleaner now, but I see the FrameReceiverTest builds it's own object files for the FrameReceiver source. Does it need to do that? It was worse before where each plugin seemed to build its own version of Frame, DataBlock etc and it made it difficult to generate coverage reports.

timcnicholls:
@GDYendell You're right - I was aware that was a difference. The FR and FP plugin dependencies are rather different. for the FP Alan builds a core library that each plugin links against for Frame, DataBlock etc. The FR plugins are at the top of the dependency chain so don't have the same requirements, although we could modify the FR build to have a core library shared between main and test apps. Also there's no coverage on the FR yet. Propose to merge this PR anyway and add both of those things as issues for future enhancements if you agree?

GDYendell:
Sure. I can make that change when I eventually add coverage support.

Other than that, it builds and I can run my latest test code.

PR #22, opened 2017-05-18
Created odin-data CMake install targets

timcnicholls:


common/include/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
...
-INSTALL(FILES ${HEADERS} DESTINATION ${CMAKE_BINARY_DIR}/include)
+# Install header files into installation prefix
 
+SET(HEADERS ClassLoader.h gettime.h IpcChannel.h IpcMessage.h IpcReactor.h OdinDataException.h SharedBufferManager.h)
 SET(RAPIDJSON_INCLUDE_DIR rapidjson)
-message("RAPIDJSON_INCLUDE_DIR ${RAPIDJSON_INCLUDE_DIR}")
-INSTALL(DIRECTORY ${RAPIDJSON_INCLUDE_DIR} DESTINATION ${CMAKE_BINARY_DIR}/include)
+
+INSTALL(FILES ${HEADERS} DESTINATION include)
+INSTALL(DIRECTORY ${RAPIDJSON_INCLUDE_DIR} DESTINATION include)
```
ulrikpedersen: Do we need to install rapidjson? Could that not collide with a system-installed version? 
I think each of the other repos include a copy of rapidjson headers - which is not great either....

frameReceiver/src/CMakeLists.txt
```diff
@@ -14,4 +14,5 @@ add_executable(frameReceiver ${APP_SOURCES})
 
 target_link_libraries(frameReceiver ${COMMON_LIBRARY} ${Boost_LIBRARIES} ${LOG4CXX_LIBRARIES} ${ZEROMQ_LIBRARIES})
 
+install(TARGETS frameReceiver RUNTIME DESTINATION bin)
```
ulrikpedersen: What is the effect of the RUNTIME bit here? The other install functions don't have it.

common/include/CMakeLists.txt
timcnicholls: I doubt there's a collision risk as RAPIDJSON is header only - there are no built artefacts. We also don't search for it as an external dependency but the plugins require it. The version we use is effectively frozen by virtue of the fact it's in our repo.

frameReceiver/src/CMakeLists.txt
ulrikpedersen: Actually I've RTFM: https://cmake.org/cmake/help/v3.0/command/install.html 
Still don't know what effect it has on the executable though 😀

timcnicholls: CMake identifies RUNTIME build artefacts as binaries associated with a project. See:

[https://cmake.org/cmake/help/v3.0/command/install.html](https://cmake.org/cmake/help/v3.0/command/install.html)

and examples therein. It's probably not necessary with that particular use since there is only one target specified in the rule - with more it can be used to distribute e.g. libraries to other destination paths. I also added it for frameProcessor. I can remove it if necessary.

common/include/CMakeLists.txt
ulrikpedersen: OK, I understand. By publishing rapidjson in a central repo like odin-data we can make use of exactly the same version throughout. We just need to remove the other copies in the other repos and start depending on this one.

timcnicholls: Agreed, it shouldn't be in the other repos (it isn't in EXCALIBUR).

ulrikpedersen:
I think it would be good to add this step to travis as is just proves the 'make install' does what we expect after making this particular change. You can even 'test' it with an ls in the prefix/* dir...

Later it would be good to run system-tests using the installed executables & libs.

frameReceiver/src/CMakeLists.txt
timcnicholls: Comment ping-pong 😉 I originally thought is also ensured a runtime artefact was executable etc when installed, but on a second read I'm not sure. I just thing it allows separation of destination.

timcnicholls:
@ulrikpedersen Added install steps and an 'ls -R' "test" - builds OK. Happy for me to merge now?

PR #28, opened 2017-06-09
Offset queuing

GDYendell:
@ulrikpedersen @ajgdls ,

This is an update to allow offset adjustments to be queued. Two things to consider before this is merged:

- When an adjustment is requested it is absolute; the difference between the raw frame id and the offset in the current file to write to. In this case it will be easier to set a specific offset to write to, but would require calculation to rewind by N frames. The alternative is that it would be relative, in which case to set a specific offset the control will have to store the cumulative adjustment, so that it knows what additional adjustment to make. Whereas rewinding by N frames is easier; you don't care where it is, you just send an adjustment of N frames and it will work.

- There is a default for the configuration of the adjustment active ID, if it is not given it will be applied to the frame id matching the offset (so it will start writing at the beginning of the file) and therefore applied to all frames, which may be useful. However, it will still be queued, so if there are other offsets yet to be used, it will not apply straight away. It would be possible for the active frame not given case to skip the queue and apply straight away (and maybe empty the queue), if that is a use case we want.

Any thoughts on these points?

### DIFF
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -410,11 +410,9 @@ FileWriterPlugin::HDF5Dataset_t& FileWriterPlugin::get_hdf5_dataset(const std::s
  */
 size_t FileWriterPlugin::calculateFrameOffset(size_t frame_no) {
     size_t frame_offset = this->adjustFrameOffset(frame_no);
-
     if (this->concurrent_processes_ > 1) {
         // Check whether this frame should really be in this process
-        // Note: this expects the frame numbering from HW/FW to start at 1, not 0!
-        if ( (((frame_no-1) % this->concurrent_processes_) - this->concurrent_rank_) != 0) {
+        if (frame_offset % this->concurrent_processes_ != this->concurrent_rank_) {
             LOG4CXX_WARN(logger_, "Unexpected frame: " << frame_no
                                 << " in this process rank: "
                                 << this->concurrent_rank_);
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -12,26 +12,27 @@
 namespace FrameProcessor
 {
 
-const std::string FileWriterPlugin::CONFIG_PROCESS             = "process";
-const std::string FileWriterPlugin::CONFIG_PROCESS_NUMBER      = "number";
-const std::string FileWriterPlugin::CONFIG_PROCESS_RANK        = "rank";
+const std::string FileWriterPlugin::CONFIG_PROCESS                        = "process";
+const std::string FileWriterPlugin::CONFIG_PROCESS_NUMBER                 = "number";
+const std::string FileWriterPlugin::CONFIG_PROCESS_RANK                   = "rank";
 
-const std::string FileWriterPlugin::CONFIG_FILE                = "file";
-const std::string FileWriterPlugin::CONFIG_FILE_NAME           = "name";
-const std::string FileWriterPlugin::CONFIG_FILE_PATH           = "path";
+const std::string FileWriterPlugin::CONFIG_FILE                           = "file";
+const std::string FileWriterPlugin::CONFIG_FILE_NAME                      = "name";
+const std::string FileWriterPlugin::CONFIG_FILE_PATH                      = "path";
 
-const std::string FileWriterPlugin::CONFIG_DATASET             = "dataset";
-const std::string FileWriterPlugin::CONFIG_DATASET_CMD         = "cmd";
-const std::string FileWriterPlugin::CONFIG_DATASET_NAME        = "name";
-const std::string FileWriterPlugin::CONFIG_DATASET_TYPE        = "datatype";
-const std::string FileWriterPlugin::CONFIG_DATASET_DIMS        = "dims";
-const std::string FileWriterPlugin::CONFIG_DATASET_CHUNKS      = "chunks";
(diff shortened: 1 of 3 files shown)


frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -646,11 +646,25 @@ void FileWriterPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMess
...
-    LOG4CXX_DEBUG(logger_, "Setting frame offset adjustment to "
-                           << config.get_param<int>(FileWriterPlugin::CONFIG_OFFSET_ADJUSTMENT));
-    frame_offset_adjustment_ = (size_t)config.get_param<int>(FileWriterPlugin::CONFIG_OFFSET_ADJUSTMENT);
+    size_t adjustment = config.get_param<size_t>(FileWriterPlugin::CONFIG_OFFSET_ADJUSTMENT);
+    size_t active_frame;
+    if (config.has_param(FileWriterPlugin::CONFIG_OFFSET_ADJUSTMENT_ACTIVE_FRAME)) {
+      // This offset will start from a specific frame
+      active_frame = config.get_param<size_t>(FileWriterPlugin::CONFIG_OFFSET_ADJUSTMENT_ACTIVE_FRAME);
+      if (active_frame < adjustment) {
+          LOG4CXX_ERROR(logger_, "Invalid offset adjustment ignored."
```
GDYendell: Comment wrong. It just throws an error.

ulrikpedersen:
As just discussed with @GDYendell:

* The ability to apply a relative offset from the client/user side is most useful. This puts less effort on the client to maintain the state of how much offset have been applied until now. The client side will just see an API which says "rewind N frames from frame no. XX".

* The ability to give a global offset when not specifying the frame ID is also useful: typically this will only be applied once before acquisition in order to support detectors that produce frames starting with ID 1 (instead of 0)...

We also discussed a corner-case which has not yet been catered for: detectors producing frames out-of-order. The frameReceiver has the feature of waiting for late packets to arrive - so each frame can "time-out" if some packets have not been received. This causes an out-of-order stream of frames to happen. Because of this the file writer must keep a "map" of some description to maintain historical offsets as well as future ones.

I propose to pull out the offset storage "map" and associated functionality into a separate (testable) class. Basically an ordered map of (absolute) offsets and frame IDs is to be wrapped in a class with methods to set a relative offset, change a global offset (recalculating all existing absolute offsets), and get an offset for a particular frame ID. @GDYendell is thinking about this now.

GDYendell:
I don't think I was clear when explaining the offset only call. It is currently just a shorter API:

`queueOffset(active_frame=10, offset=10)`

can just be

`queueOffset(active_frame=10)`

which will just start writing from the beginning of the file when it gets frame 10. All offsets before this become irrelevant because we have just started overwriting everything (even in the case of out-of-order frames).

What is the use case for re-calculating previous offsets which have already been applied to written frames?

ulrikpedersen:
> `queueOffset(active_frame=10, offset=10)`
> can just be
> `queueOffset(active_frame=10)`

Ah, I see. I had understood it to be a function that looks something like `setOffset(active_frame=default, offset)` or just plain `setOffset( offset )`. 

> What is the use case for re-calculating previous adjustments which have already been applied to written frames?
> Do you mean it will re-calculate the queued adjustments that haven't been used yet?

The only use case that I can think off for applying a global offset is the starting frame (for detectors starting with a frame ID of something larger than 0). As a corner case we also have to support frames coming out of order - and that's why it should be re-calculating offsets for previous frames... 

I agree it could potentially cause significant confusion to apply another global offset in the middle of acquisition. Perhaps a more accurate (although abstract) description of the class that we're after would go something like:

A class is required to maintain a collection of offsets, mapping a frame ID to an offset in file. The interface should provide methods to:
* Add new relative offsets to 'rewind'
* Look-up absolute offset based on frame ID
* Set a global offset to support cases where acquisition with a frame ID >=0
* Maintain a consistent state of the frame ID and absolute offsets regardless of the (frame ID) ordering of calls for offsets.

Does that make more sense? I've probably missed something... Try to sketch a class definition that support the above.

GDYendell:
> As a corner case we also have to support frames coming out of order - and that's why it should be re-calculating offsets for previous frames...

If we are recalculating offsets that have already applied, then the frames already written with those adjustments must be in the wrong place and any late frames that come after the recalculation will be in the wrong relative offset. We can fully support acquisitions starting from frame id > 0 by a normal adjustment before we get any frames, this will then apply to all future frames because will be relative to the current applied adjustment.

I think it will be best if I just sketch up what I think it needs to be and then we can more easily see if it can handle edge cases.

GDYendell:
Initial idea for new class [here](https://github.com/odin-detector/odin-data/commit/1b8ed7c37f7d837e2194846b271b72e08e3b0705)

ulrikpedersen:
> Initial idea for new class here

Thanks. I've made a few comments inline on the commit 1b8ed7c37f7d837e2194846b271b72e08e3b0705 

Take a look at the comments and tweak/change as you find appropriate. Then setup a test framework that exercises the use-cases we've discussed: plain sequence of frames, missing frames in the sequence, rewinding, frames out of order, and various combinations of this... All tests will fail at first of course but things get a lot clearer when you write out test cases. Yay TDD! :laughing:

GDYendell:
Initial implementation of this a97752d3f3.

Includes:
- Working class
- Integration into FileWriterPlugin
- Updates to make FileWriterPlugin tests pass

Does not include:
- Integration into @MattTaylorDLS updates for handling multiple acquisitions
- Tests for the new class

PR #39, opened 2017-06-23
Check frame dimensions, data type and compression against dataset

GDYendell:
Error raised if values are set and do not a match the definition
for the dataset named by the frame.

### DIFF
--- a/frameProcessor/include/Frame.h
+++ b/frameProcessor/include/Frame.h
@@ -97,9 +97,13 @@ public:
 
   void set_dimensions(const std::string& type, const std::vector<unsigned long long>& dimensions);
   dimensions_t get_dimensions(const std::string& type) const;
+  std::string get_compression() const;
+  int get_data_type() const;
   void set_parameter(const std::string& index, size_t parameter);
   size_t get_parameter(const std::string& index) const;
   bool has_parameter(const std::string& index);
+  void set_compression(std::string compression);
+  void set_data_type(int data_type);
 
 private:
   Frame();
@@ -120,6 +124,10 @@ private:
   unsigned long long frameNumber_;
   /** Map of dimensions, indexed by name */
   std::map<std::string, dimensions_t> dimensions_;
+  /** Compression type of raw data */
+  std::string compression_;
+  /** Compression type of raw data */
+  int data_type_;
   /** General parameter map */
   std::map<std::string, size_t> parameters_;
   /** Pointer to raw data block */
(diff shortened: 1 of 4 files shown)


frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -811,6 +847,9 @@ void FileWriterPlugin::configureDataset(OdinData::IpcMessage& config, OdinData::
         dset_def.compression = config.get_param<std::string>(FileWriterPlugin::CONFIG_DATASET_COMPRESSION);
         LOG4CXX_DEBUG(logger_, "Enabling compression: " << dset_def.compression);
       }
+      else {
+        dset_def.compression = "none";
```
MattTaylorDLS: Can we make this a constant somewhere?

frameProcessor/include/Frame.h
```diff
@@ -120,6 +124,10 @@ class Frame
   unsigned long long frameNumber_;
   /** Map of dimensions, indexed by name */
   std::map<std::string, dimensions_t> dimensions_;
+  /** Compression type of raw data */
+  std::string compression_;
+  /** Compression type of raw data */
```
MattTaylorDLS: comment wrong

frameProcessor/src/Frame.cpp
```diff
@@ -179,6 +181,26 @@ dimensions_t Frame::get_dimensions(const std::string& type) const
...
+ * \return compression type.
+ */
+std::string Frame::get_compression() const
+{
+  return compression_;
+}
+
+/** Retrieves the data type of the raw data.
+ *
+ * -1: Unset 0: UINT8, 1: UINT16, 2: UINT32
```
MattTaylorDLS: comma needed after Unset

GDYendell:
Frame has no concept of those enums currently; they are in the FileWriterPlugin. They could be copied to the Frame so it can return the correct types. Or maybe we should create a common place to define them.

ajgdls:
OK, I think the enums should be present in the frame as the frame now contains solid implementations that make use of them.  The file writer happens to also want to use them but it no longer owns them in my opinion.  Any class that uses frame object may or may not need to use those methods and so I think the frame should own the enums.
Perhaps this is another task that should be in a separate PR?

GDYendell:
Sounds reasonable. If we are doing that can we rename PixelType and its enum values to something more suitable for a property of a Frame; maybe DataType, or BitDepth?

ajgdls:
Sure that is fine with me.  So I'll approve this PR, we can create another for the follow up items.

PR #40, opened 2017-06-27
Filename

MattTaylorDLS:
Added concept of an acquisition ID to the Frame and FileWriterPlugin.

FWP now has a new struct defined, to hold details about acquisitions. There is one for the current acquisition and one for the next acquisition.

If the Frame object currently being processed contains an acquisition ID that doesn't match the current acquisition ID, it wont write the Frame. If it matches the next acquisition ID, the current file is closed, and the next one is started.
If the Frame object contains no acquisition ID, it acts as it did before.


frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -33,6 +33,7 @@ const std::string FileWriterPlugin::CONFIG_FRAMES              = "frames";
 const std::string FileWriterPlugin::CONFIG_MASTER_DATASET      = "master";
 const std::string FileWriterPlugin::CONFIG_OFFSET_ADJUSTMENT   = "offset";
 const std::string FileWriterPlugin::CONFIG_WRITE               = "write";
+const std::string FileWriterPlugin::ACQUISITION_ID             = "acquisitionid";
```
GDYendell: Could this be "acquisition_id"

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -502,42 +500,42 @@ void FileWriterPlugin::processFrame(boost::shared_ptr<Frame> frame)
...
 
-      // Check if we have written enough frames and stop
-      if (framesToWrite_ > 0 && framesWritten_ == framesToWrite_) {
-        this->stopWriting();
-      }
+    // Check if we have written enough frames and stop
+    if (currentAcquisition_.framesToWrite_ > 0 && framesWritten_ == currentAcquisition_.framesToWrite_) {
+      this->stopWriting();
+      // Start next acquisition
+      this->startWriting();
```
GDYendell: Have we checked we actually have a nextAcquisition here?

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -603,15 +601,31 @@ size_t FileWriterPlugin::getDatasetFrames(const std::string dset_name)
...
   if (!writing_) {
+    this->currentAcquisition_ = nextAcquisition_;
+    this->nextAcquisition_ = Acquisition();
+
+    // If filename hasn't been explicitly specified, create it from the acquisition ID and rank
+    if (this->currentAcquisition_.fileName_.empty() && (false == this->currentAcquisition_.acquisitionID_.empty())) {
+      std::stringstream filenameWithRank;
+      filenameWithRank << this->currentAcquisition_.acquisitionID_ << "_r" << this->concurrent_rank_ << ".hdf5";
+      this->currentAcquisition_.fileName_ = filenameWithRank.str();
+    }
```
GDYendell: It seems that here currentAcquisition_.fileName_ is not the filename, fileNameWithRank is. Could we rename the struct member to something more clear? Maybe just currentAcquisition_.name?

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -957,4 +972,60 @@ void FileWriterPlugin::clearHdfErrors()
...
 
+/** Check if the frame contains an acquisition ID and start a new file if it does and it's different from the current one
+ *
+ * If the frame object contains an acquisition ID, then we need to check if the current acquisition we are writing has
+ * the same ID. If it is different, then we close the current file and create a new one and start writing.
+ * If we are not currently writing then we just create a new file and start writing.
+ *
+ * \param[in] frame - Pointer to the Frame object.
+ */
+void FileWriterPlugin::checkAcquisitionIDInFrame(boost::shared_ptr<Frame> frame) {
```
GDYendell: Could this be just checkAcquisitionID? The parameter is a Frame, so I think it's clear enough.

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -957,4 +972,60 @@ void FileWriterPlugin::clearHdfErrors()
...
+  if (false == frame->get_acquisition_id().empty()) {
```
GDYendell: if (!frame->get_acquisition_id().empty())

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -957,4 +972,60 @@ void FileWriterPlugin::clearHdfErrors()
...
+    } else {
+      if (frame->get_acquisition_id() == nextAcquisition_.acquisitionID_) {
+        LOG4CXX_DEBUG(logger_, "Acquisition ID sent in frame matches next acquisition ID. Starting next");
+        startWriting();
+      } else {
+        LOG4CXX_WARN(logger_, "Unexpected acquisition ID on frame [" << frame->get_acquisition_id() << "] for frame " << frame->get_frame_number());
+      }
+    }
+  }
+}
```
GDYendell: This can be simplified a little. stopWriting() is a nop if we aren't writing, so you don't need a seperate check branch for only calling startWriting().

If (next) {stop, start}
else if (!current) {error}

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -957,4 +972,60 @@ void FileWriterPlugin::clearHdfErrors()
...
+
+/** Creates and returns the Meta Header json string to be sent out over the meta data channel
+ *
+ * This will typically include details about the current acquisition (e.g. the ID)
+ *
+ * \return - a string containing the json meta data header
+ */
+std::string FileWriterPlugin::getMetaHeader() {
```
GDYendell: Slightly confused by this. Is it an unrelated commit to the acquisition refactor?

frameProcessor/src/FileWriterPlugin.cpp
GDYendell: Sorry this was a bad place to put the tag... I am referring to the whole if chain above.

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: OK

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: No, currently it will log a warning if there is no next acquisition when it tries to startWriting(). I thought this would be preferable to silently doing nothing, but maybe we could check here and log a specific warning/debug to say there was no next acquisition to start, rather than calling startWriting and hoping.

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: the fileName is the file name. If the fileName_ has been set, then this is what will be used as the filename, regardless of what the acquisition id is. If fileName_ hasn't been set, it is created from the acquisition and rank here, but this is immediately set to fileName_. For the current acquisition that is being written, fileName_ will always be the actual name of the file being written

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: OK

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: OK

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: OK

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: It's kind of related. Because of the addition of the acquisition ID, this also needs to be sent out over the meta channel. It was then cleaner to separate all of this out into a method to generate the header that would be sent, rather than doing it inline.

MattTaylorDLS:
Yeah not sure how there are two commits for the total-frames. I usedeclipse to change the commit message, but it doesn't seem to have done it cleanly as far as github is concerned

frameProcessor/src/FileWriterPlugin.cpp
GDYendell: Oh I see I misunderstood because the name filenameWithRank it sounds like you are adding one to the other. So, it is only an option to add rank to the file name? The user could just specify any filename they want and have no concept of what rank it was in the name? I think they should provide a base name and maybe a format string and then we always do some generation.

Also there is another false == ... in here.

frameProcessor/src/FileWriterPlugin.cpp
GDYendell: Makes sense.

frameProcessor/src/FileWriterPlugin.cpp
GDYendell: I don't even think it should log a specific warning here. It is entirely reasonable to just stop writing and have no further acquisitions. It can then be configured for the next acquisition at some later time and it will start writing again when it gets the first frame for that.

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: I'll change the name of that variable. Yes, I think the user should be able to specify the exact filename that they want. If they don't specify the filename, then we will make one up from the acquisition ID and the rank. The ability to specify a template that will be used to generate a filename is a future enhancement

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: OK, I put the logging in as it might have been useful when debugging back to back acquisitions if it explicitly stated that it didn't start the new acquisition because there was no next filename. It would make it easier to track why an acquisition failed, but i'll leave it out for now and see what happens.

MattTaylorDLS:
If we were to make the acquisition_id a first class buffer parameter, then all that would need to change is instead of getting out of the 'Frame' object in the 'getAcquisitionID' function, we would get it out of the buffer.

timcnicholls:
Thanks @MattTaylorDLS 

I'll approve the PR, pending any changes you agree with @GDYendell . Feel free to merge when you're both happy.

PR #48, opened 2017-07-04
Update CLI argument parsing

GDYendell:
* Update for compression now being an int
    * Add output-dir for where to save HDF5 files
    * Add more thorough checks for no-client mode
    * Rearrange argument parsing for clearer help message
    * Some other small changes for clarity

### DIFF
--- a/frameProcessor/src/FrameProcessorApp.cpp
+++ b/frameProcessor/src/FrameProcessorApp.cpp
@@ -55,44 +55,47 @@ void parse_arguments(int argc, char** argv, po::variables_map& vm, LoggerPtr& lo
     config.add_options()
         ("logconfig,l",   po::value<string>(),
            "Set the log4cxx logging configuration file")
-        ("no-client,N",   po::bool_switch(),
-           "Enable full initial configuration to run without any client controller")
-        ("single-shot,S", po::bool_switch(),
-           "Shutdown after one dataset written")
+        ("ctrl",          po::value<std::string>()->default_value("tcp://127.0.0.1:5004"),
+           "Set the control endpoint")
+        ("sharedbuf",     po::value<std::string>()->default_value("FrameReceiverBuffer"),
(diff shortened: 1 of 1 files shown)


frameProcessor/src/FrameProcessorApp.cpp
```diff
@@ -55,44 +55,47 @@ void parse_arguments(int argc, char** argv, po::variables_map& vm, LoggerPtr& lo
     config.add_options()
         ("logconfig,l",   po::value<string>(),
            "Set the log4cxx logging configuration file")
-        ("no-client,N",   po::bool_switch(),
-           "Enable full initial configuration to run without any client controller")
-        ("single-shot,S", po::bool_switch(),
-           "Shutdown after one dataset written")
+        ("ctrl",          po::value<std::string>()->default_value("tcp://127.0.0.1:5004"),
+         "Set the name of the shared memory frame buffer")
```
ajgdls: Hi @GDYendell are you sure this comment relates to this command line parameter?  I've been confused by this, unless I misunderstand the diff.

frameProcessor/src/FrameProcessorApp.cpp
```diff
@@ -55,44 +55,47 @@ void parse_arguments(int argc, char** argv, po::variables_map& vm, LoggerPtr& lo
...
+        ("sharedbuf",     po::value<std::string>()->default_value("FrameReceiverBuffer"),
+         "Set the control endpoint")
```
ajgdls: Same comment, it looks like these might be switched actually.

GDYendell: Interesting...

GDYendell:
It should stop crashes for the specific example you gave me. If there are any others they should probably be fixed here before merging.

ajgdls:
Oh I see, you check for all the items required before going into the no client mode.

PR #50, opened 2017-07-20
Malcolm and Soft IOC integration

GDYendell:
This is a varied PR of different bits I had to add for Malcolm and general things to make it ready for production.

I have changed the IPCChannel and IPCMessage a little to support a simpler interface for status requests. FrameProcessor has been updated for this change.
I have written some new clients that should be easily importable to Malcolm and other controllers. The FrameReceiverClient is basically a stub for now.
There will be an accompanying PR in eiger-daq for EigerFan and MetaListener clients.

I plan to tag both of these as 0-2 (I think that's what we said) and that will be the version that is given to GDA on VMXi.


GDYendell:
Closing this and tagging from DLS specific branch on dls-controls for initial VMXi GDA integration. This will be tidied and made site neutral in a new PR.

Issue #37, opened 2017-06-22
Transmit buffer configuration information from FR to FP


timcnicholls:
In order to allow reconfiguration of the odin-data chain between acquisitions, e.g. load a different decoder, set up varying readout modes, the FR will receive configuration messages over its control channel. To ensure that the FP remains in step with the shared buffer configuration, the FR should send, over the frame ready notification channel, buffer configuration messages that cause the FP to tear down its current shared buffer manager instance and reconfigure according the info contained in the message. 

To allow for the case where the FP asynchronously starts/stops/dies, the FP will also be able to request buffer configuration information over the release channel to instruct the FR to send out its current configuration.

Consideration needs to be given in both the FR and FP to the handling of queued/committed buffers in the event of a buffer reconfiguration. The FR should only respond to a reconfiguration command over its control channel once all pending buffers have been released by the FP, and the FP should drop and release any buffers for which there are ready notifications queued in the event of restarting and requesting a buffer configuration.

timcnicholls:
Depends on #38

PR #52, opened 2017-07-28
Shared buffer configuration notification

timcnicholls:
This PR handles notification of buffer configuration from frame receiver to processor. Config parameters for the name of the shared memory buffer are also removed from the frame processor.


timcnicholls:
Closes #37

config/fp_py_test.config
```diff
@@ -0,0 +1,9 @@
+[FrameProcessor]
+ctrl_endpoint=tcp://127.0.0.1:5000
+sharedbuf=FrameReceiverBuffer
```
ajgdls: Is this sharedbuf entry now redundant?

ajgdls: Oh actually I misunderstood, this config is not for the FP I guess?

frameProcessor/src/SharedMemoryController.cpp
```diff
@@ -112,29 +150,50 @@ void SharedMemoryController::handleRxChannel()
...
-
-        // Set the frame number
-        frame->set_frame_number(rxMsg.get_param<int>("frame", 0));
-
-        // Loop over registered callbacks, placing the frame onto each queue
-        std::map<std::string, boost::shared_ptr<IFrameCallback> >::iterator cbIter;
-        for (cbIter = callbacks_.begin(); cbIter != callbacks_.end(); ++cbIter) {
-          cbIter->second->getWorkQueue()->add(frame);
+
+        if (sbm_) {
```
ajgdls: This check prevents crash on no valid shared memory registration?  I guess in the real application this is not a normal situation but I should probably write a unit test for this.  Note to self...

PR #53, opened 2017-08-07
Enable SWMR writing and flush on each frame

GDYendell:
This is something that GDA will probably want as they are adding their support to our Eiger Stack, but I thought it would be good to get this standalone change straight into master before we do so. I think it's a fairly simple change. I haven't added the flush to writeSubframes, as it will be going soon, I think.

### DIFF
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -157,7 +157,14 @@ void FileWriterPlugin::writeFrame(const Frame& frame) {
   status = H5DOwrite_chunk(dset.datasetid, H5P_DEFAULT,
                            filter_mask, &offset.front(),
                            frame.get_data_size(), frame.get_data());
-  H5Dflush(dset.datasetid);
+  if (status < 0) {
+    throw RuntimeException("Failed to write chunk");
+  }
+
+  status = H5Dflush(dset.datasetid);
+  if (status < 0) {
+    throw RuntimeException("Failed to flush data to disk");
+  }
 
   // Send the meta message containing the frame written and the offset written to
   rapidjson::Document document;
@@ -192,8 +199,6 @@ void FileWriterPlugin::writeFrame(const Frame& frame) {
   document.Accept(writer);
 
   publishMeta("writeframe", buffer.GetString(), getMetaHeader());
-
-  assert(status >= 0);
 }
 
 /**
@@ -637,7 +642,10 @@ void FileWriterPlugin::startWriting()
     }
 
     // Start SWMR writing
-    assert(H5Fstart_swmr_write(this->hdf5_fileid_) >= 0);
+    herr_t status = H5Fstart_swmr_write(this->hdf5_fileid_);
+    if (status < 0) {
+      throw RuntimeException("Failed to enable SWMR writing");
+    }
 
     // Reset counters
     framesWritten_ = 0;
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -161,10 +161,12 @@ void FileWriterPlugin::writeFrame(const Frame& frame) {
     throw RuntimeException("Failed to write chunk");
   }
 
+#if H5_VERSION_GE(1,9,178)
   status = H5Dflush(dset.datasetid);
   if (status < 0) {
     throw RuntimeException("Failed to flush data to disk");
   }
+#endif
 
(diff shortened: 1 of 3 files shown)


GDYendell:
Should SWMR ever not work?

I'm pretty sure it's not related. I'm not sure what the problem is, though.

ajgdls:
"Should SWMR ever not work".
I hope that it will always work.  But I assume there will be at least one time at some point in the future where it doesn't (probably at 3am during a very long experiment etc etc), we need to make sure that when it does fail the log explains what has happened to cut down on investigation time.

Currently if it failed would we get any kind of message in the logs or would it just look like a sudden crash of the application?

GDYendell:
I think it would say 'assertion failed H5Fstart_swmr_write...' with the line number, wouldn't it?

If that's not useful, do we need to have a check through and change assertions for error messages?

ulrikpedersen:
We really shouldn't depend on assert to detect error conditions. From the man page:

> If the macro NDEBUG is defined at the moment <assert.h> was last included, the macro assert() generates no code, and hence does nothing at all.

So later when we build a highly optimized version of the application we would typically switch off anything related to debugging - and set NDEBUG. Then instead of a useful assert/error message, the application will crash later on in some, seemingly unrelated fashion. So check the status output of the function and throw an exception if it fails, because it really shouldn't fail and if it does then there isn't anything sensible we can do about it to carry on.

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -157,6 +157,7 @@ void FileWriterPlugin::writeFrame(const Frame& frame) {
   status = H5DOwrite_chunk(dset.datasetid, H5P_DEFAULT,
```
ulrikpedersen: We should check this status as well and throw exception if it fails.

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -157,6 +157,7 @@ void FileWriterPlugin::writeFrame(const Frame& frame) {
...
                            filter_mask, &offset.front(),
                            frame.get_data_size(), frame.get_data());
+  H5Dflush(dset.datasetid);
```
ulrikpedersen: Does H5Dflush return a status? We should check it...

GDYendell:
I will update this PR to handle the status properly and throw an explicit error. It is probably a separate PR to go through and update the existing instances; there are about 20 in all.

ulrikpedersen:
> I will update this PR to handle the status properly and throw an explicit error. It is probably a separate PR to go through and update the existing instances; there are about 20 in all.

I agree and will create another JIRA ticket for this task.

ulrikpedersen:
A couple of related issues with the travis build: 

- the .travis-install-hdf5.sh script fails on travis. 
- The script attempt to install HDF5 v1.8.16 - which does not have SWMR so the build will fail.

Fix up the hdf5 install script and update it to v1.10.1

GDYendell:
Done that. Now the branch builds OK, but the PR build fails trying to build h5py. I retried it and got the same again.

ajgdls:
I recently had lots of trouble with pip installs.  For me it was a combination of caching and pip/virtualenv/setuptools not being at latest versions.  If you manually perform the build (ie setup virtualenv etc) does it work still?

GDYendell:
Just did a clean checkout, built OdinData, created a venv and then did `HDF5_DIR=/dls_sw/prod/tools/RHEL6-x86_64/hdf5/1-10-1/prefix pip install h5py` (note without the --user, because that doesn't work at all) and it worked OK.

GDYendell:
Tried with --no-cache, which also works OK.

ajgdls:
So perhaps the travis build is using older versions.
You could try to add the following lines to travis before it installs the requirements
pip install --upgrade pip
pip install --upgrade virtualenv

ajgdls:
Oh not so easy.  This isn't building from a virtualenv so it tried to upgrade pip in a protected location...

ajgdls:
Additional updates:
1) This should still compile if using an older version of HDF5.
You need to surround SWMR specific calls with
#if H5_VERSION_GE(1,9,178)
#endif
to ensure they only get compiled with compatible HDF5 library versions.

GDYendell:
We do need to flush even though we are using direct_chunk_write.

ulrikpedersen:
PRs #55 and #54 are waiting for this one to be merged first. 

Arguably we could do the travis fix in another PR - but it was initiated because of the addition of SWMR support which drives the update to hdf5 1.10.1.

I think this is ready to merge.

PR #54, opened 2017-08-09
Separated file writer plugin registration call into a new cpp file.

ajgdls:
This allows the file writer plugin to be subclassed by specific detectors.
Updated CMakeLists.txt files.

### DIFF
--- a/frameProcessor/include/CMakeLists.txt
+++ b/frameProcessor/include/CMakeLists.txt
@@ -1,4 +1,4 @@
 # Install header files into installation prefix
 
-SET(HEADERS DataBlock.h DataBlockPool.h Frame.h FrameProcessorPlugin.h IFrameCallback.h MetaMessage.h WorkQueue.h)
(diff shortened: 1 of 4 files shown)


ajgdls:
Yes I think @GDYendell is working on the Travis build already as part of another PR.  Is this the same error that you noticed Gary?

GDYendell:
I have not seen that error on my SWMR branch, though I do have a different problem there, I'm not sure what this is. Though, I notice that the path we are trying to curl for hdf5 does not exist: https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8.16/src/

Maybe they have moved things around?

ulrikpedersen:
The travis build of hdf5 is mentioned on #53 as well. Also #53 requires updating hdf5 to 1.10.1  which can be found here: https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.10/hdf5-1.10.1/src/hdf5-1.10.1.tar.gz

ajgdls:
@GDYendell are you happy to approve this PR now?

ajgdls:
OK @timcnicholls I'm going to merge this if you are OK with that.

PR #55, opened 2017-08-09
Fix for running FrameProcessor with no Acquisition ID

MattTaylorDLS:
http://jira.diamond.ac.uk/browse/BC-452


GDYendell:
So, does the FileWriterPlugin not actually require an acqID to work?

MattTaylorDLS:
No, if no acquisition ID is specified, it will work as it did before.

A filename will need to be specified as usual though

GDYendell:
Then defaulting the parameter as blank is probably the correct fix. Same as with frames.

MattTaylorDLS:
I'll wait until the SWMR changes have gone in, and then make my changes to those new H5 calls too before committing this

MattTaylorDLS:
Can we merge this too?

ajgdls:
I think you need to rebase from master first to fix the build?

ajgdls:
I think the reason it still fails is because you might have merged dls-controls master but you need to either a) merge odin-data master or else b) update dls-controls master to match odin-data master.

MattTaylorDLS:
Ah yes, thanks. I'm not sure why 'git merge master' chose the dls-controls branch instead of odin-detector, because the remote for my BC452 branch was odin-detector. Seems all good now

ajgdls:
Great, you probably want to include @timcnicholls to get approval that your change doesn't impact the other detectors.

timcnicholls:
@ajgdls @MattTaylorDLS Looks OK to me. Haven't explicitly tested against other detectors but can't see how it would affect them. Happy for you to merge.

GDYendell:
> I'm not sure why 'git merge master' chose the dls-controls branch instead of odin-detector, because the remote for my BC452 branch was odin-detector.

I think this would be because your master is set to track dls-controls/master.

PR #57, opened 2017-08-21
Wrapped H5 calls in the FileWriterPlugin to throw runtime exceptions

MattTaylorDLS:
BC-449

### DIFF
--- a/frameProcessor/include/FileWriterPlugin.h
+++ b/frameProcessor/include/FileWriterPlugin.h
@@ -122,7 +122,7 @@ public:
   void status(OdinData::IpcMessage& status);
   void hdfErrorHandler(unsigned n, const H5E_error2_t *err_desc);
   bool checkForHdfErrors();
-  void handleH5error(std::string message, std::string function, std::string filename, int line, bool throwException=true) const;
+  void handleH5error(std::string message, std::string function, std::string filename, int line) const;
   std::vector<std::string> readHdfErrors();
   void clearHdfErrors();
   std::string getCreateMetaHeader();
--- a/frameProcessor/include/FileWriterPlugin.h
+++ b/frameProcessor/include/FileWriterPlugin.h
@@ -122,6 +122,7 @@ public:
   void status(OdinData::IpcMessage& status);
   void hdfErrorHandler(unsigned n, const H5E_error2_t *err_desc);
   bool checkForHdfErrors();
+  void handleH5error(std::string message, std::string function, std::string filename, int line, bool throwException=true) const;
   std::vector<std::string> readHdfErrors();
   void clearHdfErrors();
   std::string getCreateMetaHeader();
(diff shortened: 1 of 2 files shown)


GDYendell:
So, where it is doing check rather than ensure it is OK to fail and carry on?

ajgdls:
I don't think I understand your comment @GDYendell .  I see a couple of places where the code is checked but does not raise an exception.  I think in reality if any of these calls fail an exception will be raised further down the method but I think this is OK as an initial swap (as we weren't raising errors prior to this PR).

ajgdls:
OK @GDYendell now I understand your comment.  You are asking if we should just always call ensure?

GDYendell:
Yeah.

Also, it seems we are doing >= 0 to pass a bool but then casting it and comparing to 0? Can we not just pass the status from the H5 function and do the logic in the helper function?

ajgdls:
I don't think that's true.  The check is for boolean followed by a question mark (?).  Then if the result is true a void(0) is returned (i.e. do nothing) else the method is called.
If the H5 calls were passed directly then I think negative numbers would give false positives.

MattTaylorDLS:
I haven't changed any of the logic in terms of what it did before if it failed. i.e. if there was an assert there before, it will do an 'ensure' and throw an error, but if there was no assert there before, it will do a 'check' and just log it. If you think we should upgrade every call to 'ensure' then we can, but i don't know what these h5 calls do hence just leaving it.

ajgdls:
Perhaps the logic of checking >=0 could be put into the helper function if that is being executed on every call?  @MattTaylorDLS  ?

MattTaylorDLS:
And yes, as for checking against 0, reading the h5 docs, it's possible that some functions might not return a negative valid but still have an error. If we checked against 0,  we'd have to also pass the expected value/condition. So it's a lot easier for each call to specifically check what it should be when it calls

MattTaylorDLS:
i.e. it's not safe to assume that if it isn't negative, it's ok.

ajgdls:
Mmm, in our class every call is being checked against >= 0 so I would vote for that to go into the check.  Even though what @MattTaylorDLS is correct, we need >= 0 on every call so it probably should go into our check?

GDYendell:
I see, that makes sense. We should consider upgrading these things to ensure now then. I would say create dataset or create dapl failing is not OK. I'm not sure what set_auto and walk actually do...

MattTaylorDLS:
Yeah i can make the dataset and dapl creation ensures.
I'll move the >= 0 into the macro too

ajgdls:
Are there any calls to check left, or are they all ensure?

MattTaylorDLS:
2 calls at the top for 'H5Eset_auto2' and 'H5Ewalk2'

GDYendell:
Are they to do with your H5 error handling, @ajgdls ?

ajgdls:
That's right.  They ensure the error messages are stored (and can therefore be logged) instead of simply printed to screen.  The question is should we just make them ensure also, because they really should work always.  If they aren't successful then something is wrong with the build or library version etc.

GDYendell:
Ensure them too then I reckon.

ajgdls:
OK I agree, it would also cut down on the code as you can remove the check macro @MattTaylorDLS and simplify the method call (no boolean).

ajgdls:
OK looks better I think.

ajgdls:
Can we get this one approved so it can be merged?  Also, is this now the last of the PRs for FileWriter, as I need to rebase the LATRD version against this so I can add the single dimension datasets.  @GDYendell are you happy with this PR?

MattTaylorDLS:
Yes i think this is the last. I'll 'squash and merge' when it's approved

PR #58, opened 2017-08-22
Updated FileWriterPlugin to be able to write single 1D datasets.

ajgdls:
Specifically:
1) Dataset is extended by the outer chunk dimension instead of 1.
2) Offset is calculated taking into account the outer chunk dimension.
3) Dataset dimensions do not need to be defined, this will now not throw.
4) If no dataset dimensions are defined then this is a 1D dataset.
5) Updated some log messages that were expecting 3D datasets.

### DIFF
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -143,11 +143,19 @@ void FileWriterPlugin::createFile(std::string filename, size_t chunk_align)
 void FileWriterPlugin::writeFrame(const Frame& frame) {
   hsize_t frame_no = frame.get_frame_number();
 
+  LOG4CXX_TRACE(logger_, "Writing frame [" << frame.get_frame_number()
+		  << "] size [" << frame.get_data_size()
+		  << "] type [" << frame.get_data_type()
+		  << "] name [" << frame.get_dataset_name() << "]");
   HDF5Dataset_t& dset = this->get_hdf5_dataset(frame.get_dataset_name());
 
   hsize_t frame_offset = 0;
   frame_offset = this->getFrameOffset(frame_no);
-  this->extend_dataset(dset, frame_offset + 1);
+  // We will need to extend the dataset in 1 dimension by the outer chunk dimension
+  // For 3D datasets this would normally be 1 (a 2D image)
+  // For 1D datasets this would normally be the quantity of data items present in a single chunk
+  uint64_t outer_chunk_dimension = this->currentAcquisition_.dataset_defs_[frame.get_dataset_name()].chunks[0];
+  this->extend_dataset(dset, ((frame_offset + 1) * outer_chunk_dimension));
 
   LOG4CXX_TRACE(logger_, "Writing frame offset=" << frame_no  <<
                          " (" << frame_offset << ")" <<
@@ -155,7 +163,7 @@ void FileWriterPlugin::writeFrame(const Frame& frame) {
 
   // Set the offset
   std::vector<hsize_t>offset(dset.dataset_dimensions.size());
-  offset[0] = frame_offset;
+  offset[0] = frame_offset * outer_chunk_dimension;
 
   uint32_t filter_mask = 0x0;
   ensureH5result(H5DOwrite_chunk(dset.datasetid, H5P_DEFAULT,
@@ -277,9 +285,12 @@ void FileWriterPlugin::createDataset(const FileWriterPlugin::DatasetDefinition& 
   ensureH5result(dataspace, "H5Screate_simple failed to create the dataspace");
 
   /* Enable chunking  */
-  LOG4CXX_INFO(logger_, "Chunking=" << chunk_dims[0] << ","
-                                     << chunk_dims[1] << ","
-                                     << chunk_dims[2]);
+  std::stringstream ss;
+  ss << "Chunking = " << chunk_dims[0];
+  for (int index = 1; index < chunk_dims.size(); index++){
+	  ss << "," << chunk_dims[index];
+  }
+  LOG4CXX_DEBUG(logger_, ss.str());
   prop = H5Pcreate(H5P_DATASET_CREATE);
   ensureH5result(prop, "H5Pcreate failed to create the dataset");
 
@@ -872,7 +883,7 @@ void FileWriterPlugin::configureDataset(OdinData::IpcMessage& config, OdinData::
         throw std::runtime_error("Cannot create a dataset without a data type");
       }
 
-      // There must be dimensions present for the dataset
+      // There may be dimensions present for the dataset
       if (config.has_param(FileWriterPlugin::CONFIG_DATASET_DIMS)) {
         const rapidjson::Value& val = config.get_param<const rapidjson::Value&>(FileWriterPlugin::CONFIG_DATASET_DIMS);
         // Loop over the dimension values
@@ -883,8 +894,10 @@ void FileWriterPlugin::configureDataset(OdinData::IpcMessage& config, OdinData::
         }
         dset_def.frame_dimensions = dims;
       } else {
-        LOG4CXX_ERROR(logger_, "Cannot create a dataset without dimensions");
-        throw std::runtime_error("Cannot create a dataset without dimensions");
+    	  // This is a single dimensioned dataset so store dimensions as NULL
+          dimensions_t dims(0);
+    	  dset_def.frame_dimensions = dims;
+          //throw std::runtime_error("Cannot create a dataset without dimensions");
       }
 
       // There might be chunking dimensions present for the dataset, this is not required
@@ -908,7 +921,11 @@ void FileWriterPlugin::configureDataset(OdinData::IpcMessage& config, OdinData::
         dset_def.compression = no_compression;
       }
 
-      LOG4CXX_INFO(logger_, "Creating dataset [" << dset_def.name << "] (" << dset_def.frame_dimensions[0] << ", " << dset_def.frame_dimensions[1] << ")");
+      if (dset_def.frame_dimensions.size() > 0){
+          LOG4CXX_DEBUG(logger_, "Creating dataset [" << dset_def.name << "] (" << dset_def.frame_dimensions[0] << ", " << dset_def.frame_dimensions[1] << ")");
+      } else {
+          LOG4CXX_DEBUG(logger_, "Creating 1D dataset [" << dset_def.name << "] with no additional dimensions");
+      }
       // Add the dataset definition to the store
       this->nextAcquisition_.dataset_defs_[dset_def.name] = dset_def;
     }
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -895,9 +895,9 @@ void FileWriterPlugin::configureDataset(OdinData::IpcMessage& config, OdinData::
         }
         dset_def.frame_dimensions = dims;
       } else {
-    	  // This is a single dimensioned dataset so store dimensions as NULL
-          dimensions_t dims(0);
-    	  dset_def.frame_dimensions = dims;
+        // This is a single dimensioned dataset so store dimensions as NULL
+        dimensions_t dims(0);
+        dset_def.frame_dimensions = dims;
       }
 
       // There might be chunking dimensions present for the dataset, this is not required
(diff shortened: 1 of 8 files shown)


frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -852,8 +863,10 @@ void FileWriterPlugin::configureDataset(OdinData::IpcMessage& config, OdinData::
         }
         dset_def.frame_dimensions = dims;
       } else {
-        LOG4CXX_ERROR(logger_, "Cannot create a dataset without dimensions");
-        throw std::runtime_error("Cannot create a dataset without dimensions");
+    	  // This is a single dimensioned dataset so store dimensions as NULL
+          dimensions_t dims(0);
+    	  dset_def.frame_dimensions = dims;
+          //throw std::runtime_error("Cannot create a dataset without dimensions");
```
GDYendell: Just delete this instead of commenting.

ajgdls:
This PR has an effect on all detectors so I want to ensure it is fully tested before we accept it.  Once using this filewriter version it is necessary to supply all chunking parameters (eg for 2D dataset of 20 x 30 you would need to supply chunk dims of "1 20 30". I think before you would get away without the "1").
I should be able to check with Percival, perhaps @MattTaylorDLS can check Eiger and @GDYendell Excalibur?

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -142,19 +142,27 @@ void FileWriterPlugin::createFile(std::string filename, size_t chunk_align)
...
   HDF5Dataset_t& dset = this->get_hdf5_dataset(frame.get_dataset_name());
 
   hsize_t frame_offset = 0;
   frame_offset = this->getFrameOffset(frame_no);
-  this->extend_dataset(dset, frame_offset + 1);
+  // We will need to extend the dataset in 1 dimension by the outer chunk dimension
+  // For 3D datasets this would normally be 1 (a 2D image)
+  // For 1D datasets this would normally be the quantity of data items present in a single chunk
+  uint64_t outer_chunk_dimension = this->currentAcquisition_.dataset_defs_[frame.get_dataset_name()].chunks[0];
+  this->extend_dataset(dset, ((frame_offset + 1) * outer_chunk_dimension));
```
GDYendell: Redundant pair of brackets around second argument.

GDYendell:
The simulation for Excalibur doesn't work currently, so I can't test against that.

ajgdls:
OK well there is no rush to take this.  The changes are small and I'm running LATRD against my branch.

ajgdls:
Actually it's failing tests due to memory access violation although I can't see why at the moment...

timcnicholls:
This can be tested with EXCALIBUR on the single-FEM system I have here, or also with the 1M system in the detector lab. The former is easier to set up. I'll run the test if you let me know when it isn't failing.

MattTaylorDLS:
Hm, if you don't supply chunk dims, it segmentation faults. Can we get it to print a suitable error?

ajgdls:
OK I wonder if that is why my test run fails.  Didn't it previously work if you didn't supply chunk dims?  That behaviour should continue so I need to understand how I've broken that.  I'll report back once I figure it out.

MattTaylorDLS:
yeah it did work before if you didn't supply chunk dims. i dont think i've ever specified them

ajgdls:
OK @GDYendell @MattTaylorDLS the updates have been made.

GDYendell:
It looks like this is already supposed to be handled here: https://github.com/dls-controls/odin-data/blob/f8348a8e50cb9c434de437db2ed3a0a7e2f65190/frameProcessor/src/FileWriterPlugin.cpp#L274 but it doesn't work properly?

MattTaylorDLS:
Doesn't crash for me anymore

ajgdls:
I looked at that code when I was investigating this issue.  However, that method is passed a definition and I do not want that check to then write directly back into the current acquisition member variable as it would look quite confusing I think.
I added the check for adding 1 in the configure method.  Also, the code you mention does already add 1, it only retains the result for the duration of the method (The vector is constructed with size 1 value 1 and then the frame dimensions are appended).

GDYendell:
That makes sense; it should be in the configure method.

Does this mean that original block will never be executed now? I.e. `if (definition.chunks.size() != dset_dims.size()) {` will never be true now.

If so we should remove it.

ajgdls:
Well I'm not sure about "never" be executed.  For a start the unit test calls that method directly and so that piece of code could be executed.  Is it not a valid check to have just in case, I think the alternative is unhandled exceptions and perhaps crash?

GDYendell:
It's valid to check, but is the logic for setting the default existing in two places a problem? It could just return a useful error in createDataset.

ajgdls:
Hi I've taken another look at the createDataset method.  Really it should be re-worked as it already assumes the frame_dimensions exist, no check so if they didn't exist it would fail in an odd way I think (creating a single 1D 1 element dataset instead of 3D 1xnxy).  Adding these checks will also make the unit tests fail as they rely on the incorrect behaviour, so the tests should be updated.
I'll make these updates and re-push.

ajgdls:
Actually @GDYendell I can't verify dataset dimensions are non zero because that is exactly what I supply for LATRD!  So I will make the check only for chunk dimensions and then throw and exception.

ajgdls:
Before merging I would still suggest that this be verified with real applications.  Ie Eiger and Excalibur.  So @timcnicholls and @MattTaylorDLS whenever you feel like some testing please go ahead.  Thanks!

ajgdls:
@MattTaylorDLS have you been able to test this version against Eiger yet?  Does it still work for you?

MattTaylorDLS:
I tested it against a fake Eiger when you first put it up, and it was all ok. I didn't test against the real Eiger i dont think as it was in use at the time, but there shouldn't be any differences once it's past the EigerFan

timcnicholls:
@ajgdls Just built and run this PR with EXCALIBUR test system and output data looks fine. But I didn't need to change dataset configuration?? Here's the config file:

```
logconfig=/u/tcn45/develop/projects/install/config/fr_log4cxx.xml
ctrl=tcp://127.0.0.1:5004
ready=tcp://127.0.0.1:5001
release=tcp://127.0.0.1:5002
no-client=true
detector=excalibur
path=/u/tcn45/develop/projects/install/lib
datasets=data 
dtype=1
dims=256
dims=2048
acqid=0
```

and here's a snippet of the output from the FP:
```
:46:39,337  FP.FrameProcessorController DEBUG - Configuration submitted: {"params":{"hdf":{"dataset":{"cmd":"create","name":"data","datatype":1,"dims":[256,2048]}}},"msg_type":"illegal","msg_val":"illegal","timestamp":"2017-10-05T15:46:39.337365"}
15:46:39,337  FP.FileWriterPlugin INFO  - {"params":{"dataset":{"cmd":"create","name":"data","datatype":1,"dims":[256,2048]}},"msg_type":"illegal","msg_val":"illegal","timestamp":"2017-10-05T15:46:39.337550"}
15:46:39,337  FP.FileWriterPlugin DEBUG - Configure dataset
15:46:39,337  FP.FileWriterPlugin DEBUG - Creating dataset [data] (256, 2048)
15:46:39,338  FP.FrameProcessorController DEBUG - Configuration submitted: {"params":{"hdf":{"file":{"name":"test.hdf5","path":"/tmp/"},"frames":0,"acquisition_id":"0","write":true}},"msg_type":"illegal","msg_val":"illegal","timestamp":"2017-10-05T15:46:39.337855"}
15:46:39,338  FP.FileWriterPlugin INFO  - {"params":{"file":{"name":"test.hdf5","path":"/tmp/"},"frames":0,"acquisition_id":"0","write":true},"msg_type":"illegal","msg_val":"illegal","timestamp":"2017-10-05T15:46:39.338118"}
15:46:39,338  FP.FileWriterPlugin DEBUG - Configure file name and path
15:46:39,338  FP.FileWriterPlugin DEBUG - Next file path changed to /tmp/
15:46:39,338  FP.FileWriterPlugin DEBUG - Next file name changed to test.hdf5
15:46:39,338  FP.FileWriterPlugin INFO  - Setting next Acquisition ID to 0
15:46:39,338  FP.FileWriterPlugin INFO  - Creating file: /tmp/test.hdf5
15:46:39,340  FP.FileWriterPlugin DEBUG - Data type: UINT16
15:46:39,340  FP.FileWriterPlugin DEBUG - Chunking = 1,256,2048
15:46:39,340  FP.FileWriterPlugin INFO  - Compression type: None
15:46:39,340  FP.FileWriterPlugin INFO  - Creating dataset: data
15:46:39,340  FP.FileWriterPlugin DEBUG - Closing intermediate open HDF objects
15:46:40,323  FP.FrameProcessorController DEBUG - Meta RX thread called: {"plugin":"hdf","parameter":"createfile","type":"string","header":{"acqID":"0","totalFrames":0}}
15:46:40,324  FW.SharedMemoryController DEBUG - Requesting shared buffer configuration from frame receiver
15:46:40,326  FW.SharedMemoryController DEBUG - RX thread called with message: {"params":{"shared_buffer_name":"ExcaliburFrameBuffer1"},"msg_type":"notify","msg_val":"buffer_config","timestamp":"2017-10-05T15:46:40.325706"}
15:46:40,326  FW.SharedMemoryController DEBUG - Shared buffer config notification received for ExcaliburFrameBuffer1
15:46:40,326  FW.SharedMemoryController DEBUG - Initialised shared buffer manager for buffer ExcaliburFrameBuffer1
15:46:45,022  FW.SharedMemoryController DEBUG - RX thread called with message: {"params":{"frame":0,"buffer_id":3},"msg_type":"notify","msg_val":"frame_ready","timestamp":"2017-10-05T15:46:45.021818"}
15:46:45,022  FW.Frame       TRACE - Frame constructed
15:46:45,022  FW.ExcaliburProcessPlugin TRACE - Reordering frame.
15:46:45,022  FW.ExcaliburProcessPlugin TRACE - Frame size: 1048760
15:46:45,022  FW.ExcaliburProcessPlugin TRACE - Raw frame number: 0
15:46:45,022  FW.ExcaliburProcessPlugin TRACE - Packets received: 132 SOF markers: 2 EOF markers: 2
15:46:45,022  FW.ExcaliburProcessPlugin TRACE - Frame data size: 1048592
15:46:45,028  FW.Frame       TRACE - Frame constructed
15:46:45,028  FW.Frame       TRACE - copy_data called with size: 1048760 bytes
15:46:45,028  FP.DataBlockPool DEBUG - Requesting DataBlock of 1048760 bytes
15:46:45,028  FP.DataBlockPool DEBUG - Allocating 2 additional DataBlocks of 1048760 bytes
15:46:45,028  FP.DataBlock   DEBUG - Constructing DataBlock, allocating 1048760 bytes
15:46:45,028  FP.DataBlock   DEBUG - Constructing DataBlock, allocating 1048760 bytes
15:46:45,028  FP.DataBlockPool DEBUG - Providing DataBlock [id=1]
15:46:45,029  FW.ExcaliburProcessPlugin TRACE - Pushing data frame.
15:46:45,029  FW.Frame       TRACE - Frame destroyed
15:46:45,029  FP.FileWriterPlugin TRACE - Writing frame [0] size [1048760] type [-1] name [data]
15:46:45,029  FP.FileWriterPlugin DEBUG - Raw frame number: 0
15:46:45,029  FP.FileWriterPlugin DEBUG - Frame offset adjustment: 0
15:46:45,029  FP.FileWriterPlugin TRACE - Writing frame offset=0 (0) dset=data
15:46:45,029  FW.Frame       DEBUG - Sending response: {"params":{"frame":0,"buffer_id":3},"msg_type":"notify","msg_val":"frame_release","timestamp":"2017-10-05T15:46:45.029516"}
15:46:45,030  FP.FileWriterPlugin TRACE - Master frame processed
15:46:45,030  FW.Frame       TRACE - Frame destroyed
15:46:45,030  FP.DataBlockPool DEBUG - Releasing DataBlock [id=1]
15:46:45,031  FP.FrameProcessorController DEBUG - Meta RX thread called: {"plugin":"hdf","parameter":"writeframe","type":"string","header":{"acqID":"0"}}
```

Is this consistent with what you were expecting?

ajgdls:
I notice this line, which means the chunking has the correct 1 at the front.

5:46:39,340  FP.FileWriterPlugin DEBUG - Chunking = 1,256,2048

I think that if you didn't specify chunking at all then it is defaulted correctly.

If you had explicitly specified chunking as 256,2048 then it would have failed.  So I'm happy that it is working as expected, especially if you report the correct output.

Shall we merge?

timcnicholls:
Ah yeah of course, I only set the dimensions, not the chunking! 

Happy for you to merge, looks fine to me. Will approve now.

PR #59, opened 2017-08-31
Generic improvements to odin-data

ajgdls:
These changes have been cherry picked from @GDYendell original PR #50.  These improvements are the simple changes we found that are not DLS specific.  Further changes will be applied in another PR, once this has been merged.
This PR does not include the python IPC changes, they will be manually applied and submitted in another PR.


ajgdls:
I have now also applied the specific changes made to the ipc_channel class by @GDYendell , these are the changes that @timcnicholls would want to work from or reapply to the final ipc_channel class wherever it lives.

ajgdls:
I think we need to be aware here that once these changes are accepted Percival will need to apply an offset of 1 because it's frame number doesn't start at 0?  Is that correct @GDYendell  @MattTaylorDLS ?  The offset can be applied during the startup using command line or file config option, again can that be confirmed @GDYendell @MattTaylorDLS ?

timcnicholls:
Are those the only changes @GDYendell ? From what I can see they are only cosmetic whitespace etc...

ajgdls:
Mmm, I've been through the commits and I can only see those.  I'll let Gary verify that.

MattTaylorDLS:
Did we want to include that commit with the frame number offset starting at 0 into odin-detector? Or should we have kept it in dls-controls? I can't remember what we said about it yesterday

GDYendell:
No that isn't all of the changes to the IPC stuff; there should be changes to ipc_message, too. It is a lot of formatting, though.

GDYendell:
> Did we want to include that commit with the frame number offset starting at 0 into odin-detector? Or should we have kept it in dls-controls? I can't remember what we said about it yesterday

We do want that because it is the only way to be able to get 0 and 1 with the current offset method. It will be refactored once when we add rewinding, though.

GDYendell:
> The offset can be applied during the startup using command line or file config option

 The offset cannot currently be set by command line. It is a runtime configuration option. Do we want that added?

MattTaylorDLS:
Ah OK. i guess so yes

ajgdls:
OK I need to find the ipc_message changes and apply those too.

ajgdls:
OK I believe this is the ipc_message.py commit that I had not applied.

timcnicholls:
Thanks @ajgdls - happy for you to merge.

PR #60, opened 2017-09-12
Update FrameProcessor README

GDYendell:
I have updated the docs in the Design.md for the latest code (mostly name changes). The only thing is the classes diagram is still wrong. I have also merged the Design into the README so that it all displays nicely on the GitHub page.

Having this up to date will make it much easier to explain to people how it all works.

(If you view the commits individually, the diff will be much more useful)


GDYendell:
Is the fr setup in master now?

Also, how did you make the class diagram / how would I update it?

ajgdls:
Yes I believe it went in on 17th August PR #52 . I've noticed it running on my LATRD application recently as you get a delayed (1 second I think) notification message once the shared memory handshake has completed.

For the diagram:
https://yuml.me/diagram/scruffy/class/draw

ajgdls:
By the way this PR isn't passing tests, is that because the branch is behind master?

GDYendell:
Fixed it... (whoops)

GDYendell:
Do we want to keep mentions of specific detectors out of the docs? (For the plugins on the class diagram)

ajgdls:
I guess in an ideal world we create an example detector that lives in the odin-data module and can be used as a template.  I would say yes keep specific detectors out of the docs here.

GDYendell:
frameprocessor class hierarchy

ajgdls:
Yes I think that looks great.

GDYendell:
Now edited your one directly I will go with this

PR #61, opened 2017-09-15
Checked FileWriterPlugin is expecting frames before starting writing

MattTaylorDLS:
BC-499


GDYendell:
This looks good, but it would be nice to make it clear that there the two cases for starting writing. If totalFrames == 0 we want to write regardless because we are in free running mode and if not then we are only writing if we are expecting > 0 frames, based on rank.

You could split the if into 3 blocks, but a comment for each would probably be clear enough.

MattTaylorDLS:
OK Maybe i'll add some comments. Think it'd be better not to have three parts of the 'if' blocks when two of them are doing the same thing

MattTaylorDLS:
Yeah ok

MattTaylorDLS:
@timcnicholls, you happy for me to merge this?

PR #62, opened 2017-09-21
Refactored odin-data python into package, added DEALER-ROUTER IpcChannels

timcnicholls:
This PR bundles up two changes:

1) Refactoring of the odin-data tools into an installable odin-data package
2) Addition of DEALER-ROUTER socket pairs to IpcChannel - a demonstration implementation of this can be seen the FrameReceiver client handling code and associated frame_receiver_client test app.


GDYendell:
@MattTaylorDLS can you make the equivalent changes in MetaListener and EigerFan? The clients will also need to be updated.

MattTaylorDLS:
Yeah i can. Will need to wait until its in dls-controls though

ajgdls:
@MattTaylorDLS and @GDYendell it should be possible to remove all duplicate ipc class files after this change.  I will test with LATRD downloading the odin-data python libraries through pip and then we should try to do the same for Eiger and Excalibur.

ajgdls:
I've just been attempting to install this python module as part of my detector module.  I used the entry

-e git+git://github.com/odin-detector/odin-data.git@dealer_router#egg=odin-data&subdirectory=tools/python

in requirements.txt and the installation fails when trying to perform 
python setup.py egg_info

Traceback (most recent call last):
  File "setup.py", line 14, in <module>
    version=versioneer.get_version(),
  File "/home/gnx91527/work/LATRD/tools/python/venv_test/src/odin-data/tools/python/versioneer.py", line 1480, in get_version
    return get_versions()["version"]
  File "/home/gnx91527/work/LATRD/tools/python/venv_test/src/odin-data/tools/python/versioneer.py", line 1453, in get_versions
    pieces = from_vcs_f(cfg.tag_prefix, root, verbose)
  File "/home/gnx91527/work/LATRD/tools/python/venv_test/src/odin-data/tools/python/versioneer.py", line 1110, in git_pieces_from_vcs
    pieces["distance"] = int(count_out)  # total number of commits
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

It looks like you added a file called versioneer.py and that has the trouble.  @timcnicholls have you see this?  Is it possible that this is failing because there is no version for versioneer to read from github?

timcnicholls:
@ajgdls Yes I think so. Try setting a git tag on your current repo, that should fix it.

timcnicholls:
@ajgdls Did you resolve the versioneer issue?

ajgdls:
Hi @timcnicholls I have not yet tried to tag on our fork as I've got some other detector work I had to get finished before everyone leaves for a conference.
I will try this today (right now).

ajgdls:
Hi @timcnicholls yes it worked thanks.  I've already approved it so you'll have to get a 👍 instead.

PR #63, opened 2017-09-28
Generic updates (part 2)

ajgdls:
This PR contains further generic updates to odin-data concerned with logging and some additional helper python classes for use with ZeroMQ.  Specifically:

- Logging updated to work with Graylog.  Default settings should still be the same as before, only if Graylog is specified in the config file will it be used.
- Minor logging typo fixes and updates to logging levels.
- Added ZeroMQ helper python class for communicating with frameProcessor or frameReceiver (will be used by adapters).

Most of these changes were originally made by @GDYendell and @MattTaylorDLS.


ajgdls:
The last commit was just to get the Travis build working.  At DLS the includes must already be present from somewhere else.  @GDYendell are you happy with this?

PR #64, opened 2017-10-04
Added logging configuration helper module.

ajgdls:
This module contains a default logging configuration to make it easy to keep logging consistent across python applications.  It also provides helper functions for adding handlers, and specifically a helper for setting up logging to graylog.


ajgdls:
This should be the last PR from the DLS specific branch present in dls-controls fork of Odin-data.  After this is merged we should have migrated all generic code required over to odin-data.
Note I have tested the logging to a graylog server manually but I'm not sure how to setup a test on Travis without setting up a graylog server on Travis, writing a test that logs a message and then requests the message from graylog.  I'm sure this is possible but I don't want to invest that time right now, and graylog handling is by default not enabled when using this module.

GDYendell:
I think this actually goes some way towards being able to merge back into @ulrikpedersen 's original module and not having to duplicate nearly as much code. If these new methods and the dictionary were combined into a class it would be possible write a small wrapper class in each application, which would just inherit from the base class and make a few add_* calls in the init. Then the main could simply import the child class and call setup_logging.

What do you think?

ajgdls:
It sounds like a reasonable idea, I'm not aware of Ulrik's original module (at least I don't think I am?).  I think it would be nice to have the logging setup within a class, but I think both methods are reasonably clean.  @GDYendell where is the original module?

GDYendell:
https://github.com/dls-controls/python-logging-configuration

ajgdls:
I see, they are extremely similar.  The proper thing to do would probably be for that module to have the custom helper functions, and then install as an egg and it could be pip installed just as we do for other python modules (including odin-detector and this module).  Perhaps we need to look at doing that and discuss with @ulrikpedersen, we can raise a ticket for that. 

So back to this PR.  Do we therefore not add this file to this repository and then Matthew needs to use it in Eiger until we get something merged into the logging module?  Or do we add it here for now until we get something merged into the logging module?  I'm easy either way, it's really up to @GDYendell  and @MattTaylorDLS , I'm guessing as @timcnicholls approved he doesn't mind if it lives here even temporarily.

timcnicholls:
I'm happy for it to live here temporarily. 

If the final solution creates a dependence on the dls-controls python logging module this needs to be widely available (and not DLS specific) since odin-data has usage outside DLS-related projects. If that isn't possible then we either maintain a version here within the odin-data package, or the DLS detector repos that require it import the dependency and inject it into the logging configuration for odin-data applications.

ajgdls:
So currently it is just a logging configuration setup which means it doesn't need to be called at all.  It can be called by any main python applications.  I argue that it might be a useful module and so could / should be made more widely available (not necessarily living in dls-controls or at least publicly available) but I think @ulrikpedersen will make that decision as it is his baby.  Right now, let's add this file to this repo and then I'll raise a ticket to track this conversation (for which we need Ulrik to have his say).  @GDYendell if that is OK with you then if you approve I'll merge this specific PR.

GDYendell:
I think we should merge this for now. I guess to make this work in EigerDAQ we just need to add a few add_* calls?

ajgdls:
Yep simply add a single call to setup the graylog and then call setup_logging in your main.  See my test example below:

```
def main():
    add_graylog_handler("xxxx-xx-xxxx-xx.diamond.ac.uk", 12202)
    setup_logging()
    log = logging.getLogger("test_logconfig")
    log.info("Sending a test logging message...")

```

PR #67, opened 2017-10-06
Fix unsigned current_frame_seen variable in frame decoders

timcnicholls:
Provides underlying compatibility with issue [BC-319](BC-319-current-frame-seen). 

N.B. FrameDecoder plugins copying the pattern used in PERCIVAL and EXCALIBUR decoders will need changes to their associated member variables (current_frame_seen_) and buffer map iterators used in monitor_buffers.


timcnicholls:
Merging this myself since it's an isolated change as discussed.

PR #68, opened 2017-10-18
Allow IPCMessages to be decoded correctly on python 3.5

SophieKSTFC:
Using odin-data on python3.5 (e.g. on embedded system) causes issues decoding IPCMessages received over e.g. ZMQ sockets, since the json library in that version only decodes strings and not bytes or bytearrays.


timcnicholls:
Agreed with @GDYendell - this would make it clearer.

GDYendell:
Cool.

If you want the comments, could you use the format # <Comment> rather than '''<Comment>'''.

PR #71, opened 2017-10-26
Added ability to write blocks of frames in the FileWriterPlugin

MattTaylorDLS:
BC-539

### DIFF
--- a/frameProcessor/include/FileWriterPlugin.h
+++ b/frameProcessor/include/FileWriterPlugin.h
@@ -130,6 +130,7 @@ public:
   void stopAcquisition();
   void startCloseFileTimeout();
   void runCloseFileTimeout();
+  size_t calcNumFrames(size_t totalFrames);
 
 private:
   /** Configuration constant for process related items */
@@ -138,6 +139,8 @@ private:
   static const std::string CONFIG_PROCESS_NUMBER;
   /** Configuration constant for this process rank */
   static const std::string CONFIG_PROCESS_RANK;
+  /** Configuration constant for the number of frames per block */
+  static const std::string CONFIG_PROCESS_BLOCKSIZE;
 
   /** Configuration constant for file related items */
   static const std::string CONFIG_FILE;
@@ -225,6 +228,8 @@ private:
   Acquisition currentAcquisition_;
   /** Details of the next acquisition to be written */
   Acquisition nextAcquisition_;
+  /** Number of frames to write consecutively in a file */
+  size_t frames_per_block;
   /** Timeout for closing the file after receiving no data */
   size_t timeoutPeriod;
   /** Mutex used to make starting the close file timeout thread safe */
--- a/frameProcessor/include/FileWriterPlugin.h
+++ b/frameProcessor/include/FileWriterPlugin.h
@@ -229,7 +229,7 @@ private:
   /** Details of the next acquisition to be written */
   Acquisition nextAcquisition_;
   /** Number of frames to write consecutively in a file */
-  size_t frames_per_block;
+  size_t frames_per_block_;
   /** Timeout for closing the file after receiving no data */
   size_t timeoutPeriod;
   /** Mutex used to make starting the close file timeout thread safe */
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -68,7 +68,7 @@ FileWriterPlugin::FileWriterPlugin() :
     hdf5_fileid_(0),
     hdf5ErrorFlag_(false),
     frame_offset_adjustment_(0),
-	frames_per_block_(1),
+    frames_per_block_(1),
     timeoutPeriod(0),
     timeoutThreadRunning(true),
 	timeoutThread(boost::bind(&FileWriterPlugin::runCloseFileTimeout, this))
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -17,6 +17,7 @@ namespace FrameProcessor
 const std::string FileWriterPlugin::CONFIG_PROCESS             = "process";
 const std::string FileWriterPlugin::CONFIG_PROCESS_NUMBER      = "number";
 const std::string FileWriterPlugin::CONFIG_PROCESS_RANK        = "rank";
+const std::string FileWriterPlugin::CONFIG_PROCESS_BLOCKSIZE   = "frames_per_block";
 
 const std::string FileWriterPlugin::CONFIG_FILE                = "file";
 const std::string FileWriterPlugin::CONFIG_FILE_NAME           = "name";
@@ -67,6 +68,7 @@ FileWriterPlugin::FileWriterPlugin() :
     hdf5_fileid_(0),
     hdf5ErrorFlag_(false),
     frame_offset_adjustment_(0),
+    frames_per_block(1),
     timeoutPeriod(0),
     timeoutThreadRunning(true),
 	timeoutThread(boost::bind(&FileWriterPlugin::runCloseFileTimeout, this))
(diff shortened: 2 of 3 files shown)


frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -450,14 +452,14 @@ size_t FileWriterPlugin::getFrameOffset(size_t frame_no) const {
...
-    if (frame_offset % this->concurrent_processes_ != this->concurrent_rank_) {
+    if ((frame_offset / frames_per_block) % this->concurrent_processes_ != this->concurrent_rank_) {
       LOG4CXX_WARN(logger_,"Unexpected frame: " << frame_no <<
                            " in this process rank: " << this->concurrent_rank_);
       throw std::runtime_error("Unexpected frame in this process rank");
     }
 
     // Calculate the new offset based on how many concurrent processes are running
-    frame_offset = frame_offset / this->concurrent_processes_;
+    frame_offset = ((frame_offset / (frames_per_block * this->concurrent_processes_)) * frames_per_block) + (frame_offset % frames_per_block);
```
GDYendell: Is it possible to split into logical chunks with variable names explaining what each part is? It's very difficult to read this and understand what is going on.

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -798,6 +798,8 @@ void FileWriterPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMess
       if (nextAcquisition_.totalFrames_ > 0 && nextAcquisition_.framesToWrite_ == 0) {
     	// We're not expecting any frames, so just clear out the nextAcquisition for the next one and don't start writing
     	this->nextAcquisition_ = Acquisition();
+    	this->currentAcquisition_ = Acquisition();
+        framesWritten_ = 0;
```
GDYendell: There are tabs/spaces problems here.

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -855,6 +857,23 @@ void FileWriterPlugin::configureProcess(OdinData::IpcMessage& config, OdinData::
       LOG4CXX_DEBUG(logger_, "Process rank is already " << this->concurrent_rank_);
     }
   }
+
+  // Check to see if the frames per block is being set
+  if (config.has_param(FileWriterPlugin::CONFIG_PROCESS_BLOCKSIZE)) {
+    size_t blocksize = config.get_param<size_t>(FileWriterPlugin::CONFIG_PROCESS_BLOCKSIZE);
```
GDYendell: block_size

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -855,6 +857,23 @@ void FileWriterPlugin::configureProcess(OdinData::IpcMessage& config, OdinData::
...
+    if (this->frames_per_block != blocksize) {
+      // If we are writing a file then we cannot change concurrent rank
```
GDYendell: Comment copy-paste error.

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -1247,4 +1266,36 @@ void FileWriterPlugin::runCloseFileTimeout()
...
+ * \param[in] totalFrames - The total number of frames in the acquisition
+ * \return - The number of frames that this FileWriter is expected to write
+ */
+size_t FileWriterPlugin::calcNumFrames(size_t totalFrames)
+{
+	size_t result = 0;
+
+	// Work out how many 'rounds' where all processes are writing whole blocks
+	double blocksNeeded = totalFrames / frames_per_block;
+	size_t numWholeRoundsNeeded = blocksNeeded / this->concurrent_processes_;
```
GDYendell: It would be good to do some explicit rounding here.

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -1247,4 +1266,36 @@ void FileWriterPlugin::runCloseFileTimeout()
...
+	result = numWholeRoundsNeeded * frames_per_block;
+
+	// Now work out if there is any left over half-complete rounds
```
GDYendell: ...are any...

frameProcessor/src/FrameProcessorApp.cpp
```diff
@@ -108,6 +108,8 @@ void parse_arguments(int argc, char** argv, po::variables_map& vm, LoggerPtr& lo
            "Set the Acquisition Id of the acquisition")
         ("timeout",       po::value<size_t>()->default_value(0),
            "Set the timeout period for closing the file (milliseconds)")
+        ("blocksize",       po::value<size_t>()->default_value(1),
```
GDYendell: block-size

frameProcessor/include/FileWriterPlugin.h
```diff
@@ -225,6 +228,8 @@ class FileWriterPlugin : public FrameProcessorPlugin
   Acquisition currentAcquisition_;
   /** Details of the next acquisition to be written */
   Acquisition nextAcquisition_;
+  /** Number of frames to write consecutively in a file */
+  size_t frames_per_block;
```
GDYendell: Should be frames_per_block_ as it's private.

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -1247,4 +1266,36 @@ void FileWriterPlugin::runCloseFileTimeout()
...
+	{
+		remaining = leftover - (this->concurrent_rank_ * frames_per_block);
+	    if (remaining > 0) {
+	    	result += std::min(remaining, frames_per_block);
+	    }
+	}
+
+    return result;
+}
+
```
GDYendell: There are a whole load of tabs in this function.

GDYendell:
I found my example of a similar thing to calcNumFrames. I used base_frames and remaining, rather than wholeRounds and leftover. Not sure if that is more or less clear.

I think 'result' is slightly unclear, though.

ajgdls:
Can the whole file be re-formatted in a single shot to eliminate tabs?
I agree we should try and tidy the file to be consistent with its use of variables, either camelCase or otherwise.  Perhaps when @MattTaylorDLS removes the file writing specific code into its own class is the best time to make these widespread changes.

GDYendell:
> Can the whole file be re-formatted in a single shot to eliminate tabs?

I did that a short time ago, but Eclipse keeps sneaking them back in.

ajgdls:
@GDYendell votes for clion?

timcnicholls:
Eclipse can re-format in a single shot against a particular profile (e.g. no tabs, 4 spaces). It can be set up to use spaces and not tabs also.

timcnicholls:
Let's not be prescriptive about choice of IDE, esp since e.g. CLion is a paid product. Let's just specify a code style convention (no tabs, X spaces), underscores vs camelCase etc and ensure any editor used conforms. I'm not aware of an IDE that can't be set up to match.

ajgdls:
Yes I was only joking really, you'll never get me away from eclipse.  It certainly is possible to setup eclipse, we just need to make sure it is done (that is a note to myself for today).

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -68,7 +68,7 @@ FileWriterPlugin::FileWriterPlugin() :
     hdf5_fileid_(0),
     hdf5ErrorFlag_(false),
     frame_offset_adjustment_(0),
-    frames_per_block(1),
+	frames_per_block_(1),
```
GDYendell: Tabs?

MattTaylorDLS:
I have set up my eclipse to do it, but  there's time where it ignores that and still puts in tabs.

timcnicholls:
I'd be happy to get away from eclipse too :wink:

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: Yep. This is a case where eclipse has put it back in grrrr!

PR #73, opened 2017-11-01
Ipc adapter

ajgdls:
Status is read from all clients at a configurable rate.
All status is stored in dict along with arguments and a few other useful items (like connected).
GETs can be made on any dict or drilled down into a dict.
If multiple clients are configured then a GET will return array of values.  None returned for non-connected clients.
PUT can either contain an array of values, or a single value which will be fanned out to all clients.
Introspection of status items possible.

### DIFF
--- a/tools/python/odin_data/ipc_client.py
+++ b/tools/python/odin_data/ipc_client.py
@@ -3,8 +3,8 @@ from threading import RLock
 
 import zmq
 
-from ipc_channel import IpcChannel
-from ipc_message import IpcMessage, IpcMessageException
+from odin_data.ipc_channel import IpcChannel
+from odin_data.ipc_message import IpcMessage, IpcMessageException
 
 
 class IpcClient(object):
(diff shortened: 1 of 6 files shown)


timcnicholls:
@ajgdls since our discussion yesterday I realised this approach to the IpcAdapter has a major drawback: the entire Tornado server is blocked while the IpcClient ZMQ poll() waits on the response from the downstream process. This means it will be unresponsive to client requests or any background processing by other adapters. 

What this should really do is, in IPCClient, use the IOLoop pattern and callbacks on the IpcChannel in order to allow the Tornado server to continue to service other requests. pyzmq already allows [integration with Tornado IOLoops](http://pyzmq.readthedocs.io/en/latest/eventloop.html) and odin-data anticipates this by using the ZMQ IOloop instance if available.

I should have spotted this sooner sorry. I will provide a demonstration version of this to integrate into your PR.

ajgdls:
Oh of course good spot.  I can implement that change to avoid blocking the tornado server.

ajgdls:
@timcnicholls I have made a first implementation of non-blocking message send and receive.

timcnicholls:
@ajgdls Thanks! I'll take a look. Just finishing up the FR client control branch so this would be perfect to test it with. I'll have a crack tomorrow.

ajgdls:
Please note @GDYendell and @timcnicholls I have now pushed updates to this PR that allow the adapter to work with the new config and status requests.

PR #74, opened 2017-11-15
Bc540 file writer plugin refactor

MattTaylorDLS:
Not sure if we want this merged into odin-detector at this point, but this pull request is mainly to allow a review of the refactor...

As we're hoping to get this released in time for the next run (in less than a week!), then propt reviews are appreciated!

N.B. If and when this is merged, then the various detector File Processor plugins will need to be changed to match the various method renames and signatures.

### DIFF
--- a/frameProcessor/src/Acquisition.cpp
+++ b/frameProcessor/src/Acquisition.cpp
@@ -36,7 +36,7 @@ Acquisition::~Acquisition() {
 /**
  * Processes a frame
  *
- * This method checks that the frame is valid before using an HDF5 file writer to write the frame to file
+ * This method checks that the frame is valid before using an HDF5File to write the frame to file
  *
  * \param[in] frame - The frame to process
  * \return - The Status of the processing.
@@ -47,16 +47,9 @@ ProcessFrameStatus Acquisition::process_frame(boost::shared_ptr<Frame> frame) {
   try {
     if (check_frame_valid(frame)) {
 
-      uint64_t outer_chunk_dimension = 1;
-      if (dataset_defs_.size() != 0) {
-        outer_chunk_dimension = dataset_defs_.at(frame->get_dataset_name()).chunks[0];
-      }
-
       hsize_t frame_no = frame->get_frame_number();
 
-      size_t frame_offset = 0;
-
-      frame_offset = this->adjust_frame_offset(frame_no);
+      size_t frame_offset = this->adjust_frame_offset(frame_no);
 
       if (this->concurrent_processes_ > 1) {
         // Check whether this frame should really be in this process
@@ -66,16 +59,21 @@ ProcessFrameStatus Acquisition::process_frame(boost::shared_ptr<Frame> frame) {
         }
       }
 
-      boost::shared_ptr<HDF5FileWriter> writer = this->get_file_writer(frame_offset);
+      boost::shared_ptr<HDF5File> file = this->get_file(frame_offset);
 
-      if (writer == 0) {
-        LOG4CXX_ERROR(logger_,"Unable to get writer for this frame");
+      if (file == 0) {
+        LOG4CXX_ERROR(logger_,"Unable to get file for this frame");
         return status_invalid;
       }
 
       size_t frame_offset_in_file = this->get_frame_offset_in_file(frame_offset);
 
-      writer->write_frame(*frame, frame_offset_in_file, outer_chunk_dimension);
+      uint64_t outer_chunk_dimension = 1;
+      if (dataset_defs_.size() != 0) {
+        outer_chunk_dimension = dataset_defs_.at(frame->get_dataset_name()).chunks[0];
+      }
+
+      file->write_frame(*frame, frame_offset_in_file, outer_chunk_dimension);
 
       // Send the meta message containing the frame written and the offset written to
       rapidjson::Document document;
@@ -115,10 +113,10 @@ ProcessFrameStatus Acquisition::process_frame(boost::shared_ptr<Frame> frame) {
       // or if no master frame has been defined. If either of these conditions
       // are true then increment the number of frames written.
       if (master_frame_ == "" || master_frame_ == frame->get_dataset_name()) {
-        size_t dataset_frames = writer->get_dataset_frames(frame->get_dataset_name());
+        size_t dataset_frames = file->get_dataset_frames(frame->get_dataset_name());
         frames_processed_++;
         LOG4CXX_TRACE(logger_, "Master frame processed");
-        size_t current_file_index = file_writer->get_file_index() / concurrent_processes_;
+        size_t current_file_index = current_file->get_file_index() / concurrent_processes_;
         size_t frames_written_to_previous_files = current_file_index * frames_per_block_ * blocks_per_file_;
         size_t total_frames_written = frames_written_to_previous_files + dataset_frames;
         if (total_frames_written == frames_written_) {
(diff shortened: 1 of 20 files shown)


MattTaylorDLS:
The refactor in short:

Separated FileWRiterPlugin into 3 classes - FileWriterPlugin, Acquisition, HDF5FileWriter.

FWP mainly deals with configuration now, and creates the acquisition objects. It just passes frames on to the acquisition object to deal with
Acquistion is created per acquisition, and does the main processing, offsetting, checking etc. It has up totwo HDF5FileWriters at any time, with open files to write to.
HDF5FileWriter contains all of the hdf5 calls for creating, writing, closing files.

frameProcessor/src/Acquisition.cpp
```diff
@@ -0,0 +1,485 @@
...
+  size_t file_index = get_file_index(frame_offset);
+
+  // Get the file writer for this frame index
+  if (file_index == file_writer->get_file_index()) {
+    return this->file_writer;
+  } else if (previous_file_writer != 0 && file_index == previous_file_writer->get_file_index()) {
+    return this->previous_file_writer;
+  } else if (file_index > file_writer->get_file_index()) {
+    LOG4CXX_TRACE(logger_,"Creating new file as frame " << frame_offset <<
+        " won't go into fileIndex " << file_writer->get_file_index() << " as it requires " << file_index);
```
GDYendell: file_index, or even just file index

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -20,6 +21,7 @@ const std::string FileWriterPlugin::CONFIG_PROCESS             = "process";
 const std::string FileWriterPlugin::CONFIG_PROCESS_NUMBER      = "number";
 const std::string FileWriterPlugin::CONFIG_PROCESS_RANK        = "rank";
 const std::string FileWriterPlugin::CONFIG_PROCESS_BLOCKSIZE   = "frames_per_block";
+const std::string FileWriterPlugin::CONFIG_PROCESS_BLOCKS_PER_FILE   = "blocks_per_file";
```
GDYendell: Pad other definitions

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -535,118 +112,40 @@ void FileWriterPlugin::extend_dataset(HDF5Dataset_t& dset, size_t frame_no) cons
...
-      // Check if the frame has defined subframes
-      if (frame->has_parameter("subframe_count")) {
-        // The frame has subframes so write them out
-        this->writeSubFrames(*frame);
-      } else {
-        // The frame has no subframes so write the whole frame
-        this->writeFrame(*frame);
-      }
+  // Start a new acquisition if the frame object contains a different acquisition ID from the current one
+  bool can_process_frame = check_acquisition_id(frame);
```
GDYendell: Could you change check_acquisition_id to frame in acquisition and then you don't have to do this assignment for clarity.

frameProcessor/src/Acquisition.cpp
```diff
@@ -0,0 +1,485 @@
...
+ * Creates a file
+ *
+ * This method creates a new HDF5FileWriter object with the given file_number.
+ * The file will be created, the datasets populated within the file, and a meta message sent
+ *
+ * \param[in] file_number - The file_number to create a file for
+ */
+void Acquisition::create_file(size_t file_number) {
+  // Set previous writer to current writer
+  close_file(previous_file_writer);
```
GDYendell: This syntax seems a little strange... Should close_file not take a file rather than a file_writer? Or maybe it show be file_writer->close_file?

frameProcessor/src/Acquisition.cpp
```diff
@@ -0,0 +1,485 @@
...
+  previous_file_writer = file_writer;
+
+  file_writer = boost::shared_ptr<HDF5FileWriter>(new HDF5FileWriter());
+
+  // Create the file
+  boost::filesystem::path full_path = boost::filesystem::path(file_path_) / boost::filesystem::path(filename_);
+  file_writer->create_file(full_path.string(), file_number);
```
GDYendell: I see. It does, but needs to do other stuff within the class as well. Adding longer comment in summary...

frameProcessor/src/Acquisition.cpp
MattTaylorDLS: Yes it needs to do a bit more than just calling close_file on the FileWriter. It's called close_file here to mirror the "create_file" method. Maybe it could be named better

frameProcessor/src/Acquisition.cpp
MattTaylorDLS: ok

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: ok

MattTaylorDLS:
Yes, i was toying with the idea of calling the new file something like just File, but it is more of a writer, than a class containg File attributes etc.

It's a bit tricky as it is one File per HDF5FileWriter, so you can do just treat the Writer as a file, but i'll see what names i can come up with that might help

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: not sure what you mean here! Do you mean just put the function call in the if ?

GDYendell:
I think it would be nicer for it to be called File within the context of FileWriterPlugin owning it. Maybe some renaming of things on the FileWriter would make the name File more logical.

frameProcessor/src/FileWriterPlugin.cpp
GDYendell: Yes. Sorry!

MattTaylorDLS: ok. i've renamed it and put it in the if

MattTaylorDLS:
hmm ok i'll change it to HDF5File

MattTaylorDLS:
I have a change to make it possible to specify whether to use the earliest version of HDF5 to create a file, to make it backward compatible. I'll need that change on this branch if we're going to release it from there for this run. So should i commit it to this branch now? It will mean that if we do decide to accept this pull request into odin-detector, it will get included too

GDYendell:
We can do it in a separate PR and if it doesn't look right then it won't go in just yet. It would be good to get this in.

Are we ready to merge this now?

MattTaylorDLS:
would be good to get alans input and then it's good to merge if we're happy to have it in odin-detector

ajgdls:
I'll try hard to get it reviewed this afternoon.

MattTaylorDLS:
> Currently the Acquisition check_frame_valid method will still suffer the same issues with dimensions causing crashes due to the log message assuming certain dimensionality.

This is true, it will be caught by the new try catch now however so the application shouldn't crash, just the frame wont be processed and an error message will be logged.

> All MetaMessagePublisher methods are public
They were public before when they were in FrameProcessorPlugin. so i maintained that visibility.

> Regarding todo comment in FileWriterPlugin, we should add some form of health state and an error message to the status I think
Yes i agree, and i think we discussed that in a meeting a week or two ago, but i think we said it was outside the scope of this refactor. (That TODO has been in a long time)

> Acquisition class is using hard-coded strings which I think can be problematic
Ok, i'll put in statics

PR #76, opened 2017-11-21
Implemented option to use earlier version of hdf5 library to write

MattTaylorDLS:
Note that the default behavior is to use the latest. I.e. it is backwards compatible, you have to specify to use the earliest version

### DIFF
--- a/frameProcessor/include/Acquisition.h
+++ b/frameProcessor/include/Acquisition.h
@@ -56,6 +56,8 @@ public:
   std::string file_path_;
   /** Name of the file to write to */
   std::string filename_;
+  /** Use the earliest version of hdf5 */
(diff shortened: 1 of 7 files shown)


MattTaylorDLS:
Hi yeah I have. The call fails if you specify a max of earliest_version. 

From the HDF5 pages: "Currently, high must be set to the pre-defined value H5F_LIBVER_LATEST. H5F_LIBVER_LATEST corresponds to the version of the HDF5 Library in use. "

Seemed odd to me too, but it is right. I've tested opening a file with an earlier version of h5dump and it's ok

ajgdls:
OK no problems, if you've checked it then that's fine.

PR #79, opened 2017-12-12
First simple implementation of requesting current configuration.

ajgdls:


frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -284,6 +284,30 @@ void FileWriterPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMess
...
+  boost::lock_guard<boost::recursive_mutex> lock(mutex_);
+
+  // Return the configuration of the file writer plugin
+  std::string process_str = get_name() + "/" + FileWriterPlugin::CONFIG_PROCESS + "/";
+  reply.set_param(process_str + FileWriterPlugin::CONFIG_PROCESS_NUMBER, concurrent_processes_);
+  reply.set_param(process_str + FileWriterPlugin::CONFIG_PROCESS_RANK, concurrent_rank_);
+  reply.set_param(process_str + FileWriterPlugin::CONFIG_PROCESS_BLOCKSIZE, frames_per_block_);
+  reply.set_param(process_str + FileWriterPlugin::CONFIG_PROCESS_BLOCKS_PER_FILE, blocks_per_file_);
+
+  std::string file_str = get_name() + "/" + FileWriterPlugin::CONFIG_FILE + "/";
```
MattTaylorDLS: Whilst it's true that when you configure the file params, they get put into the "next_acquisition_" member, as soon as the acquisition is started, they'll be moved to the "current_acquisition_" and so these will no longer have meaningful values. Do we want to handle this? (i.e. give both the current and next?)

frameProcessor/src/FrameProcessorController.cpp
```diff
@@ -107,6 +111,13 @@ void FrameProcessorController::handleCtrlChannel()
       LOG4CXX_DEBUG(logger_, "Control thread reply message (configure): "
                              << replyMsg.encode());
     }
+    else if ((ctrlMsg.get_msg_type() == OdinData::IpcMessage::MsgTypeCmd) &&
+    		 (ctrlMsg.get_msg_val() == OdinData::IpcMessage::MsgValCmdRequestConfiguration)) {
```
MattTaylorDLS: tab

ajgdls:
Thanks @MattTaylorDLS fixed layout.  Regarding the specific FileWriter plugin case, I think we should not be returning both current and next, only the configuration items that get set.  If when we send the configuration it goes into the next acquisition then that is all we return.
If this value is then sucked into current acquisition I think that should be a status item, and it is valid to have the next acquisition provide empty responses at this stage.  We should ensure that the response is empty values and not dangling/undefined values - I assume that is the case?

ajgdls:
OK but that is a different PR to this one.  We should probably sit down and go through all status items to ensure we are providing what is needed.  For example some buffer statistics would also be helpful (how big are our own internal stores, how many in use etc).

MattTaylorDLS:
Yeah, whenever the next becomes current, next is given a new value. And the constructor initialises everything (except the strings which'll get default empty value).

Agree the stats should be a separate PR

PR #80, opened 2017-12-14
Added file writer config options for chunk alignment

MattTaylorDLS:
BC-589 - Ability to set hdf5 chunk threshold and value per file

### DIFF
--- a/frameProcessor/include/Acquisition.h
+++ b/frameProcessor/include/Acquisition.h
@@ -34,7 +34,15 @@ public:
   ProcessFrameStatus process_frame(boost::shared_ptr<Frame> frame);
   void create_file(size_t file_number=0);
   void close_file(boost::shared_ptr<HDF5File> file);
-  void start_acquisition(size_t concurrent_rank, size_t concurrent_processes, size_t frame_offset_adjustment, size_t frames_per_block, size_t blocks_per_file);
+  void start_acquisition(
+      size_t concurrent_rank,
+      size_t concurrent_processes,
+      size_t frame_offset_adjustment,
+      size_t frames_per_block,
+      size_t blocks_per_file,
+      bool use_earliest_hdf5,
+      size_t alignment_threshold,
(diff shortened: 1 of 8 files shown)


MattTaylorDLS:
Alan included in review so that he's aware, but as he's away i guess he won't see this until the new year.

GDYendell:
I don't know what the protocol is for reviewing commit messages, but you have a typo that you might want to change.

MattTaylorDLS:
oops! I'll correct it at the end and squish merge

MattTaylorDLS:
Should have mentioned, this now changes the default alignment from 4mb (with a threshold of 65536) to 1byte and 1byte threshold which is no alignment. (This is the hdf5 library default value if alignment hadn't been set)

ajgdls:
Are you able to run performance tests using the defaults, are these now unavoidable without a config file? I think it's probably OK to enforce the configuring of these if you want good performance we just need to make it clear in the documentation (when we finally write some)

MattTaylorDLS:
Yes i've run performance tests on the Eiger, writing over GPFS and there's no performance hit.

These are 'per acquisition' configurations so to change them, you need to specify the values when you configure the FrameProcessor to create the acquisition (i.e. you can't do it from the command line arguments).

ajgdls:
Oh OK I see.

MattTaylorDLS:
(that's the same as the hdf5 version config option that was put in a couple of weeks ago).

ajgdls:
Recording what @MattTaylorDLS and I just spoke about:  I believe certain parameters (hdf5 version, alignment, threshold) should not be set for each acquisition.  Instead they can be set ad-hoc and the values are then used for all subsequent acquisitions until they are changed again.  This means that the parameters can be configured as part of the application startup and then remain in place.  These kinds of parameters will not likely be changed very often and so it is not necessary to force them to be set for each acquisition.

PR #83, opened 2018-01-15
Implemented the ability to GET or PUT a single indexed item.

ajgdls:
Implemented the ability to GET or PUT a single indexed item from the list of connected processes.  The final value of the URI can now be set as an integer index.  This will then return a single item.  If no index is specified then the whole array is either set or returned.

### DIFF
--- a/tools/python/odin_data/odin_data_adapter.py
+++ b/tools/python/odin_data/odin_data_adapter.py
@@ -102,10 +102,37 @@ class OdinDataAdapter(ApiAdapter):
                 request_items = request_command.split('/')
                 logging.debug(request_items)
                 # Now we need to traverse the parameters looking for the request
+                client_index = -1
+                # Check to see if the URI finishes with an index
+                # If it does then we are only interested in reading that single client
+                try:
+                    index = int(request_items[-1])
+                    if index >= 0:
+                        # This is a valid index so remove the value from the URI
+                        request_items = request_items[:-1]
+                        # Set the client index for submitting config to
+                        client_index = index
+                        logging.debug("URI without index: %s", request_command)
+                except ValueError:
+                    # This is OK, there is simply no index provided
+                    pass
 
-                response_items = []
-                for client in self._clients:
-                    paramset = client.parameters
+                if client_index == -1:
+                    response_items = []
+                    for client in self._clients:
+                        paramset = client.parameters
+                        logging.debug("Client paramset: %s", paramset)
+                        try:
+                            item_dict = paramset
+                            for item in request_items:
+                                item_dict = item_dict[item]
+                        except KeyError, ex:
+                            logging.debug("Invalid parameter request in HTTP GET: %s", request_command)
+                            item_dict = None
+                        response_items.append(item_dict)
+                else:
+                    response_items = []
+                    paramset = self._clients[client_index].parameters
                     logging.debug("Client paramset: %s", paramset)
                     try:
                         item_dict = paramset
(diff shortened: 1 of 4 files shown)


tools/python/odin_data/odin_data_adapter.py
```diff
@@ -114,7 +141,7 @@ def get(self, path, request):
                     except KeyError, ex:
                         logging.debug("Invalid parameter request in HTTP GET: %s", request_command)
                         item_dict = None
-                    response_items.append(item_dict)
+                    response_items = item_dict
```
GDYendell: Could the code 124-131 and 136-143 be a function that takes paramset and returns item_dict?

tools/python/odin_data/odin_data_adapter.py
```diff
@@ -150,6 +177,23 @@ def put(self, path, request):  # pylint: disable=W0613
         if request_command.startswith("config/"):
             request_command = request_command.replace("config/", "", 1)  # Take the rest of the URI
             logging.debug("Configure URI: %s", request_command)
+            client_index = -1
+            # Check to see if the URI finishes with an index
+            # eg hdf5/frames/0
+            # If it does then we are only interested in setting that single client
+            uri_items = request_command.split('/')
```
GDYendell: Could this use the same (more similar) variable name as the request_items above in get?

tools/python/odin_data/odin_data_adapter.py
```diff
@@ -150,6 +177,23 @@ def put(self, path, request):  # pylint: disable=W0613
...
+            # Check for an integer
+            try:
+                index = int(uri_items[-1])
+                if index >= 0:
+                    # This is a valid index so remove the value from the URI
+                    request_command = request_command.replace("/"+uri_items[-1], "", 1)
```
GDYendell: Spaces around the '+'.

I'm also not 100% sure this is safe. If there is a "/<index>" earlier in the string, it will be replaced instead. It needs to replace from the end really as opposed to the other instance of this where it replaces from the start.

ajgdls: Agreed, although it's unlikely to be repeated the code should be replacing from the back of the string, I'll update...

tools/python/odin_data/odin_data_adapter.py
ajgdls: yes

tools/python/odin_data/odin_data_adapter.py
GDYendell: I have some nice ways to remove prefix and suffix using regex, would you like me to push them?

ajgdls: Yes of course go for it.

ajgdls:
We should add a test with a /index in the middle of the string as well as at the end, I can do that when I get back to this.

GDYendell:
I've  just noticed there are already some tests and this is inconsistent with them.

GDYendell:
We would need a different function to remove from the middle. I thought we weren't going down that route?

ajgdls:
OK @GDYendell updates made.

GDYendell:
Looks good. Three things:

1. Use remove_prefix instead of replace. It's much clearer.
2. Use self instead of OdinDataAdapter to call static methods.
3. Add spaces around the '+' on line 173

ajgdls:
Why would you call a static method with self?  That is confusing if you are reading the code.

GDYendell:
It's just a method that belongs to the class, but has no access to the parameters. It specifies that the method will not alter the object state.

ajgdls:
I'm not sure you answered my question.  You appear to have described a static method to me.

GDYendell:
https://realpython.com/blog/python/instance-class-and-static-methods-demystified/

ajgdls:
Nudge @GDYendell applied fix for (1) and (3).

GDYendell:
Just realised, if you subclass and override the statics, the base class get and put will still use the base class statics. I'm not sure that matters here.

ajgdls:
Coming from my C++ I would say overriding statics is :-1:

PR #84, opened 2018-01-18
Update datasets to accept changes.

ajgdls:
Allow creation of dataset without the need for a "command" to create.
Allow parameters of dataset to be specified when created, or set at a later time.
Return dataset configuration as config request items, not status items.

### DIFF
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -31,8 +31,6 @@ const std::string FileWriterPlugin::CONFIG_FILE_NAME                   = "name";
 const std::string FileWriterPlugin::CONFIG_FILE_PATH                   = "path";
 
 const std::string FileWriterPlugin::CONFIG_DATASET                     = "dataset";
-const std::string FileWriterPlugin::CONFIG_DATASET_CMD                 = "cmd";
-const std::string FileWriterPlugin::CONFIG_DATASET_NAME                = "name";
 const std::string FileWriterPlugin::CONFIG_DATASET_TYPE                = "datatype";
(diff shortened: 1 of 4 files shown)


frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -466,90 +517,89 @@ void FileWriterPlugin::configure_file(OdinData::IpcMessage& config, OdinData::Ip
...
+  LOG4CXX_DEBUG(logger_, "Configuring dataset [" << dataset_name << "]");
+
+//  std::string dataset_name;
+//  // There must be a name present for the dataset
+//  if (config.has_param(FileWriterPlugin::CONFIG_DATASET_NAME)) {
+//    dataset_name = config.get_param<std::string>(FileWriterPlugin::CONFIG_DATASET_NAME);
+//  } else {
+//    LOG4CXX_ERROR(logger_, "Cannot configure a dataset without a name");
+//    throw std::runtime_error("Cannot configure a dataset without a name");
+//  }
```
GDYendell: Could you get rid of this commented block?

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -573,6 +623,8 @@ void FileWriterPlugin::status(OdinData::IpcMessage& status)
   status.set_param(get_name() + "/processes", (int)this->concurrent_processes_);
   status.set_param(get_name() + "/rank", (int)this->concurrent_rank_);
 
+  /*
+   * TODO: Remove this section once configuration requests have been verified.
```
MattTaylorDLS: How/when are these going to be verified?

PR #85, opened 2018-02-07
Fix bug in FileWriterPlugin to do with configuring no frames

MattTaylorDLS:
Fixes an issue where if a file writer was configured and wasn't going to
receive any frames then it would clear the current acquisition even if
it wasn't finished

### DIFF
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -311,7 +311,9 @@ void FileWriterPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMess
       if (next_acquisition_->total_frames_ > 0 && next_acquisition_->frames_to_write_ == 0) {
         // We're not expecting any frames, so just clear out the nextAcquisition for the next one and don't start writing
         this->next_acquisition_ = boost::shared_ptr<Acquisition>(new Acquisition());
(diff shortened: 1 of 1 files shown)


MattTaylorDLS:
Well I guess it could, but as this acquisition isn't actually going to be handled by the processor, it seemed more logical just to completely throw it away. Otherwise, the status would be reported with a 'half-filled' in acquisition. e.g. it would have an acquisition id, but have no frames, never be writing, and the filename would be blank.

MattTaylorDLS:
I think at the time i remember there being another reason it had to be cleared but can't remember why now, and it might have been overtaken by the changes made to the FileWriterPlugin.

ajgdls:
So just to confirm, this situation is where you have configured the file writer, and then without starting the writer you configure it again with a different setup?

MattTaylorDLS:
This fix is to fix the situation where:

You have configured an acquisition and have started capturing, but don't get all of the frames. Then, you get a configuration for another acquisition, but this acquisition contains no frames to be sent to you (i.e. there are only 10 frames and so all 10 would be processed by an earlier FileWriter in the rank chain). However, this bug caused the current acquisition to be nulled, and so it never properly closed down the file sending out the meta message to say it had closed the file. This fix doesn't null the current acquisition in this case if it's still writing, and so when you do get the next frames from another acquisition, it will close it down properly and start the new acquisition.

PR #87, opened 2018-02-08
Restrict Work Queue size to prevent unlimited memory use

MattTaylorDLS:
BC-563


MattTaylorDLS:
@GDYendell @ajgdls are you happy with this?

PR #89, opened 2018-02-14
Fix for IpcMessage problem introduced with changes to dataset creation

MattTaylorDLS:
This fixes a memory corruption issue that occurs when configuring a dataset, stemming from a problem in the IpcMessage get_param() call. The underlying issue in the IpcMessage should be fixed long term, this PR is just to fix the immediate issue in dataset creation.

BC-645


ajgdls:
@GDYendell it only appears to be a problem when creating a new paramset from the current paramset using multi levelled paths.  In the example you gave the endpoint is a simple string and not a complex rapid object.
The object is deep copied but somehow the recursive nature of the call appears to corrupt the copying.

PR #92, opened 2018-02-26
Do not shut down the reactor when SharedMemoryController is destroyed.

ajgdls:
This stops the application from being reconfigured for a new buffer setup.


ajgdls:
Yes @GDYendell as far as I can tell.  The re-configuring was resulting in the current SharedMemoryController destruction.  This was stopping the reactor, which we do not want.  Probably when this was first implemented we only destroyed this object during shut down of the application.

PR #94, opened 2018-03-02
Fix block_size=0 bug and add timeout_active to status

GDYendell:
For BC-623 and BC-624


GDYendell:
@ajgdls I would like to merge this after your PR goes int, if you have no issues.

ajgdls:
No issues from me.  You probably don't want to wait for @MattTaylorDLS to return I would just go ahead.

PR #95, opened 2018-03-03
BC-668: inject static fields into graylog configuration

ulrikpedersen:
This help fix [BC-668](https://jira.diamond.ac.uk/browse/BC-668)

We would still need to actually add the static fields in the main() invocation...


ulrikpedersen:
argg! shouldn't have created this PR. Ignore!

PR #98, opened 2018-03-22
Add static fields to graylog handler

MattTaylorDLS:
BC-668


MattTaylorDLS:
I think the codeclimate repo owner can change the settings that dictate what's an error and isn't

ulrikpedersen:
Proposal to overcome the codeclimate moan: remove the arg debug=True. This flag is not truly a 'debug' flag as it does not affect the log-level in any way - it only adds a certain set of meta-data like file and line no. to all log messages sent to graylog (regardless of log-level). We always want that enabled so can hardcode that parameter to 'True'

PR #99, opened 2018-03-27
Added handling of configuration requests to frame receiver

timcnicholls:


ajgdls:
Hi @timcnicholls thanks for this.  I will build it in and check it as soon as I have completed the initial tests with the detector.

timcnicholls:
@ajgdls No worries. Don't forget I'm on leave after today until 9th of Apr. So if you need it merged before that, feel free to go ahead when happy.

ajgdls:
Sorry @timcnicholls just noticed your message.  OK I have already an update to fix this problem.  I can't push it right now because it is on a different (polluted) branch but I can push and then merge this once we are in a position to do this, probably next week.  Thanks.

timcnicholls:
@ajgdls well spotted! I'll fix that and update.

PR #100, opened 2018-03-28
Add low and high index attributes to dataset when in block write mode

MattTaylorDLS:
Albula needs the low and high index to be set on the dataset to be able
to read it properly
BC-682


MattTaylorDLS:
Ulrik and Alan agreed ok to merge in meeting

PR #102, opened 2018-04-09
Fix blocking status

ajgdls:


ajgdls:
Two out of three is good enough for me and @GDYendell is away for this week so I'm going to merge this now.

PR #106, opened 2018-04-13
Prevent frameReceiver controller from blocking when RX thread startup fails

### LINKED ISSUE #104: FrameReceiver locks up when attempting to bind already opened RX ports
If multiple frameReceiver instances are started with the same RX ports defined, the FR locks up without correctly initialising or reporting the error. This is most likely due to one of the reactors failing to start up correctly and the handshaking of the inter-thread channel not running.

Reported by @ajgdls

timcnicholls:
The RX thread startup error handling has been improved to prevent the main
thread controller from blocking when an initialisation error occurs (e.g.
failing to bind to RX ports). The RxThread start() method was calling join()
when the thread init fails, without allowing the thread to terminate. This
prevented reporting of the RX thread init error and the startup of the main
controller reactor, rendering the application unresponsive.

Fixes #104

### DIFF
--- a/frameReceiver/src/FrameReceiverRxThread.cpp
+++ b/frameReceiver/src/FrameReceiverRxThread.cpp
@@ -42,22 +42,31 @@ FrameReceiverRxThread::~FrameReceiverRxThread()
 //! This method starts the RX thread proper, blocking until the thread is started or has
 //! signalled an initialisation error, in which event an exception is thrown.
 //!
-void FrameReceiverRxThread::start()
+//! \return - bool indicating if thread initialisation and startup succeeded
+//!
+bool FrameReceiverRxThread::start()
 {
+
+  bool init_ok = true;
+
   rx_thread_ = boost::shared_ptr<boost::thread>(
     new boost::thread(boost::bind(&FrameReceiverRxThread::run_service, this)));
 
   // Wait for the thread service to initialise and be running properly, so that
   // this constructor only returns once the object is fully initialised (RAII).
-  // Monitor the thread error flag and throw an exception if initialisation fails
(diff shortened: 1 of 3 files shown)


frameReceiver/src/FrameReceiverRxThread.cpp
```diff
@@ -42,22 +42,31 @@ FrameReceiverRxThread::~FrameReceiverRxThread()
...
   // this constructor only returns once the object is fully initialised (RAII).
-  // Monitor the thread error flag and throw an exception if initialisation fails
+  // Monitor the thread error flag, log an error and exit false if failed.
 
   while (!thread_running_)
   {
     if (thread_init_error_) {
-      rx_thread_->join();
-      throw OdinData::OdinDataException(thread_init_msg_);
+      run_thread_ = false;
```
ulrikpedersen: Shouldn't we still throw an exception? It seems like an exceptional situation to me... (apologies, I may have arrived in the middle of this work without knowing the previous discussions...)

timcnicholls: No, I took this out deliberately because it is then untrapped apart from the outer-level catch-all, which them terminates the application. If this happens an error is printed and the status of the FR is reported to clients as unconfigured. This seemed like a more robust response to a misconfiguration than killing the process. If it threw an exception here it would just have to be caught in the calling function, which would then implement the same error/status behaviour anyway.

timcnicholls: BTW, the exception was never thrown before, as the join() blocked the main event loop completely - see #106.

PR #110, opened 2018-06-07
Development updates that should be in upstream master

ajgdls:
These were developed whilst testing Excalibur but are more generally useful:
Setting and reporting errors from the FP applications.
Configuration updates.
Updated dependencies.


frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -221,108 +222,119 @@ void FileWriterPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMess
...
+          }
+          LOG4CXX_INFO(logger_,
+                       "FrameProcessor will not receive any frames from this acquisition and so no output file will be created");
+        } else {
+          this->start_writing();
         }
-        LOG4CXX_INFO(logger_, "FrameProcessor will not receive any frames from this acquisition and so no output file will be created");
       } else {
-        this->start_writing();
+        this->stop_writing();
```
GDYendell: So this was a pretty huge bug? How did we not notice this?

frameProcessor/src/FrameProcessorPlugin.cpp
```diff
@@ -51,7 +51,46 @@ std::string FrameProcessorPlugin::get_name()
   return name_;
 }
 
-/** Configure the plugin.
+/** Set the error state.
+ *
+ * Sets the error level for this plugin
+ *
+ * \param[in] msg - std::string error message.
+ * \param[in] level - int error level of the message.
```
GDYendell: No level anymore

frameProcessor/src/FrameProcessorPlugin.cpp
```diff
@@ -51,7 +51,46 @@ std::string FrameProcessorPlugin::get_name()
...
+    }
+  }
+  if (!found_error){
+    error_messages_.push_back(msg);
+  }
+}
+
+/** Clear any error state.
+ *
+ * Clears the error level and any message
```
GDYendell: Level

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: Is there a bug here? Or just the diff not coping very well?

ajgdls: @MattTaylorDLS can you quickly check that this has not been changed from your version as I'm not sure why I would have changed this unless I really thought I had found a bug.  Looking at the version I have I believe it is correct, if the CONFIG_WRITE param is false then we stop writing...

ajgdls:
Check updated comments @GDYendell

frameProcessor/src/FileWriterPlugin.cpp
MattTaylorDLS: yeah i had a look yesterday, i think it's fine. it's just you added a try catch around it all and the diff didn't really like that

PR #111, opened 2018-06-11
Bug fix for top level parameter arrays.

ajgdls:
Problem with top level param arrays.  The name was tested to see if parameter existed but the [] was not removed before the test and so the test would always fail.  This would result in single item arrays only for top level parameters.


ajgdls:
Sorry minor bug squash found by Matthew.  This needs merging before we make a release.  Top level parameter arrays would only ever be size 1 and display the last item before this fix.  Top level arrays have only recently been used for error messages which is why we haven't seen this before.

PR #118, opened 2018-09-10
Move frame offsets to separate plugin

MattTaylorDLS:
Also includes other changes to add number of frames processed to status message, logging changes and change the acquisition object to configure and generate filenames in a clearer way

### DIFF
--- a/frameProcessor/include/Acquisition.h
+++ b/frameProcessor/include/Acquisition.h
@@ -64,6 +64,8 @@ public:
   std::string file_path_;
   /** Name of the file to write to */
   std::string filename_;
+  /** Configured value for the filename to use */
+  std::string configured_filename_;
   /** Use the earliest version of hdf5 */
   bool use_earliest_hdf5_;
   /** HDF5 file chunk alignment threshold */
--- a/frameProcessor/include/Acquisition.h
+++ b/frameProcessor/include/Acquisition.h
@@ -38,7 +38,6 @@ public:
   bool start_acquisition(
       size_t concurrent_rank,
       size_t concurrent_processes,
-      size_t frame_offset_adjustment,
       size_t frames_per_block,
       size_t blocks_per_file,
       bool use_earliest_hdf5,
@@ -48,7 +47,7 @@ public:
   bool check_frame_valid(boost::shared_ptr<Frame> frame);
   size_t get_frame_offset_in_file(size_t frame_offset) const;
   size_t get_file_index(size_t frame_offset) const;
-  size_t adjust_frame_offset(size_t frame_no) const;
+  size_t adjust_frame_offset(boost::shared_ptr<Frame> frame) const;
   boost::shared_ptr<HDF5File> get_file(size_t frame_offset);
   std::string get_create_meta_header();
   std::string get_meta_header();
@@ -84,8 +83,6 @@ public:
   size_t concurrent_processes_;
   /** Rank of this file writer */
   size_t concurrent_rank_;
-  /** Offset between raw frame ID and position in dataset */
-  size_t frame_offset_adjustment_;
   /** Number of frames to write consecutively in a file */
   size_t frames_per_block_;
   /** Number of blocks to write in a file  */
--- a/frameProcessor/include/Acquisition.h
+++ b/frameProcessor/include/Acquisition.h
@@ -40,6 +40,7 @@ public:
       size_t concurrent_processes,
       size_t frames_per_block,
       size_t blocks_per_file,
+      std::string file_extension,
       bool use_earliest_hdf5,
       size_t alignment_threshold,
       size_t alignment_value);
@@ -66,6 +67,8 @@ public:
   std::string filename_;
   /** Configured value for the filename to use */
   std::string configured_filename_;
+  /** File extension to use */
+  std::string file_extension_;
   /** Use the earliest version of hdf5 */
   bool use_earliest_hdf5_;
   /** HDF5 file chunk alignment threshold */
--- a/frameProcessor/include/Acquisition.h
+++ b/frameProcessor/include/Acquisition.h
@@ -65,7 +65,7 @@ public:
   std::string file_path_;
   /** Name of the file to write to */
   std::string filename_;
-  /** Configured value for the filename to use */
+  /** Configured value to be used as the prefix to generate the filename. */
   std::string configured_filename_;
   /** File extension to use */
   std::string file_extension_;
--- a/frameProcessor/include/Frame.h
+++ b/frameProcessor/include/Frame.h
@@ -118,6 +118,24 @@ public:
    */
   unsigned long long get_frame_number() const { return this->frameNumber_; }
 
+  /** Set the frame offset for this Frame.
+   *
+   * This method sets the frame offset that is used to calculate
+   * the position in the data file this frame belongs.
+   *
+   * \param[in] number - the frame offset.
+   */
+  void set_frame_offset(int64_t offset) { this->frameOffset_ = offset; }
+
+  /** Get the frame offset for this Frame.
+   *
+   * This method returns the frame offset that is used to calculate
+   * the position in the data file this frame belongs.
+   *
+   * \return the frame offset.
+   */
+  int64_t get_frame_offset() const { return this->frameOffset_; }
+
   /** Get the acquisition id for this Frame
    *
    * This method gets the id of the parent acquisition of this frame
@@ -173,6 +191,8 @@ private:
   size_t bytes_per_pixel;
   /** Frame number */
   unsigned long long frameNumber_;
+  /** Frame offset */
+  int64_t frameOffset_;
   /** Vector of dimensions */
   dimensions_t dimensions_;
   /** Compression type of raw data */
--- a/frameProcessor/src/Acquisition.cpp
+++ b/frameProcessor/src/Acquisition.cpp
@@ -32,7 +32,6 @@ const std::string META_STOP_ITEM         = "stopacquisition";
 Acquisition::Acquisition() :
         concurrent_rank_(0),
         concurrent_processes_(1),
-        frame_offset_adjustment_(0),
         frames_per_block_(1),
         blocks_per_file_(0),
         frames_written_(0),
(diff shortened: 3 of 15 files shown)


frameProcessor/include/Acquisition.h
```diff
@@ -65,6 +64,8 @@ class Acquisition : public MetaMessagePublisher
   std::string file_path_;
   /** Name of the file to write to */
   std::string filename_;
+  /** Configured value for the filename to use */
```
GDYendell: This description is a bit cryptic...

frameProcessor/src/Acquisition.cpp
```diff
@@ -478,33 +474,34 @@ boost::shared_ptr<HDF5File> Acquisition::get_file(size_t frame_offset) {
...
   size_t frame_offset = 0;
-  if (frame_no < this->frame_offset_adjustment_) {
-    // Deal with a frame arriving after the very first frame
-    // which was used to set the offset: throw a range error
-    throw std::range_error("Frame out of order at start causing negative file offset");
+
+  LOG4CXX_DEBUG(logger_, "Raw frame number: " << frame_no << ", Frame offset adjustment: " << frame_offset_adjustment);
+
+  if (frame_offset_adjustment < 0) {
+    if (-frame_offset_adjustment > frame_no) {
```
GDYendell: Make this
`if (frame_no + frame_offset_adjustment < 0 ) { ...`
and it will be clear enough that you don't need the comment explaining what is happening.

frameProcessor/src/Acquisition.cpp
```diff
@@ -580,7 +577,14 @@ std::string Acquisition::generate_filename(size_t file_number) {
   } else {
     char number_string[7];
     snprintf(number_string, 7, "%06d", file_number + 1);
```
GDYendell: While we are here. Is it a bad idea to use padding that would allow creating a million raw files in an acquisition (i.e. in the same directory)?

frameProcessor/src/Acquisition.cpp
```diff
@@ -580,7 +577,14 @@ std::string Acquisition::generate_filename(size_t file_number) {
   } else {
```
GDYendell: Do we want to get rid of *_rX.h5 file names while we are here? We could make this block just define the number string and then allow the block below create the full filename from filename or acquisition id in the same way.

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -628,6 +611,7 @@ void FileWriterPlugin::status(OdinData::IpcMessage& status)
   status.set_param(get_name() + "/writing", this->writing_);
   status.set_param(get_name() + "/frames_max", (int)this->current_acquisition_->frames_to_write_);
   status.set_param(get_name() + "/frames_written", (int)this->current_acquisition_->frames_written_);
+  status.set_param(get_name() + "/frames_processed", (int)this->current_acquisition_->frames_processed_);
```
GDYendell: :+1:

frameProcessor/src/OffsetAdjustmentPlugin.cpp
```diff
@@ -0,0 +1,97 @@
...
+    if (config.has_param(FIRST_FRAME_OFFSET_CONFIG)) {
+      first_frame_number_ = (uint64_t) config.get_param<uint64_t>(FIRST_FRAME_OFFSET_CONFIG);
+      LOG4CXX_INFO(logger_, "Setting first frame number to " << first_frame_number_);
+    }
+  }
+  catch (std::runtime_error& e)
+  {
+    std::stringstream ss;
+    ss << "Bad ctrl msg: " << e.what();
+    this->set_error(ss.str());
```
GDYendell: Maybe log this to console, too.

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -512,8 +493,10 @@ void FileWriterPlugin::configure_file(OdinData::IpcMessage& config, OdinData::Ip
     LOG4CXX_DEBUG(logger_, "Next file path changed to " << this->next_acquisition_->file_path_);
   }
   if (config.has_param(FileWriterPlugin::CONFIG_FILE_NAME)) {
+    // Set both the configured filename and the filename_ value. Configured will only be used  in 'block' writing mode
+    this->next_acquisition_->configured_filename_ = config.get_param<std::string>(FileWriterPlugin::CONFIG_FILE_NAME);
     this->next_acquisition_->filename_ = config.get_param<std::string>(FileWriterPlugin::CONFIG_FILE_NAME);
```
GDYendell: I think this needs reworking as part of getting rid of _rX filenames. Setting two parameters from the same configuration could be confusing.

frameProcessor/src/Acquisition.cpp
MattTaylorDLS: Can't add a size_t (unsigned) and a int64_t (signed) together and check if less than 0 as the signed int gets converted to unsigned before addition so it has to be done this way unfortunately.

frameProcessor/src/Acquisition.cpp
MattTaylorDLS: I'm not sure, but it's the padding that Eiger uses so think we have to have 6. Unless we want to add another config parameter to specify how many digits we want in the string?

frameProcessor/test/FrameProcessorTest.cpp
```diff
@@ -700,6 +730,96 @@ BOOST_AUTO_TEST_CASE( AdjustUID )
...
+  // Check the new offset is applied when on frame 2
+  offset = 2;
+  frame->set_frame_number(offset);
+  frame->set_frame_offset(offset);
+  plugin.process_frame(frame);
+  BOOST_CHECK_EQUAL(35, frame->get_frame_offset());
+}
+
+BOOST_AUTO_TEST_SUITE_END(); //UIDAdjustmentPluginUnitTest
+
```
ajgdls: :+1:

frameProcessor/include/Frame.h
```diff
@@ -173,6 +191,8 @@ class Frame
   size_t bytes_per_pixel;
   /** Frame number */
   unsigned long long frameNumber_;
+  /** Frame offset */
+  int64_t frameOffset_;
```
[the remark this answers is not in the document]
MattTaylorDLS: Yes

frameProcessor/src/Acquisition.cpp
MattTaylorDLS: Nope it's used whenever you are in 'block mode'
If a filename is specified, it uses that as the prefix, otherwise it will use the acquisition id as the prefix

frameProcessor/src/Acquisition.cpp
MattTaylorDLS: Could do, Up for discussion with all the interested detector 'teams' i guess

frameProcessor/src/Acquisition.cpp
ajgdls: If they ever want to do a 2.1 billion frame scan then I'm going to run away and hide.

frameProcessor/include/Frame.h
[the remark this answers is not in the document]
ajgdls: I'm not sure I can think of a case where a negative value would be used (i.e. jumping forward to leave a gap in the acquisition) but maybe it could be a valid situation.  OK leave it as int.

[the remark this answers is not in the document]
MattTaylorDLS: I could be used if you frame counter doesn't reset at the start of every acquisition. You then need to supply a negative number to overwrite existing data.
Don't we want to supply a negative number for Percival, or whichever detector it is that starts at 1 instead of 0? Or will that be done in the FrameDecoder plugin now?

frameProcessor/src/Acquisition.cpp
GDYendell: I give it 5 years.

frameProcessor/include/Frame.h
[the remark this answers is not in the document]
ajgdls: Oh I see I had it the wrong way around in my head, I thought we subtracted the value.  So in this case we will never need to use a positive number?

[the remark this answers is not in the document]
MattTaylorDLS: We will need to use a positive number for the 'normal' way it would work with Malcolm, where the detector does reset to 0 whenever you rewind, so you'd need to send a positive value to ensure the new frames are put into the right place (later on) in the file

MattTaylorDLS:
I've made the changes we talked about yesterday, after confirming with Ulrik he's happy to hardcode in .h5 file exstension

frameProcessor/src/OffsetAdjustmentPlugin.cpp
GDYendell: I don't think it does: https://github.com/odin-detector/odin-data/blob/8e87b11d99f83073584c5b6b46a4a9f96d851e59/frameProcessor/src/FrameProcessorPlugin.cpp#L60

GDYendell:
My only remaining thought is that on filename vs configure_filename. I don't think these describe the differences between the two very well. Also, does this still allow us to set the entire file name if we aren't in block mode? I thought we were getting rid of that and always generating it with 00001.h5 on the end?

MattTaylorDLS:
Yes the naming is a little unfortunate maybe. The configuration you pass to the FrameProcessor is called file/name as discussed, so it makes sense to have and set the configured_filename member from this (even though technically it's just a prefix). The filename member variable is then set when the full filename is generated, such that if you get the 'filename' variable from the Acquisition object, you will get the actual filename that it's currently writing to. So they both make sense on their own, 

And no, this doesn't allow you to set the entire filename, whatever mode you're in as it's always going to generate the filename from the configured filename (or acquisition ID if the configured_filename is empty)

frameProcessor/src/OffsetAdjustmentPlugin.cpp
MattTaylorDLS: Yeah it doesn't look like it. I copied this try catch from the existing FileWriterPlugin::configure() so if we want to log it to console, probably should go in there as well

GDYendell:
Ah that makes sense. Names are hard.

timcnicholls:
Sorry for the late comment but is this going to hard code us into using .h5 for all HDF files? That might be fine in DLS but isn't the case elsewhere, where .hdf5 is also common. I'd be uncomfortable if that wasn't configurable...

MattTaylorDLS:
I've put back in the extension parameter into the FP adapter, can you just have a quick look to make sure it's all good please @ajgdls and then i'll Merge this PR.

ajgdls:
It's OK @MattTaylorDLS I've just seen your other comment where you have allowed for the full stop (or not).
Great I'm happy.

PR #121, opened 2018-10-01
Live view plugin

ANeaves:
Initial Version of the Live View Plugin


frameProcessor/src/LiveViewPlugin.cpp
```diff
@@ -0,0 +1,366 @@
...
+  rapidjson::StringBuffer buffer;
+  buffer.Clear();
+  rapidjson::Writer<rapidjson::StringBuffer, rapidjson::UTF8<> > writer(buffer);
+
+  document.Accept(writer);
+
+  LOG4CXX_TRACE(logger_, "LiveViewPlugin Header Built, sending down socket.");
+  publish_socket.send(buffer.GetString(), ZMQ_SNDMORE);
+  LOG4CXX_TRACE(logger_, "LiveViewPlugin Sending frame raw data");
+  publish_socket.send(size, frame_data_copy, 0);
```
MattTaylorDLS: Do we know what ZMQ does here? Does it make its own copy, or does it just use the raw pointer and size, and read from it when it actually gets round to sending? I.e. is there a possibility that the frame has been pushed on to the next plugin, which then finishes with it, and says it can be released, all before ZMQ has had a chance to read from it and send it?

frameProcessor/src/LiveViewPlugin.cpp
```diff
@@ -0,0 +1,366 @@
...
+{
+  std::string delim = ",";
+  datasets.clear();
+  if(!value.empty())
+  {
+    //delim value string by comma
+    boost::split(datasets, value, boost::is_any_of(delim));
+  }
+  std::string dataset_string = "";
+  for(int i = 0; i< datasets.size(); i++)
```
MattTaylorDLS: should be a space between "i" and "<", and the "for" and "("

frameProcessor/src/LiveViewPlugin.cpp
```diff
@@ -0,0 +1,366 @@
...
+const std::string LiveViewPlugin::DEFAULT_DATASET_NAME = "";
+
+/*Config Names*/
+const std::string LiveViewPlugin::CONFIG_FRAME_FREQ =  "frame_frequency";
+const std::string LiveViewPlugin::CONFIG_PER_SECOND =  "per_second";
+const std::string LiveViewPlugin::CONFIG_SOCKET_ADDR = "live_view_socket_addr";
+const std::string LiveViewPlugin::CONFIG_DATASET_NAME = "dataset_name";
+
+/*Enum style arrays for the header*/
+const std::string LiveViewPlugin::DATA_TYPES[] = {"uint8","uint16","uint32"};
```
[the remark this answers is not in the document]
ANeaves: Is it in that order? As in, 3 would mean 64bit and 4 would mean float?

frameProcessor/src/LiveViewPlugin.cpp
ANeaves: ```frame_data_copy``` here is a void* pointer to data copied from ```frame->get_data()```. Looking into the code for IPCChannel, it looks like the send method copies the data into a ZMQ Message before telling the socket to publish, so I believe the data should be copied before we call ```this->push(frame)```. I will do a little more digging to make sure though.

ANeaves: Sorry, slight mistake there, ```frame_data_copy``` is a copy of the pointer to the shared data, no data is copied at that point. However, IPC Channel still copies the data from that pointer into a ZMQ message before sending, using ```memcpy```, which I believe is a blocking method, so we shouldn't be able to get to sending the frame on until at least the data has been copied, even if something slows the actual sending of the message.

ajgdls:
Hi @ANeaves I'm going to build this plugin into an Excalibur example which I can then test with some real hardware.  This will likely happen next week.  Would there be any access to a viewing tool by then?  It doesn't need to be completed/fully operational I'll just use whatever there is to try to live view the Excalibur test pattern.

frameProcessor/src/LiveViewPlugin.cpp
GDYendell: Do you need to make the frame_data_copy at all? Can we just pass frame->get_data() to avoid this confusion?

Presumably we wouldn't connect a plugin to the LiveViewer anyway, the LiveViewer and FileWriter would both connect to the (e.g.) ExaliburFrameProcessor, in parallel.

frameProcessor/src/LiveViewPlugin.cpp
```diff
@@ -0,0 +1,363 @@
...
+/**
+ * Gets the type of the data, based on the enum value
+ * Values as follows:
+ * 0 - raw_8bit
+ * 1 - raw_16bit
+ * 2 - raw_32bit
+ * \param[in] type - enum value
+ * \return string value representing data type, or "unknown" if an unrecognised enum value
+ */
+std::string LiveViewPlugin::getTypeFromEnum(int32_t type)
```
GDYendell: Could this be moved to FrameProcessorDefinitions so it can be used everywhere?

timcnicholls:
@ajgdls If you're willing to use python3 / PyQT5 we have a demo app that works with the live view plugin. We won't be releasing this into the wild though it was just a proof of principle.

frameProcessor/src/LiveViewPlugin.cpp
GDYendell: *This* meaning the arrays of strings and the conversion methods for data type and compression.

ajgdls:
Hi @timcnicholls sure I think if I create the relevant virtualenv I can install everything at DLS and use those tools.  I'll make do with whatever you have got to test the live image plugin initially, and then I'll look at the areaDetector implementation.  Thanks.

timcnicholls:
@ajgdls OK we can try this later in the week when you're at DLS. Or I can make a venv and install in /dls_sw if you have a preferred location? 

BTW, we tested it with exactly this setup - EXCALIBUR plugin chain with PCAP playback.

frameProcessor/src/LiveViewPlugin.cpp
MattTaylorDLS: Ah yes good point @GDYendell , in which case i guess there would be an even greater chance that memory would be free'd before it was used if there was no plugin behind it in the chain. I don't know how the parallel chain would work with the ready and release - does it know how many chains it's being used on and so waits for 2 releases?

But it sounds like the IPC channel does a copy as @ANeaves says so it's ok with respect to my initial comment.

doc/Live_View_Spec.md
```diff
@@ -0,0 +1,73 @@
...
+
+The data published on the ZMQ socket is a multipart message generated from the frame of data. It comes in two parts, a header and a data blob.
+
+### Header
+The header contains the following data as a json string
+- frame_num - *int*
+  - The number assigned to the frame. This is the number used to calculate if the frame should be shown when *per_second* is **false**.
+- acquisition_id - *string*
+  - The Acquisition ID number assigned by the detector. This can be blank.
+- dtype - *string*
```
ajgdls: Could you add the possible values here?

doc/Live_View_Spec.md
```diff
@@ -0,0 +1,73 @@
...
+  - The data type of the following data blob. Represents the size of each pixel of data.
+- dsize - *int*
+  - The total size of the data blob in bytes
+- compression *string*
```
ajgdls: Could you add the possible values here?

frameProcessor/src/LiveViewPlugin.cpp
```diff
@@ -0,0 +1,366 @@
...
+    boost::posix_time::ptime now = boost::posix_time::microsec_clock::local_time();
+    int32_t frame_num = frame->get_frame_number();
+    int32_t elapsed_time = (now - time_last_frame).total_milliseconds();
+
+    if(per_second != 0 && elapsed_time > time_between_frames) //time between frames too large, showing frame no matter what the frame number is
+    {
+      LOG4CXX_TRACE(logger_, "Elapsed time " << elapsed_time << " > " << time_between_frames);
+      PassLiveFrame(frame);
+    }
+    else if(frame_freq != 0 && frame_num % frame_freq == 0)
```
ajgdls: I think this might be an issue when we have multiple FP applications with a rank and process number.  If I'm right then we could get frame numbers that are not continuous, so for the case where there are 4 FPs then the first gets frames 0,4,8...
I think in this case you will need to take into account the rank and number of processes.  @MattTaylorDLS I think will be able to confirm the non continuous frame numbers.  In other plugins we have two configuration options to allow the rank and number of processes to be set, and then these values can be used to calculate a continuous frame number from the actual incoming frame numbers.  I hope this makes sense.

frameProcessor/src/LiveViewPlugin.cpp
ajgdls: @MattTaylorDLS it works with shared pointers to the memory.  So the pointers maintain an internal reference count to how many pointers there actually are pointing to a block.  I only release the memory when the frame object is destroyed, and the frame object can only be destroyed when there are no more shared pointers pointing to it.  So we can have parallel chains without any trouble as long as they do not actually alter the memory (which is the original reason for having the memory as a const ptr).

frameProcessor/src/LiveViewPlugin.cpp
ANeaves: Done, its been moved into ```FrameProcessorDefinitions.h```

frameProcessor/src/LiveViewPlugin.cpp
MattTaylorDLS: This is true, but I guess it depends on what it means to say "I want to live view every X frames"
If you want to see every X frames for the whole acquisition, then it's ok as it is, and you'll need to put this live view plugin on to every FP node. If you want to see every X frames that passes through this FP node, then you will need to take into account the rank, process, frames_per_block, blocks_per_file etc.

ajgdls: Yes agreed, perhaps then it is fine as it is.

GDYendell: We initially said it would be easiest to only put the plugin on one of the processes, but with the script generation I think it would be just as easy (easier, actually) to load it on all processes. Plus it will make the rate much more configurable anyway.

ANeaves: One potential concern with what you've said is that you can only bind a single ZMQ socket to an address. If we attempted to have a LiveViewPlugin on every node, and they bound to the same address, it could cause issues. ```IpcChannel``` does not appear to have any sanity checking when told to bind to an address, so could cause an exception if said address has already been bound

GDYendell: We might be able to use an XSUB/XPUB to connect multiple publishers to multiple subscribers. Could this live in the OdinServer adapter?

timcnicholls: They would have to be configured to have distinct endpoints. `IpcChannel` does the correct thing in throwing an exception if `bind` fails. Suggest we avoid the rank/process complexity for now until we understand the usage better. The LV plugin could also just count frames it is pushed and send one in X if necessary.

MattTaylorDLS: Yeah i think the internal counter would be the simplest option for now. (It'll look quite odd with flurries of activity and then nothing for a while when in 'block' writing mode, but there's no way around that without each node having the plugin running.

ANeaves:
Hey guys, I've tried to implement all the changes mentioned in your comments (including changing the plugin to count the frames manually rather than relying on the frame number), so if you could have another look through and check I've understood your comments correctly, and not missed anything, I'd appreciate it.

ajgdls:
Hi @timcnicholls if you guys have already tested the live view with Excalibur then I don't need to.  Thanks.

timcnicholls:
@ajgdls Well only with PCAP files. Would love to see it run on the real thing! Happy to install it somewhere useful (in /dls_sw/.... ? Can never remember where things go).

frameProcessor/test/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
-
 set(CMAKE_INCLUDE_CURRENT_DIR on)
 ADD_DEFINITIONS(-DBOOST_TEST_DYN_LINK)
+ADD_DEFINITIONS(-DBUILD_DIR="${CMAKE_BINARY_DIR}")
```
GDYendell: One final thing: What does this do?

ajgdls:
@timcnicholls  oh OK sure.  I will email you the installation location.

frameProcessor/test/CMakeLists.txt
ANeaves: Allows the tests to compile from multiple sources rather than the one 1000(ish) line long ```FrameProcessorTest.ccp``` file. I've split the tests for the Live View out into a separate file ```LiveViewUnitTest.cpp```, mainly for maintainability/readability reasons. I stole that particular line from the Frame Receiver, where all the tests are in separate source files and compiled into a single binary file.

doc/Live_View_Spec.md
ANeaves: Done, document now lists the possible string values for dtype and compression type

PR #122, opened 2018-10-01
Make debugging consistent between FR and FP

ajgdls:
I have:
1) Added debug levels into FP application
2) Added --debug switch into FP application command line parameters (NOTE: To make it consistent with the FR I have changed the -d switch.  This used to specify the detector, now it sets the debug level and the only way to specify the detector is with the --detector flag).
3) Added a control item (debug) to allow setting the debug level from the control interface whilst the applications are running.

I have updated the core debug level messages for the FrameProcessor but I have not updated the plugin specific messages (FileWriter plugin).  This should be carried out but some consideration is needed on what level for each type of message.  For that I will probably ask @MattTaylorDLS to provide some input for the various file writer classes.


ajgdls:
Hi @MattTaylorDLS I'm in Italy this Thursday for Percival F2F so next week?

frameProcessor/src/FrameProcessorController.cpp
```diff
@@ -8,13 +8,18 @@
...
 
 namespace FrameProcessor
 {
+IMPLEMENT_DEBUG_LEVEL;
+
 const std::string FrameProcessorController::META_RX_INTERFACE        = "inproc://meta_rx";
 
 const std::string FrameProcessorController::CONFIG_SHUTDOWN          = "shutdown";
 
+const std::string FrameProcessorController::CONFIG_DEBUG             = "debug";
```
timcnicholls: I wonder if `debug_level` would be more explicit? We should also expose this for the FR control interface too. I'm happy to do it separately unless you want to add it on this PR @ajgdls ?

ajgdls: Hi @timcnicholls it should already work with the FR interface, I added the relevant calls and it appeared to work when I tested it.
See commit 
"Updated FrameReceiver application to allow setting the debug level from the control interface"

I will change the text to be debug_level instead of debug.

frameProcessor/src/FrameProcessorApp.cpp
```diff
@@ -61,6 +64,8 @@ void parse_arguments(int argc, char** argv, po::variables_map& vm, LoggerPtr& lo
     // and in the configuration file
     po::options_description config("Configuration options");
     config.add_options()
+        ("debug,d",      po::value<unsigned int>()->default_value(debug_level),
```
GDYendell: Could the command line option be **debug_level**, too?

GDYendell: **debug-level** ***

MattTaylorDLS: I'm not the biggest fan of dashes in command line options, but that would be fine.

timcnicholls: The GNU convention for long options is [dashes]( https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html). (Python supports this in argparse by transmuting dashes in option names into underscores in the generated attributes).

ajgdls:
Because I have a need for all green ticks, I've made the command line debug changes. ;)

PR #123, opened 2018-10-03
Add status reporting to SharedMemoryController

timcnicholls:
This PR is to address an issue identified on detector projects where the FR/FP are configured immediately prior to the start of the run. The FP, when configuring its SharedMemoryController instance, requests shared buffer configuration info from the FR asynchronously once the ZMQ channels are connected. There is currently no way to interrogate the FP to determine if has completed. If data starts arriving prior to this completion, frames are lost from the data path. The FP status reporting is therefore modified to include a shared_memory/configured flag which a controlling client can poll prior to starting data acquisition.

### DIFF
--- a/frameProcessor/src/SharedMemoryController.cpp
+++ b/frameProcessor/src/SharedMemoryController.cpp
@@ -11,6 +11,8 @@
 namespace FrameProcessor
 {
 
+const std::string SharedMemoryController::SHARED_MEMORY_CONTROLLER_NAME = "shared_memory";
+
 /** Constructor.
  *
  * The constructor sets up logging used within the class. It also creates the
@@ -29,7 +31,8 @@ SharedMemoryController::SharedMemoryController(boost::shared_ptr<OdinData::IpcRe
                                                const std::string& txEndPoint) :
     reactor_(reactor),
     rxChannel_(ZMQ_SUB),
-    txChannel_(ZMQ_PUB)
+    txChannel_(ZMQ_PUB),
+    sharedBufferConfigured_(false)
 {
   // Setup logging for the class
   logger_ = Logger::getLogger("FW.SharedMemoryController");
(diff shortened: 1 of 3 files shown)


frameProcessor/src/SharedMemoryController.cpp
```diff
@@ -98,6 +101,8 @@ void SharedMemoryController::setSharedBufferManager(const std::string& shared_bu
       new OdinData::SharedBufferManager(shared_buffer_name)
   );
 
+  sharedBufferConfigured_ = true;
```
GDYendell: Should this be set to false at the beginning of setSharedBufferManager, so that we know it isn't OK to start an acquisition while it is reconfiguring?

timcnicholls: No, this mechanism is only invoked if a new SharedMemoryController is constructed due to the appropriate incoming configuration message. It is set to false in the member initialisation list of the class, so is false until the SharedBufferManager is constructed once a notification arrives from the FR. From the perspective of a controlling client, the setSharedBufferManager operation is atomic w.r.t handling client channel requests so there is no way for the transient state you describe to be interrogated by a client.

GDYendell: Ah, so the only way tt gets a NotifyBufferConfig is if it sends a BufferConfigRequest (i.e. in the constructor)?

timcnicholls: Actually that's a good point. From the perspective of the FP, what I said is true. However, if the FR is restarted/reconfigured it sends a NotifyBufferConfig message unsolicited, so setSharedBufferManager can be called independently of the SharedMemoryController constructor. I'll set it to false at the start of the method as you suggest, thanks.

timcnicholls:
Update: 8/10/18. Rebased from master fixing conflict following merge of PR #122 . 

@ajgdls Can you check you are happy with this PR please?

ajgdls:
Yes looks fine sorry for the delay.

timcnicholls:
@ajgdls Thanks - wanted to be sure it wasn't clobbering your PR. Rebase/merge/PR with conflicts always makes me nervous 😬

PR #126, opened 2018-10-11
Improve FrameProcessor error handling

GDYendell:
Handle errors and expose them to the client more consistently.
Stop the swallowing of exceptions at the top of the FrameProcessorApp.
Implement a process_updates method in the odin_data_adapter update_loop
  to allow child classes to perform some extra logic after a parameter tree update.

### DIFF
--- a/frameProcessor/src/FrameProcessorApp.cpp
+++ b/frameProcessor/src/FrameProcessorApp.cpp
@@ -557,7 +557,7 @@ int main(int argc, char** argv)
 
   } catch (const std::exception& e) {
     LOG4CXX_ERROR(logger, e.what());
-    // Nothing to do, terminate gracefully(?)
+    throw;
   }
   return 0;
 }
--- a/frameProcessor/src/FrameProcessorApp.cpp
+++ b/frameProcessor/src/FrameProcessorApp.cpp
@@ -556,7 +556,7 @@ int main(int argc, char** argv)
     LOG4CXX_DEBUG_LEVEL(1, logger, "FrameProcessorController run finished. Stopping app.");
 
   } catch (const std::exception& e) {
-    LOG4CXX_ERROR(logger, e.what());
+    LOG4CXX_ERROR(logger, "Caught unhandled exception in FrameProcessor, application will terminate: " << e.what());
     throw;
   }
   return 0;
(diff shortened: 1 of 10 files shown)


ajgdls:
Can we merge this now?  I would like to update the debug messages for the odin-data plugins (specifically FileWriter) but I think it would be good to make those changes with this PR in place.

frameProcessor/src/FrameProcessorApp.cpp
```diff
@@ -557,7 +557,7 @@ int main(int argc, char** argv)
 
   } catch (const std::exception& e) {
     LOG4CXX_ERROR(logger, e.what());
```
ulrikpedersen: I would probably add a little notice in the log message to say that this is an unhandled exception and the application will crash from here... Even nicer (bonus points) would be to add a stacktrace to the log message - but don't let that hold up the merge...

Question about re-throwing: will a debugger still produce a stack-trace so that we can track where the problem originated from? I think (hope) the answer is yes but I've never fully understood exceptions and the effect on the stack.

dw96: In my limited experience (say running FP from a shell, bad memory handling), an unhandled exception produce a segfault without a stacktrace. Only seen one trace so far. There's always the choice of examining the resulting core file in gdb but that can be of limited use.
Running for instance the FP within gdb would produce a trace, but that's beyond this scope I suspect.

ulrikpedersen: Yes, of course the stacktrace isn't printed by default (that's the nice-to-have addition that we could think about later, using something like [GNU Backtraces](http://www.gnu.org/software/libc/manual/html_node/Backtraces.html) or even something higher-level). 

I'm referring to the GDB/coredump stack trace. I'm concerned about whether or not the stack trace is tangled up in some way when the program ends up in an exception handler and is then re-thrown (and not captured) as opposed to a crash in a more 'normal' part of code...?

ajgdls:
@ulrikpedersen as Gary isn't here I have pushed an updated error message in response to your comment.  Is the text suitable?

ulrikpedersen:
Yes, thanks @ajgdls  Ready to merge I think.

Issue #124, opened 2018-10-08
Travis Occasionally fails for Live View Tests


ANeaves:
Travis will occasionally fail the build during testing because of some tests in the Live View, due to delays in closing ZMQ sockets during test fixture teardown. This means that when rebuilding the fixture for the next test the ZMQ socket binding will raise an exception "Address already in use".

ANeaves:
Fixed by https://github.com/odin-detector/odin-data/pull/128. Socket ports increment with each test, so they won't reuse the same address and won't get this issue anymore.

PR #127, opened 2018-10-12
Fix implementation of debug level logging

timcnicholls:
The recent change to make debug level logging available more widely across odin-data has exposed a flaw in the old mechanism of using `IMPLEMENT_DEBUG_LEVEL` to define (vs declare) the global `debug_level` variable and the `set_debug_level` function. 

Both the FR and FP need to call `set_debug_level` in both the main app (when parsing command-line or config file options) and in their respective controller, when modifying the debug level in response to a configuration request. This led to multiple references to IMPLEMENT_DEBUG_LEVEL in different namespaces, causing there to be multiple debug levels per application 😬 .

This PR solves this problem **and** that raised by @hir12111's recent [EXCALIBUR PR](https://github.com/dls-controls/excalibur-detector/pull/24) by putting the concrete definition of the debug level into a source file in the core odin-data library, which all applications link to. This avoids any need for `IMPLEMENT_DEBUG_LEVEL` - code can make use debug levels simply by including `DebugLevelLogger.h`.

Downstream plugins will need to remove references to `IMPLEMENT_DEBUG_LEVEL` once this is merged. Has been tested on excalibur-detector.

### DIFF
--- a/common/include/DebugLevelLogger.h
+++ b/common/include/DebugLevelLogger.h
@@ -15,13 +15,10 @@ using namespace log4cxx::helpers;
 typedef unsigned int DebugLevel;
 
 extern DebugLevel debug_level;
+void set_debug_level(DebugLevel level);
 
 #define LOG4CXX_DEBUG_LEVEL(level, logger, message) { \
   if (LOG4CXX_UNLIKELY(level <= debug_level)) {\
     LOG4CXX_DEBUG(logger, message); }}
-
-#define IMPLEMENT_DEBUG_LEVEL \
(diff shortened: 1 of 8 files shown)


ajgdls:
Looks good but did the build fail?

GDYendell:
It looks like the problem in #124 , it just needs to be re-run.

timcnicholls:
Yes, the build is failing due #124. The test apps all pass on the command line.

timcnicholls:
Re-ran the travis job and it passed OK.

PR #129, opened 2018-10-18
Add support for UDP trailer mode to frame receiver

timcnicholls:
This PR adds support for UDP 'trailer mode' to the frame receiver, where the packet level metadata is appended to the payload. This is required for some STFC-supported detector systems. The primary change is to the logic of building the `io_vec`list for receiving UDP packets and a change to the `process_packet` argument list to allow that method to identify the source of incoming packets. **NB** this requires a modification to the process_packet method in **all** UDP frame decoders (e.g. PERCIVAL, EXCALIBUR).

This change has been validated with a modified EXCALIBUR decoder plugin.


timcnicholls:
Since @GDYendell is on leave and we want to push out a tagged release soon, I'm going to merge this without his approval.

Issue #125, opened 2018-10-08
Add statistics reset to FrameReceiver


timcnicholls:
It should be possible to reset packets_lost counters etc in the frameReceiver from a client, by sending a `reset_statistics` IpcMessage command. The equivalent configuration mechanism that exists in the frameProcessor will be refactored to match.

Issue raised in JIRA as [BC-769](https://jira.diamond.ac.uk/browse/BC-769).

PR #130, opened 2018-10-18
Reset frame receiver statistics

timcnicholls:
This PR addresses #125 and JIRA ticket [BC-769](https://jira.diamond.ac.uk/browse/BC-769). A `reset_statistics` command is implemented in `IpcMessage` and causes the FR and its loaded decoder plugin to reset any frame and packet level statistics. Derived decoder classes should implement a `reset_statistics` method to implement this.

@ajgdls will review this for implementation in the frame processor + migration of `clear_errors` from config parameter to command. Suggest that implementation is added to this branch and PR?

### DIFF
--- a/common/include/IpcMessage.h
+++ b/common/include/IpcMessage.h
@@ -92,6 +92,7 @@ public:
     MsgValCmdRequestConfiguration,   //!< Request configuration command message
     MsgValCmdBufferConfigRequest,    //!< Buffer configuration request
     MsgValCmdBufferPrechargeRequest, //!< Buffer precharge request
+    MsgValCmdResetStatistics,        //!< Reset statistics command
     MsgValCmdShutdown,               //!< Process shutdown request
     MsgValNotifyIdentity,            //!< Identity notification message
     MsgValNotifyFrameReady,          //!< Frame ready notification message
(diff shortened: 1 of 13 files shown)


ajgdls:
@timcnicholls I'm looking at this today.

ajgdls:
@timcnicholls I have implemented the command within the FP.  How do you see this being exposed through the REST interface?  Is it a specific PUT request that we need to look for within the FR and FP adapters, or do we need to implement another way (eg POST)? So far we have had:

status GET
config GET PUT

but we haven't really supported alternative commands before.

timcnicholls:
@ajgdls Thanks, looks good. I'd suggest that, at the REST interface level in odin-control adapters, commands are mapped onto PUT requests, from the perspective that they are requests that modify the state of resources (i.e. the state of the processes) rather than POST being the creation of a new resource (e.g. a new DB record).

timcnicholls:
Hi @ajgdls I can't review my own PR but I'm happy to merge this whenever you're ready.

ajgdls:
OK sure we have discussed that before.  So in the specific case of the reset I guess we will check for the reset_statistics URI but not look for any specific value sent with the request?

ajgdls:
Oh sorry @timcnicholls I think I should probably add the adapter code also as it is really part of this.  I'm looking at it now so should be ready for merging tomorrow?

timcnicholls:
Yeah that's fine, just didnt want to miss merging as we're both contributing to PR :wink:

PR #132, opened 2018-10-18
FP plugin log with debug levels

ajgdls:


timcnicholls:
@ajgdls I think it is safe to merge this. The build failure is unrelated. I've seen a couple of sporadic failures today on other PRs, but restarting the build seems to cure it. @ANeaves had fixed the original socket re-use issue in the LiveView test (and tested repeatedly) but this now seems to happen. We'll look into it ASAP.

ANeaves:
Looking into the failure now

PR #137, opened 2018-10-23
Version control status

ajgdls:
This PR implements a version interface object, then applies it to the FP application plugins and auto versions all of the odin-data plugins using cmake and git method calls to retrieve the information.  The result is the following status example for the hdf5 plugin:

"hdf":{"version":{"major":0,"minor":5,"patch":0,"short":"0.5.0","full":"0-5-0dls1-116-g9289602"}}

This change will force all plugin developers to implement the version methods (otherwise applications will not compile) but it doesn't place a restriction on how the version information is generated.
The version interface object can also be applied to the FR application objects, @timcnicholls we should look at how you would like that implemented in the FR controller and push additional commits before merging this PR.


common/include/IVersionedObject.h
```diff
@@ -0,0 +1,33 @@
+//
+// Created by gnx91527 on 22/10/18.
+//
+
+#ifndef ODINDATA_IVERSIONEDOBJECT_H
+#define ODINDATA_IVERSIONEDOBJECT_H
+
+#include <string>
+
+namespace FrameProcessor
```
timcnicholls: Should this be in namespace `OdinData` rather than `FrameProcessor`?

ajgdls: Mm that is a good point it should be OdinData as we are going to use it for both FR and FP.

ajgdls:
Hi @timcnicholls if you don't mind integrating into the FR that would be great.

timcnicholls:
@ajgdls Sure, I'll have a crack. I'll fix the namespace for `IVersionedObject` while I'm at it.

timcnicholls:
@ulrikpedersen Yes agree this is sensible. The core application should report its version at startup as well as the plugins.

@ajgdls I also notice in the FP status response the plugins report their version but not the core app. In principle they could be different. Should we add this also?

ajgdls:
Yes I can add that.  Just to be sure we are going to make the core app log its version to info at startup and then supply it as status in the same way as the plugins?

timcnicholls:
Yes, I think that would be useful.

timcnicholls:
E.g. I've just put the following into `ExcaliburProcessPlugin` constructor:
```
 LOG4CXX_INFO(logger_, "ExcaliburProcessPlugin version " << this->get_version_long()<< " loaded.");
```

timcnicholls:
Hi @ajgdls and @ulrikpedersen I've pushed more commits to this branch with version reporting in the FR and also printout at startup/load of app and plugin versions to the log file.

I have a suggestion to float: at present we are populating the status requests with all the version info, e.g.:
```
2018-10-24 14:13:49,729 INFO configure_odin_data.py - Sending status request to frame receiver
2018-10-24 14:13:49,730 INFO configure_odin_data.py - Got response: {
    "id": 1,
    "msg_type": "ack",
    "msg_val": "status",
    "params": {
        "buffers": {
            "empty": 799,
            "mapped": 0
        },
        "decoder": {
            "fem_packets_lost": [
                0,
                0,
                0,
                0,
                0,
                0
            ],
            "name": "ExcaliburFrameDecoder",
            "packets_lost": 0,
            "version": {
                "full": "0.1-149-gfe042e9",
                "major": 0,
                "minor": 1,
                "patch": 149,
                "short": "0.1.149"
            }
        },
        "frames": {
            "received": 0,
            "released": 0,
            "timedout": 0
        },
        "status": {
            "buffer_manager_configured": true,
            "configuration_complete": true,
            "decoder_configured": true,
            "ipc_configured": true,
            "rx_thread_configured": true
        },
        "version": {
            "full": "0.5.0-128-g46e8d6f",
            "major": 0,
            "minor": 5,
            "patch": 0,
            "short": "0.5.0"
        }
    },
    "timestamp": "2018-10-24T14:13:49.729811"
}
2018-10-24 14:13:49,731 INFO configure_odin_data.py - Sending status request to frame processor
2018-10-24 14:13:49,733 INFO configure_odin_data.py - Got response: {
    "id": 2,
    "msg_type": "ack",
    "msg_val": "status",
    "params": {
        "excalibur": {
            "bitdepth": "12-bit",
            "packets_lost": 0,
            "version": {
                "full": "0.1-149-gfe042e9",
                "major": 0,
                "minor": 1,
                "patch": 149,
                "short": "0.1.149"
            }
        },
        "hdf": {
            "acquisition_id": "0",
            "file_name": "test.hdf5_000001.h5",
            "file_path": "/tmp/",
            "frames_max": 0,
            "frames_processed": 0,
            "frames_written": 0,
            "processes": 1,
            "rank": 0,
            "timeout_active": false,
            "version": {
                "full": "0.5.0-128-g46e8d6f",
                "major": 0,
                "minor": 5,
                "patch": 0,
                "short": "0.5.0"
            },
            "writing": true
        },
        "plugins": {
            "names": [
                "excalibur",
                "hdf"
            ]
        },
        "shared_memory": {
            "configured": true
        },
        "version": {
            "full": "0.5.0-128-g46e8d6f",
            "major": 0,
            "minor": 5,
            "patch": 0,
            "short": "0.5.0"
        }
    },
    "timestamp": "2018-10-24T14:13:49.732478"
}
```
which means that polling the apps for status of acquisition returns quite a lot of static information. I wonder if the version reporting would be better placed in the reply to the `request_configuration` commands instead? It's reflective of the _configuration_ of the applications and their plugin chains, rather than the changing _status_ in my mind. Any opinion?

I'm happy to migrate this over in both apps if you agree...

GDYendell:
Could version information be a third, separate parameter tree?

timcnicholls:
@GDYendell Yes it could, e.g. in response to a `request_version` message from a client. That would work too.

ajgdls:
It is true that we call request_configuration as often as we request status from our higher level applications.  This is a simple model, whereas we could only request the configuration when we believe it has changed.  Both status and configuration can change.
However the version information is completely static so perhaps it should be a separate request altogether.

timcnicholls:
OK, sounds like we're erring on the side of a separate request. 

Arguably, the version info isn't _completely_ static, since a configuration request could result in a different set of plugins being loaded with different version information, but I'm happy with the separate request, especially if polling configuration is a frequent event.

I'll have a crack at implementing `request_version` and undo it if @ulrikpedersen objects too strenuously :wink: .

timcnicholls:
OK, `request_version` is implemented. Here's the output:
```
2018-10-24 15:47:17,011 INFO configure_odin_data.py - Sending version request to frame receiver
2018-10-24 15:47:17,012 INFO configure_odin_data.py - Got response: {
    "id": 1,
    "msg_type": "illegal",
    "msg_val": "request_version",
    "params": {
        "version": {
            "decoder": {
                "full": "0.1-149-gfe042e9",
                "major": 0,
                "minor": 1,
                "patch": 149,
                "short": "0.1.149"
            },
            "odin-data": {
                "full": "0.5.0-142-g075515e",
                "major": 0,
                "minor": 5,
                "patch": 0,
                "short": "0.5.0"
            }
        }
    },
    "timestamp": "2018-10-24T15:47:17.011640"
}
2018-10-24 15:47:17,012 INFO configure_odin_data.py - Sending version request to frame processor
2018-10-24 15:47:17,014 INFO configure_odin_data.py - Got response: {
    "id": 1,
    "msg_type": "ack",
    "msg_val": "request_version",
    "params": {
        "version": {
            "excalibur": {
                "full": "0.1-149-gfe042e9",
                "major": 0,
                "minor": 1,
                "patch": 149,
                "short": "0.1.149"
            },
            "hdf": {
                "full": "0.5.0-142-g075515e",
                "major": 0,
                "minor": 5,
                "patch": 0,
                "short": "0.5.0"
            },
            "odin-data": {
                "full": "0.5.0-142-g075515e",
                "major": 0,
                "minor": 5,
                "patch": 0,
                "short": "0.5.0"
            }
        }
    },
    "timestamp": "2018-10-24T15:47:17.013606"
}
```
I like this a lot more.

ajgdls:
Looks good, I think the FR should return message type "ack".

timcnicholls:
Oops, well spotted! Was too busy looking at the version fields...

ulrikpedersen:
> I'll have a crack at implementing request_version and undo it if @ulrikpedersen objects too strenuously wink .

For the record: I have no objections - it looks wonderful! Thanks :-)

PR #143, opened 2018-10-29
Version adapter

ajgdls:


ajgdls:
This exposes the version information as a top level item:

'''python ./latrd_http_client_2.py
DEBUG:root:Sending msg to: http://127.0.0.1:8888/api/0.1/fp/version/
RETURN: {u'value': [{u'odin-data': {u'major': 0, u'short': u'0.5.0', u'minor': 5, u'full': u'0-5-0dls1-147-g10c558a-dirty', u'patch': 0}}]}'''

tools/python/odin_data/odin_data_adapter.py
```diff
@@ -230,6 +230,9 @@ def put(self, path, request):  # pylint: disable=W0613
                         status_code = 503
                         response['error'] = 'Invalid command requested: {}'.format(command)
 
+            # Request a version update from the application after any configuration changes
```
MattTaylorDLS: Why do we request version update after any configuration changes?

ajgdls: We do not want to continuously poll for versions because they will rarely change.  There are two valid times to check the version.
1) When the applications are started (considered to have happened during a re-connection event).
2) When the libraries have been loaded, or some change has caused objects to get re-built.  This only happens when a configuration change takes place.  In theory we could specialise further to only check for version information when certain conditions have been met, or when certain configuration items have been altered, but this would turn 1 line of code into possibly 100 lines of code.

After discussion on Thursday with @timcnicholls we decided that configuration changes are relatively rare (compared to status updates) and so the overhead of re-checking the version information each time is minimal when compared with checking it on every status update.

MattTaylorDLS: I see. I think it still seems a little odd to be requesting the version every time you set the filname, number of frames etc.
I've not really seen much of these adapters before so not entirely sure how they work, but could you not just add 1 "if" line, to see if the request contains any of a predefined list of keywords which would possibly trigger a version change?

ajgdls: Yes in principle you could do that.  However, the list would need to live in the specific FR and FP adapters which would then result in the necessity to override the put method and perform the specific checks in each adapter.  I can do this, @timcnicholls do you know a suitable list of FR configuration parameters that could result in the re-build of the objects?

MattTaylorDLS: Ah right, yes not as easy as I'd hoped then. Well if we're happy with it sending the version request every time then you can keep it as is, but does seem a little odd to me

ajgdls: I'll make the change in the FP adapter to only check versions after a plugin command.  For the FR I've just looked and to be honest most configuration changes there result in a re-build of some part of the system.  I think for the FR it is reasonable to check for any config command.

timcnicholls: @ajgdls For the FR, it's the `decoder_path` and `decoder_type` config parameters that could result in a version change as a new decoder is loaded. The FR is capable of unloading an existing decoder and reloading a fresh one, so it could be the same name decoder from a different path (and hence a different version).

ajgdls: OK thanks @timcnicholls I'll implement those.

ajgdls:
And it was less than 100 lines ;)

MattTaylorDLS:
I knew it was possible!

PR #146, opened 2018-10-31
Made some updates to the readme.

ajgdls:
This is not a thorough revision of the README but I thought I could get this in before the release.  It contains some info regarding the JSON configuration files as well as updated command line options for the FR and FP.


timcnicholls:
Thank you for doing this @ajgdls !

Issue #150, opened 2018-11-07
Add concept of frames dropped to base Frame Decoder class


MattTaylorDLS:
Currently, the Frame Decoder class has a frames_timedout_ variable. It should however also have a frames_dropped_ variable which can be used to report how many complete frames have been dropped due to the shared memory buffer being full.
As per the frames_timedout_, it will be up to the implementing decoder class to set this value, but the value should belong in the base class, and be reported in the "status" message.

timcnicholls:
Happy to take this on

MattTaylorDLS:
cool thanks, not a lot to it so should be fairly quick

PR #152, opened 2018-11-08
Add dropped frames count to FrameReceiver

timcnicholls:
This PR address #150, adding a dropped frames count to the `FrameReceiver::FrameDecoder` class and reporting that value in the status response. 

It is the responsibility of the derived decoder plugins to increment the `frames_dropped_` member variable as necessary, since buffer management is implementation dependent.


ulrikpedersen:
Such a "clean" PR - not messing around in nested bits of code here! :-)

PR #157, opened 2018-11-20
Add MetaWriter application to odin-data

MattTaylorDLS:
Added the application to run the Meta Writer to odin-data.
Improved the dynamic class loading


tools/python/odin_data/meta_writer/__init__.py
```diff
@@ -0,0 +1,4 @@
+from pkg_resources import require
+require('pygelf==0.3.1')
+require("h5py==2.8.0")
+require('pyzmq==16.0.2')
```
timcnicholls: File missing a trailing newline

MattTaylorDLS:
Nope. To the adapter it will be, but the startup script will need to change, to run it from the different location, and also supply some different command line arguments.
It'll need to specify the log server address, and the "path" to the Eiger/Excalibur writer module.

PR #160, opened 2018-12-04
Ensure Meta flush variables are integers

MattTaylorDLS:


MattTaylorDLS:
It can be used instead of the flush frequency to flush every N seconds, instead of flushing every N frames. Can probably leave it for now, but might need to put it in by the time Malcolm is using odin

PR #162, opened 2018-12-17
Live view count fix

ANeaves:
Small change to the way the Live View plugin counts frames for the downscale filter. If the frames are filtered by tags and/or datasets, then it will only count frames that match those filters first. This should more closely match how you'd expect these filters to interact with one another.


ANeaves:
Should I merge this now?

Issue #105, opened 2018-04-12
Inconsistent logger naming in FrameProcessor


timcnicholls:
The log4cxx logger hierarchy (dotted naming convention) in the FrameProcessor inconsistently uses FP.xxxx and the legacy fileWriter FW.xxxx . This means not all messages get logged to the appropriate appenders unless both hierarchies are defined in the logging configuration. 

Example:
```
10:08:59,374  FP.Acquisition TRACE - Master frame processed
10:08:59,374  FP.Acquisition INFO  - Closing file /tmp/test.hdf5
10:08:59,374  FP.FrameProcessorController TRACE - Meta RX thread called: {"plugin":"Acquisition","parameter":"writeframe","type":"string","header":{"acqID":"0"}}
10:08:59,383  FW.Frame       TRACE - Frame destroyed
10:08:59,383  FP.FrameProcessorController TRACE - Meta RX thread called: {"plugin":"Acquisition","parameter":"closefile","type":"string","header":{"acqID":"0"}}
10:08:59,383  FP.DataBlockPool DEBUG - Releasing DataBlock [id=3]
10:08:59,383  FP.FrameProcessorController TRACE - Meta RX thread called: {"plugin":"Acquisition","parameter":"stopacquisition","type":"string","header":{"acqID":"0"}}
10:08:59,576  FW.SharedMemoryController DEBUG - RX thread called with message: {"params":{"frame":9381,"buffer_id":4},"msg_type":"notify","msg_val":"frame_ready","timestamp":"2018-04-12T10:08:59.575798"}
10:08:59,576  FW.Frame       TRACE - Frame constructed
10:08:59,576  FW.PercivalProcessPlugin TRACE - Processing raw frame.
10:08:59,576  FW.PercivalProcessPlugin TRACE - Raw frame number: 9381 offset frame number: 3
10:08:59,576  FW.Frame       TRACE - Frame constructed
10:08:59,582  FW.Frame       TRACE - copy_data called with size: 4178944 bytes
10:08:59,582  FP.DataBlockPool DEBUG - Requesting DataBlock of 4178944 bytes
```

timcnicholls:
@ajgdls @ulrikpedersen I realised we already had this issue open on logger hierarchy naming as discussed yesterday. Pretty simple to fix - would you like me to take this on?

PR #168, opened 2019-01-11
Fix inconsistent logger hierarchy names in frameProcessor

timcnicholls:
Addresses issue #105 - all log4cxx loggers in the frameProcessor now have the `FP.` prefix, rather than the mix of `FP.` and `FW.`. 

As part of this fix the canonical log4cxx config files for frameReceiver, frameProcessor and frameSimulator have been tidied up to be consistent and to suppress expensive source location format specifiers.


GDYendell:
@ajgdls and @MattTaylorDLS, we still have these conversion patterns in the config files generated by ADOdin. Do we want to remove them there, too?

ajgdls:
@GDYendell I'm not sure, probably not.  If it saves processing then it doesn't take too long to figure out where messages have come from most of the time.

PR #170, opened 2019-01-24
Update travis blosc version to 1.14.2.

ajgdls:
This is the reason the build was failing either on DLS build system or on travis, the Blosc versions were different and had breaking API changes between the two versions.


ajgdls:
Nudge @skyfrench @MattTaylorDLS are you happy with these changes?

Issue #171, opened 2019-01-30
Live View plugin doesn't load correctly


ajgdls:
If the default port of 5020 is already in use then I think the live view plugin fails to complete its initialisation.
This happens before it is possible to send a control or init message to the plugin telling it to use a different address.

ANeaves:
Fixed with PR https://github.com/odin-detector/odin-data/pull/172

PR #172, opened 2019-02-04
Live view startup fix

ANeaves:
A fix for https://github.com/odin-detector/odin-data/issues/171 , so that it no longer crashes if trying to bind to a socket address that is already in use. Will warn the user if the socket is unbound after initialization and when attempting to process a frame of data.

### DIFF
--- a/frameProcessor/src/LiveViewPlugin.cpp
+++ b/frameProcessor/src/LiveViewPlugin.cpp
@@ -45,7 +45,7 @@ LiveViewPlugin::LiveViewPlugin() :
 
   if(!is_bound_)
   {
-    LOG4CXX_WARN(logger_, "Socket is unbound. Check if default address " << DEFAULT_IMAGE_VIEW_SOCKET_ADDR << " is already in use");
+    LOG4CXX_WARN(logger_, "Socket is unbound after initilization. Check if default address " << DEFAULT_IMAGE_VIEW_SOCKET_ADDR << " is already in use");
   }
 }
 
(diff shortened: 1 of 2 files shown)


ANeaves:
Yes, the plugin will always start up with the default socket address and then re-bind it when it is configured. This was done so that the configuration for the address is optional, and because the socket needs to be created in the constructor method.

ajgdls:
My only concern would be the generation of the error message when starting up, even if a perfectly valid address is then applied as part of the initial JSON configuration.
The error message could be misinterpreted as an actual error when in fact the final ready state of the system would not contain any errors.
Do you think it is possible to suppress the initial binding until the first frame arrives, so that there has been time to re-configure? Or perhaps there is some other way to avoid the error message during the very first initialisation?

ANeaves:
Binding can't be delayed right until a frame needs to be sent, as there needs to be time between telling a ZMQ socket to bind to a socket and it actually being ready. It could be delayed until a configuration message is received, but then there is an issue if you don't configure any part of the live view it won't have a chance to set up the socket as the configure method never gets called.

The message could be demoted to a warning rather than an error, like with the other messages that are caused by having an unbound socket?

ANeaves:
Socket State is being tracked now, and it emits a warning if process_frame is called with an unbound socket.

ajgdls:
OK yes I think I like that idea.  @timcnicholls is right we will always send a configuration to set up the socket, and this configuration will be supplied as part of the initialisation JSON file.

frameProcessor/src/LiveViewPlugin.cpp
```diff
@@ -38,9 +39,13 @@ LiveViewPlugin::LiveViewPlugin() :
 
   set_frame_freq_config(DEFAULT_FRAME_FREQ);
   set_per_second_config(DEFAULT_PER_SECOND);
-  set_socket_addr_config(DEFAULT_IMAGE_VIEW_SOCKET_ADDR);
   set_dataset_name_config(DEFAULT_DATASET_NAME);
   set_tagged_filter_config(DEFAULT_TAGGED_FILTER);
+
+  if(!is_bound_)
```
timcnicholls: I think you're still emitting this warning in the constructor, which will now happen every time the plugin is loaded, since is_bound_ is initialised to false by the member init list. I'd suggest this is redundant with the new 'bind on configure' approach. You can emit a warning the first time a frame is pushed if the socket is still unbound at that point maybe.

PR #174, opened 2019-02-15
Gap fill plugin

ajgdls:
Generic plugin for inserting gaps into datasets.

### DIFF
--- a/frameProcessor/include/GapFillPlugin.h
+++ b/frameProcessor/include/GapFillPlugin.h
@@ -1,0 +1,67 @@
+/*
+ * GapFillPlugin.h
+ *
+ *  Created on: 5 Feb 2019
+ *      Author: gnx91527
+ */
+
+#ifndef ODINDATA_GAPFILLPLUGIN_H
+#define ODINDATA_GAPFILLPLUGIN_H
+
+#include <log4cxx/logger.h>
+#include <log4cxx/basicconfigurator.h>
+#include <log4cxx/propertyconfigurator.h>
+#include <log4cxx/helpers/exception.h>
+using namespace log4cxx;
+using namespace log4cxx::helpers;
+
+
+#include "FrameProcessorPlugin.h"
+#include "ClassLoader.h"
+
+namespace FrameProcessor
+{
+
+    class GapFillPlugin : public FrameProcessorPlugin
+    {
+
+    public:
+
+        GapFillPlugin();
+        virtual ~GapFillPlugin();
+        void process_frame(boost::shared_ptr<Frame> frame);
+        boost::shared_ptr<Frame> insert_gaps(boost::shared_ptr<Frame> frame);
+        void configure(OdinData::IpcMessage& config, OdinData::IpcMessage& reply);
+        int get_version_major();
+        int get_version_minor();
+        int get_version_patch();
+        std::string get_version_short();
+        std::string get_version_long();
+
+        /*Config Names*/
+        /** The required grid for the image [y, x]*/
+        static const std::string CONFIG_GRID_SIZE;
+        /** The chip size of the image in pixels [y, x]*/
+        static const std::string CONFIG_CHIP_SIZE;
+        /** The gaps to insert in the x grid direction (must be grid[x] + 1 in dimension*/
+        static const std::string CONFIG_GRID_X_GAPS;
+        /** The gaps to insert in the y grid direction (must be grid[y] + 1 in dimension*/
+        static const std::string CONFIG_GRID_Y_GAPS;
+
+    private:
+
+        void requestConfiguration(OdinData::IpcMessage& reply);
+
+        /** Pointer to logger */
+        LoggerPtr logger_;
+
+        std::vector<int> grid_;
+        std::vector<int> chip_;
+        std::vector<int> gaps_x_;
+        std::vector<int> gaps_y_;
+
+    };
+
+}/* namespace FrameProcessor */
+
+#endif //ODINDATA_GAPFILLPLUGIN_H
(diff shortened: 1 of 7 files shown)


frameProcessor/src/GapFillPlugin.cpp
```diff
@@ -0,0 +1,307 @@
...
+     * Constructor for this class.
+     */
+    GapFillPlugin::GapFillPlugin()
+    {
+        logger_ = Logger::getLogger("FP.GapFillPlugin");
+        logger_->setLevel(Level::getAll());
+        LOG4CXX_INFO(logger_, "GapFillPlugin version " << this->get_version_long() << " loaded");
+    }
+
+    GapFillPlugin::~GapFillPlugin()
```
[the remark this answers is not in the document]
ajgdls: Hi @ANeaves yes the destructor is empty in this case.  I'll add a comment to mention this.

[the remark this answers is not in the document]
ANeaves: I figured it was supposed to be. Sorry for the nitpick 👍

frameProcessor/src/GapFillPlugin.cpp
```diff
@@ -0,0 +1,307 @@
...
+     */
+    void GapFillPlugin::process_frame(boost::shared_ptr<Frame> frame)
+    {
+        LOG4CXX_TRACE(logger_, "GapFillPlugin Process Frame.");
+
+        // Call the insert gaps method and push the resulting frame if it is not null
+        boost::shared_ptr<Frame> gap_frame = this->insert_gaps(frame);
+        if (gap_frame){
+            this->push(gap_frame);
+        }
```
GDYendell: Could we have a configuration_valid method that does the top half of insert_gaps. The result of this could be checked instead of checking the frame is null. I also wonder if we should push the original frame if it is not possible to add gaps.

frameProcessor/src/GapFillPlugin.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+            // Malloc the required memory
+            void *new_image = malloc(img_x * img_y * frame->get_data_type_size());
+            // Memset to ensure empty values
+            memset(new_image, 0, img_x * img_y * frame->get_data_type_size());
+
+            // Loop over the y grid
+            int current_gap_y = 0;
+            for (int y_index = 0; y_index < grid_[0]; y_index++) {
+                current_gap_y += gaps_y_[y_index];
```
GDYendell: I would use current_offset_y/x instead of gap. Below you use the word offset to mean the same thing and it is clearer.

ajgdls: Gap and offset do not mean the same thing in this context.  Offset refers to the pointer offset and there is one for src and destination pointers.  Current_gap_y really does mean the total gap pixels in y so far, this is not the same as the offset which also includes the y size of the chip * number of chips traversed so far.

GDYendell: True, it isn't exactly the same, but I think offset is a better name. Current gap sounds (to me) like it means the gap between the current chip and the next, but it is really a cumulative offset. The offset for the chip in the dest compared to the src.

frameProcessor/src/GapFillPlugin.cpp
ajgdls: I don't think we should push the original frame.  The writer will not be setup to handle it, and we don't push the raw frame if the process plugin fails.
I'll implement the configurarion_valid method.

ajgdls: p.s. we still need to check for a null frame just as it would be immoral not to do so :)

GDYendell: Good point, the dimensions will be wrong.

frameProcessor/include/GapFillPlugin.h
```diff
@@ -0,0 +1,68 @@
+/*
+ * GapFillPlugin.h
+ *
+ *  Created on: 5 Feb 2019
+ *      Author: gnx91527
```
MattTaylorDLS: Think we said we were going to put real names in the comments

frameProcessor/src/GapFillPlugin.cpp
```diff
@@ -0,0 +1,325 @@
...
+        }
+        // Create the return frame
+        dimensions_t img_dims(2);
+        img_dims[0] = img_y;
+        img_dims[1] = img_x;
+        gap_frame = boost::shared_ptr<FrameProcessor::Frame>(new FrameProcessor::Frame(frame->get_dataset_name()));
+        gap_frame->set_frame_number(frame->get_frame_number());
+        gap_frame->set_dimensions(img_dims);
+        gap_frame->copy_data(static_cast<void*>(new_image), img_x * img_y * frame->get_data_type_size());
+        gap_frame->set_data_type(frame->get_data_type()); // 0: UINT8, 1: UINT16, 2: UINT32
```
MattTaylorDLS: does this comment need to be here? looks like a copy and paste thing

ajgdls: Well I guess not, I thought it was useful to see the types so left it in.  I can remove it.

frameProcessor/include/GapFillPlugin.h
ajgdls: I never would have agreed to my real name, not after 15 years of support requests for old code ;)
I can change it.

ajgdls:
@timcnicholls as you are here today I've added you for review of this plugin.

ajgdls:
@ANeaves @MattTaylorDLS are you now happy with these changes?

ajgdls:
I'm going to merge this now so that @timcnicholls is free to make the release.

PR #175, opened 2019-02-15
Update FrameReceiverController to retain decoder config between rebuilds

GDYendell:
Add update method to IPCMessage to add / overwrite parameters from
  from another IPCMessage
Update FrameReceiverController to call update when a different decoder
  config is received, while retaining any previous config that has not
  been overridden, instead of setting them to defaults.
Add unit test for update method

### DIFF
--- a/frameReceiver/src/FrameReceiverController.cpp
+++ b/frameReceiver/src/FrameReceiverController.cpp
@@ -461,7 +461,16 @@ void FrameReceiverController::configure_frame_decoder(OdinData::IpcMessage& conf
         new IpcMessage(config_msg.get_param<const rapidjson::Value&>(CONFIG_DECODER_CONFIG)));
     if (*new_decoder_config != *(config_.decoder_config_))
     {
-      config_.decoder_config_->update(*new_decoder_config);
+      if (new_decoder_class(config_msg))
+      {
+        // We have been asked to load a different decoder class - Replace entire configuration
+        config_.decoder_config_.swap(new_decoder_config);
+      }
+      else
+      {
+        // Just update the configuration for the currently loaded decoder
+        config_.decoder_config_->update(*new_decoder_config);
+      }
       LOG4CXX_DEBUG_LEVEL(3, logger_,
         "Built new decoder configuration message: " << config_.decoder_config_->encode()
       );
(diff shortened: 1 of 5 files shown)


GDYendell:
This fixes the problem I was having by only copying the parameters over. I wonder if `update()` should copy over (or redefine) anything else, such as timestamp and msg_id. Thoughts?

frameReceiver/test/IpcMessageUnitTest.cpp
```diff
@@ -113,6 +113,52 @@ BOOST_AUTO_TEST_CASE( CreateAndModifyParametersInEmptyIpcMessage )
...
+  BOOST_CHECK_EQUAL(msg1.get_param<int>("paramInt3"), paramIntVal3);
+  BOOST_CHECK_EQUAL(msg1.get_param<std::string>("paramStr"), paramStringVal);
+
+  OdinData::IpcMessage msg2;
+
+  // Create a different message and update the first with it
+  int newParamIntVal2 = 90310;
+  int paramIntVal4 = 42;
+  std::string newParamStringVal = "newParamStr";
+  std::string paramStringVal2("paramStr2");
```
ajgdls: Slight inconsistency here when defining strings.  Both are valid of course but just looks a bit odd using both ways right after each other.

GDYendell: Good spot.

ajgdls:
Good point @timcnicholls .  I think the one thing missing from this PR description is why it was raised.  We found that when switching to use JSON files to load most of the config at initialisation, we then couldn't then change the bit depth without having to re-send all of the required config (specifically the port maps I think).
For your comment I think this would be a potential issue only if we were re-loading a decoder for a different detector?

GDYendell:
We should probably either

- Not allow changing the decoder at runtime at all

or 

- Check for decoder_type / decoder_path changing and then use `swap` as before.

timcnicholls:
Yes @ajgdls it's an edge case to be honest. It could be that for some reason a detector has different decoders for different operational modes because there is a fundamental difference in the data format or packet structure. 

I think we also need to keep sight of the fact that the DLS trend for 'JSON file at initialisation' isn't the only operation mode for odin-data and we deliberately allowed for partial configuration plus successive config messages to bring the system into a fully configured state. I would be reluctant to lose this capability. @gdy for this reason I would not vote for preventing decoder reconfig at runtime as per your last comment.

frameReceiver/src/FrameReceiverController.cpp
```diff
@@ -534,6 +543,32 @@ void FrameReceiverController::configure_frame_decoder(OdinData::IpcMessage& conf
...
+  {
+    std::string new_path = config_msg.get_param<std::string>(CONFIG_DECODER_PATH);
+    if (config_.decoder_path_ != new_path)
+    {
+      return true;
+    }
+  }
+  if (config_msg.has_param(CONFIG_DECODER_TYPE))
+  {
+    std::string new_path = config_msg.get_param<std::string>(CONFIG_DECODER_TYPE);
```
timcnicholls: Should be `new_type` rather than `new_path` ??

GDYendell: Thanks!

GDYendell:
That was my preference. This should now handle a new decoder without passing any previous configuration onto it. I think this is important as config items may match between classes and a user might be expecting the defaults for the new class.

Suggestions for a better name for the new method are welcome...

ajgdls:
I think this is the correct way.  If you are updating the configuration of the current decoder you should be able to change a single parameter.  If you are loading a new decoder then you are starting again with a new parameter set.
To me this sounds right and is not specific to the way Diamond will configure these applications.

timcnicholls:
Agreed, I think this is a good solution. Happy for this to be merged.

PR #176, opened 2019-02-21
WIP: Frame refactor (into IFrame, DataBlockFrame & SharedBufferFrame)

skyfrench:
Initial outline of IFrame, DataBlockFrame and SharedBufferFrame classes for discussion.


frameProcessor/src/SharedMemoryController.cpp
```diff
@@ -177,15 +178,18 @@ void SharedMemoryController::handleRxChannel()
...
-          boost::shared_ptr<Frame> frame;
-          frame = boost::shared_ptr<Frame>(new Frame("raw"));
-          frame->wrap_shared_buffer(sbm_->get_buffer_address(bufferID),
-                                    sbm_->get_buffer_size(),
-                                    bufferID,rxMsg.get_param<int>("frame", 0),
-                                    &txChannel_);
+          int frame_number = rxMsg.get_param<int>("frame", 0);
+
+          FrameProcessor::IFrameMetaData frame_meta("raw",
+                                                    FrameProcessor::raw_16bit,
```
skyfrench: Previously Frame stored the data type as an integer (-1 unset, 0 int8, 1 int16 and 2 int32). Now the IFrame stores the data type as a member of its IFrameMeta instance using the pre-existing DataType enum (in FrameProcessorDefinitions.h). This enum doesn't have an 'unset'.

Here in the SharedMemoryController the Frame previously had an unset data type (and compression). In the current refactor it can't have an unset type ... 

... can a SharedBufferFrame ever have a data type or compression?
If not is there a difference in the meta data required by a DataBlockFrame and a SharedBufferFrame. Should 'unset' be added to the enums? Maybe data type and compression shouldn't be part of the IFrameMetaData but back to being members of DataBlockFrame (but not SharedBufferFrame).
@GDYendell had been wondering what should be part of 'IFrameMetaData' vs 'IFrame' ...

frameProcessor/src/BloscPlugin.cpp
```diff
@@ -145,17 +144,9 @@ boost::shared_ptr<Frame> BloscPlugin::compress_frame(boost::shared_ptr<Frame> sr
...
-  LOG4CXX_DEBUG_LEVEL(3, logger_, "Copying compressed data to output frame. (" << compressed_size << " bytes)");
-  // I wish we had a pointer swap feature on the Frame class and avoid this unnecessary copy...
-  dest_frame->copy_data(dest_data_ptr, compressed_size);
+  dest_frame->resize(compressed_size);
 
-  // I wish we had a shallow-copy feature on the Frame class...
-  dest_frame->set_data_type(src_frame->get_data_type());
-  dest_frame->set_frame_number(src_frame->get_frame_number());
-  dest_frame->set_acquisition_id(src_frame->get_acquisition_id());
-  dest_frame->set_dimensions(src_frame->get_dimensions());
```
skyfrench: Should this (old code) have also been setting the Frame compression type? dest_frame->set_compression(3)? (3 = Blosc)

frameProcessor/src/SharedMemoryController.cpp
timcnicholls: From the frameReceiver's perspective, there is no standard labelling of shared buffer frames with a data type or compression flag. The shared buffer frame payload may also not map onto one of the supported data types until unpacking/reordering by a FP plugin. For instance, tightly packed and bit-scrambled data from systems like PERCIVAL may have 12  bit pixel data walking across byte boundaries. It may not make sense therefore in call cases for the SharedBufferFrame 'wrapper' to require these to be set.

However, if we consider the EIGER 'pass-through' mode of passing the SharedBufferFrame directly through to the HDF plugin, maybe those fields have to be present (or set early in the plugin chain) to allow the frame to be written. So perhaps those fields need to be present, have a default unset or 'raw' value in the enum, and have a mechanism for them to be set explicitly in this pass-through mode?

This also touches on the desire to have a generic frame header populated on all frame types by the frame receiver to track e.g. receive time on the frame. Perhaps we should also add these fields to that and then have the above code use them to set the SharedBufferFrame fields?

I'm rambling :wink: - perhaps we can talk about this at the meeting?

frameProcessor/src/SumPlugin.cpp
```diff
@@ -45,26 +45,25 @@ namespace FrameProcessor {
...
+  void SumPlugin::process_frame(boost::shared_ptr <IFrame> frame)
   {
     LOG4CXX_TRACE(logger_, "Received a new frame...");
-    uint64_t sum_val = 0;
-    switch (frame->get_data_type()) {
+    switch (frame->get_meta_data().get_data_type()) {
       case raw_8bit:
-        sum_val = calculate_sum<uint8_t>(frame);
-        frame->set_parameter(SUM_PARAM_NAME, sum_val);
+        frame->meta_data().set_parameter<uint8_t>(SUM_PARAM_NAME, calculate_sum<uint8_t>(frame));
```
GDYendell:
```suggestion
        frame->meta_data().set_parameter<uint64_t>(SUM_PARAM_NAME, calculate_sum<uint8_t>(frame));
```

frameProcessor/src/Frame.cpp
```diff
@@ -1,479 +1,132 @@
...
- * 0: UINT8, 1: UINT16, 2: UINT32
- *
- * \param data type.
- */
-void Frame::set_data_type(int data_type)
-{
-  data_type_ = data_type;
-}
+    void *Frame::get_image_ptr() const {
+      return this->get_data_ptr() + image_offset_;
```
timcnicholls: EL7 compiler warns about void pointer arithmetic at this point:
```
odin-data/frameProcessor/src/Frame.cpp: In member function virtual void* FrameProcessor::Frame::get_image_ptr() const:
odin-data/frameProcessor/src/Frame.cpp:56:37: warning: pointer of type void used in arithmetic [-Wpointer-arith]
       return this->get_data_ptr() + image_offset_;
```

timcnicholls:
Tested working OK with EXCALIBUR chain running  10 frames PCAP file through simulator:
Screenshot 2019-04-30 at 14 28 02

frameProcessor/include/Frame.h
```diff
@@ -1,226 +1,84 @@
...
-  ParameterValue value;
-  /** The type of this parameter */
-  FrameProcessor::DataType type;
-} Parameter;
+ public:
 
-typedef unsigned long long dimsize_t;
-typedef std::vector<dimsize_t> dimensions_t;
+  /** Base constructor */
+  Frame(const FrameMetaData &meta_data, const size_t &data_size, const int &image_offset = 0);
```
timcnicholls: The Frame (and subclass) non-copy constructors now require a FrameMetaData object as an argument rather than e.g. a trivial list of dataset name and size. I can see the sense in this, although I wonder if it would leave us with a more intuitive interface if there was the option to lazily construct a frame and set the metadata arguments afterwards? This may be problematic - happy to hear opinions.

frameProcessor/include/FrameProcessorDefinitions.h
```diff
@@ -8,16 +8,20 @@
...
+#include <stdexcept>
+#include "stdint.h"
+
 namespace FrameProcessor
 {
   /**
    * Enumeration to store the pixel type of the incoming image
    */
-  enum DataType { raw_8bit, raw_16bit, raw_32bit, raw_64bit, raw_float };
+  enum DataType { raw_unknown, raw_8bit, raw_16bit, raw_32bit, raw_64bit, raw_float };
```
timcnicholls: The enumerations of `DataType` and `CompressionType` now have an initial 'unknown' entry. This shifts the values needed in the FP dataset definitions passed to the HDF plugin as integers:
```
{
        "hdf": {
          "dataset": {
            "data": {
              "datatype": 2,   <- was 1
              "dims": [256, 2048],
              "chunks": [1, 256, 2048],
              "compression": 1  <- was 0
            }
          }
        }
      },
```
This is a fairly trivial change to make once but is a compatibility issue. It occurs to me that having them as ints in the dataset definition is opaque. Perhaps we should resolve them back during config parsing from strings?

frameProcessor/include/DataBlock.h
```diff
@@ -28,32 +28,57 @@ namespace FrameProcessor
...
-  /** Number of bytes allocated for this DataBlock */
-  size_t allocatedBytes_;
-  /** Unique index of this DataBlock */
-  int index_;
-  /** Void pointer to the allocated memory */
-  void *blockPtr_;
-  /** Static counter for the unique index */
-  static int indexCounter_;
-};
+    class DataBlock {
```
MattTaylorDLS: Looks like the formatting here has been changed to 4 spaces. Should be 2

frameProcessor/include/Frame.h
```diff
@@ -1,226 +1,84 @@
...
+  FrameMetaData &meta_data();
+
+  /** Return the MetaData */
+  const FrameMetaData &get_meta_data() const;
+
+  /** Return deep copy of the MetaData */
+  FrameMetaData get_meta_data_copy() const;
+
+  /** Return the image offset */
+  int get_image_offset() const;
```
MattTaylorDLS: Can this be moved below the set_meta_data() so that it's next to the set_image_offset() ?

frameProcessor/include/FrameProcessorDefinitions.h
```diff
@@ -28,6 +32,29 @@ namespace FrameProcessor
   const std::string DATA_TYPES[] = {"uint8","uint16","uint32","uint64","float"};
```
MattTaylorDLS: This needs to be updated to have the unkown type at the front so they match the DataType enum

frameProcessor/include/FrameProcessorDefinitions.h
```diff
@@ -28,6 +32,29 @@ namespace FrameProcessor
...
   const std::string COMPRESS_TYPES[] = {"none","LZ4","BSLZ4"};
```
MattTaylorDLS: This needs to be updated to have the unknown type at the front so that it matches the CompressionType enum

frameProcessor/include/FrameProcessorDefinitions.h
MattTaylorDLS: A matching enum is used in the live view plugin tests, which will possibly need to be updated as they dont have the unknown entry at the start

frameProcessor/src/Acquisition.cpp
```diff
@@ -382,39 +384,42 @@ void Acquisition::stop_acquisition() {
...
+  const FrameMetaData frame_meta_data = frame->get_meta_data();
+  DatasetDefinition dataset = dataset_defs_.at(frame_meta_data.get_dataset_name());
+  CompressionType frame_compression_type = frame_meta_data.get_compression_type();
+  if (frame_compression_type >= 0 && frame_compression_type != dataset.compression) {
     std::stringstream ss;
-    ss << "Invalid frame: Frame has compression " << frame->get_compression() <<
+    ss << "Invalid frame: Frame has compression " << frame_compression_type <<
           ", expected " << dataset.compression <<
           " for dataset " << dataset.name <<
           " (0: None, 1: LZ4, 2: BSLZ4, 3: Blosc)";
```
MattTaylorDLS: Need to update this log message to account for unknown compression type

frameProcessor/src/Acquisition.cpp
```diff
@@ -382,39 +384,42 @@ void Acquisition::stop_acquisition() {
...
   }
-  if (frame->get_data_type() >= 0 && frame->get_data_type() != dataset.data_type) {
+  DataType frame_data_type = frame_meta_data.get_data_type();
+  if (frame_data_type >= 0 && frame_data_type != dataset.data_type) {
     std::stringstream ss;
-    ss << "Invalid frame: Frame has data type " << frame->get_data_type() <<
+    ss << "Invalid frame: Frame has data type " << frame_data_type <<
        ", expected " << dataset.data_type <<
        " for dataset " << dataset.name <<
        " (0: UINT8, 1: UINT16, 2: UINT32, 3: UINT64, 4: FLOAT)";
```
MattTaylorDLS: need to update this log message to account for unknown data type

frameProcessor/src/BloscPlugin.cpp
MattTaylorDLS: yes

frameProcessor/src/DataBlockPool.cpp
```diff
@@ -25,12 +25,12 @@ DataBlockPool::~DataBlockPool()
  * the pool specified by the index parameter.
  *
  * \param[in] index - Index of DataBlockPool to add new DataBlocks to.
```
MattTaylorDLS: params in comment don't match the actual params

frameProcessor/src/DataBlockPool.cpp
```diff
@@ -41,9 +41,9 @@ void DataBlockPool::allocate(const std::string& index, size_t nBlocks, size_t nB
  * \param[in] nBytes - Size of the DataBlock required in bytes.
```
MattTaylorDLS: params in comment don't match the actual params

frameProcessor/src/DataBlockPool.cpp
```diff
@@ -54,9 +54,9 @@ boost::shared_ptr<DataBlock> DataBlockPool::take(const std::string& index, size_
  * \param[in] index - Index of DataBlockPool to take the DataBlocks from.
```
MattTaylorDLS: params in comment don't match the actual params

frameProcessor/src/DataBlockPool.cpp
```diff
@@ -66,9 +66,9 @@ void DataBlockPool::release(const std::string& index, boost::shared_ptr<DataBloc
  * \param[in] index - Index of DataBlockPool to get the free count from.
```
MattTaylorDLS: params in comment don't match the actual params

frameProcessor/src/DataBlockPool.cpp
```diff
@@ -78,9 +78,9 @@ size_t DataBlockPool::getFreeBlocks(const std::string& index)
  * \param[in] index - Index of DataBlockPool to get the in-use count from.
```
MattTaylorDLS: params in comment don't match the actual params

frameProcessor/src/DataBlockPool.cpp
```diff
@@ -90,9 +90,9 @@ size_t DataBlockPool::getUsedBlocks(const std::string& index)
  * \param[in] index - Index of DataBlockPool to get the total count from.
```
MattTaylorDLS: params in comment don't match the actual params

frameProcessor/src/DataBlockPool.cpp
```diff
@@ -102,9 +102,9 @@ size_t DataBlockPool::getTotalBlocks(const std::string& index)
  * \param[in] index - Index of DataBlockPool to get the total bytes allocated from.
```
MattTaylorDLS: params in comment don't match the actual params

frameProcessor/src/Frame.cpp
```diff
@@ -1,479 +1,132 @@
...
+    FrameMetaData Frame::get_meta_data_copy() const {
+      return this->meta_data_;
+    }
 
-  return parameter.value.float_val;
-}
+/** Return the image offset
+ * @return offset from beginning of data memory to image data
+ * */
+    int Frame::get_image_offset() const {
```
MattTaylorDLS: can this be moved below set_meta_data() so that it's next to set_image_offset?

frameProcessor/src/Frame.cpp
```diff
@@ -1,479 +1,132 @@
...
-  }
-  return iter->second;
-}
+/** Update the data size
+ * @param size new data size
+ * */
+    void Frame::set_data_size(size_t size) {
+      if (size > this->data_size_)
+        throw std::runtime_error("Can't set frame data size to be bigger than allocated memory size.");
+      LOG4CXX_INFO(logger_, "Updating size of frame data");
```
MattTaylorDLS: This should be TRACE (or DEBUG?) instead of INFO as it's potentially called every frame.

frameProcessor/src/FrameMetaData.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+    /** Return acquistion ID
+     * @return std::string acquisition ID
+     */
+    const std::string &FrameMetaData::get_acquisition_ID() const {
+      return this->acquisition_ID_;
+    }
+
+    /** Set acquisition ID
+     * @param std::string acquisition id
```
MattTaylorDLS: params in comment don't match the actual params

frameProcessor/src/FrameMetaData.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+    /** Return compression type
+     * @return compression type
+     */
+    CompressionType FrameMetaData::get_compression_type() const {
+      return this->compression_type_;
+    }
+
+    /** Set compression type
+     * @ param CompressionType compression type
```
MattTaylorDLS: params in comment don't match the actual params

frameProcessor/src/HDF5File.cpp
```diff
@@ -223,7 +223,7 @@ void HDF5File::write_frame(const Frame& frame, hsize_t frame_offset, uint64_t ou
   uint32_t filter_mask = 0x0;
   ensure_h5_result(H5DOwrite_chunk(dset.dataset_id, H5P_DEFAULT,
       filter_mask, &offset.front(),
-      frame.get_data_size(), frame.get_data()), "H5DOwrite_chunk failed");
+      frame.get_data_size(), frame.get_data_ptr()), "H5DOwrite_chunk failed");
```
MattTaylorDLS: Should this be get_image_ptr() instead of get_data_ptr()?
Same applies to other places where get_data_ptr is used (e.g. in other plugins)

skyfrench: Good point. Calling get_data_size is also not appropriate as this still includes the header; I need to update this when set_image_offset is called.
Would it be cleaner to have just one method instead of both get_data_ptr and get_image_ptr? Is there any need to have a get_data_ptr method - should a plugin ever call it? Or should get_data_ptr do what get_image_ptr is trying to do and always return without the header?

timcnicholls: I believe we *do* need get_data_ptr() and get_data_size() methods - plugins often need to extract information from the frame header. See the EXCALIBUR processing plugin as an example. It's only when we get to the file writing that you need to be able to offset beyond any existing header to ensure only image data is written to HDF.

GDYendell: Does anything other than the detector ProcessPlugin need the header, though? All the other plugins should be detector agnostic, so they wouldn't understand the header.

timcnicholls: That's correct but the detector plugins **do** need access to it, so we shouldn't remove the methods. They are completely generic information about the frame as an abstract container for data. The alternative would be to have only the get_image_xxx methods and have to do pointer arithmetic subtraction to get a pointer to the header. 

I believe when we decided to implement get_image_ptr() and get_image_size() we agreed that these would default to returning the same values is get_data_xxx methods unless a call to set the image offset had been explicitly made by an upstream plugin. 

I'm not arguing against the change @MattTaylorDLS originally requested above, just to retain the get_data_xxx methods.

GDYendell: An alternative would be that the Process plugins would call get_data_ptr to extract what they need from the header and then set the ptr offset so that further calls to get_data_ptr return only the image. I am happy with either.

timcnicholls: That's not what we agreed originally. What happens if you flexibly chain multiple process plugins that need access to the header while applying corrections to resulting image data? How will they know if the ptr offset has been manipulated? 

The mechanism we discussed was:

1) Detector processing plugins that yield image data in frame capable of being written to HDF set the image offset as appropriate. This defaults to zero so that get_data_ptr() and get_image_ptr() return the same address on image-only frames.
2) Plugins that continue to require access to header information on appropriate frame types can still get at the data pointer through the get_data_xxxx methods, cast to a header pointer as necessary and extract header fields
3) Detector agnostic plugins operating on image arrays, including the file writer, then use the get_image_xxx methods, which will either point to the start of the frame data if it only contains image arrays, or will have been offset past the header by an upstream plugin.

If my understanding of the discussion is correct, then the conclusion we reached is to retain get_data_ptr() and get_data_size() and add get_image_ptr() and get_image_size().

MattTaylorDLS: Yes I think that sounds the best option

frameProcessor/src/SharedBufferFrame.cpp
```diff
@@ -0,0 +1,49 @@
...
+  shared_id_ = frame.shared_id_;
+}
+
+/** Destroy frame
+ *
+ */
+SharedBufferFrame::~SharedBufferFrame () {
+  OdinData::IpcMessage txMsg(OdinData::IpcMessage::MsgTypeNotify,
+                             OdinData::IpcMessage::MsgValNotifyFrameRelease);
+  txMsg.set_param("frame", meta_data_.get_dataset_name());
```
timcnicholls: The `frame` field on the release notification should be the frame number, not the dataset name; this line should be:
```
txMsg.set_param("frame", static_cast<uint64_t>(meta_data_.get_frame_number()));
```

frameProcessor/src/SharedBufferFrame.cpp
```diff
@@ -31,7 +31,7 @@ SharedBufferFrame::SharedBufferFrame(const SharedBufferFrame& frame) : Frame(fra
 SharedBufferFrame::~SharedBufferFrame () {
   OdinData::IpcMessage txMsg(OdinData::IpcMessage::MsgTypeNotify,
                              OdinData::IpcMessage::MsgValNotifyFrameRelease);
-  txMsg.set_param("frame", meta_data_.get_dataset_name());
+  txMsg.set_param("frame", meta_data_.get_frame_number());
```
timcnicholls: I believe you need to explicitly cast the output of get_frame_number to `uint64_t` as in my comment above, so that it resolves to a concrete implementation of set_param. `IpcMessage` doesn't have a `set_param` method accepting `long long` arguments so the linker complains about unresolved references to templated methods if you dont:
```
make[1]: *** [frameProcessor/src/CMakeFiles/frameProcessor.dir/all] Error 2
../../lib/libFrameProcessor.so: undefined reference to `void OdinData::IpcMessage::set_value<long long>(rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >&, long long const&)'
collect2: error: ld returned 1 exit status
```

frameProcessor/src/HDF5File.cpp
MattTaylorDLS: I think we also need the ability to set_image_size(), as the image might not take up all the space in image_ptr to end of data. (as in the case of eiger)

timcnicholls:
For posterity: logging FP memory usage during acquisition of EXCALIBUR frames from PCAP with refactored Frame implemention:
fp_mem
Real memory usage climbs as shared buffer memory is accessed sequentially, then is stable. No evidence of memory leak.

GDYendell:
Would it be good to merge the Kafka Plugin and then update it to use the new API before merging this?

timcnicholls:
Actions:
@ajgdls will test this with the blosc plugin
@timcnicholls will test this with the gap fill plugin

timcnicholls:
Update: 17/5/19 Retested gap and blosc plugins after recent blosc image_size fix and still working correctly.

MattTaylorDLS:
Are we nearly ready to merge this?

skyfrench:
Maybe? :)

GDYendell:
Kafka needs to go first, but it has outstanding reviewers.

timcnicholls:
I'm happy and keen for it to be merged - all the tests we agreed have been done. 

@GDYendell If we merge #185 first, that's going to delay this merge since the Kafka plugin will have to be refactored as part of this PR. Can't we merge this then rebaseline and refactor #185?

GDYendell:
I don't think it is going to have to change much based on these changes either way. It just seemed the easier way around to do things. They are both ready to go in, so it doesn't make much difference.

timcnicholls:
@GDYendell Let's go with your proposal then if it's not going to delay this PR any further.

skyfrench:
As suggested: Kafka plugin merged into master, then this refactor rebased and the Kafka plugin updated.

timcnicholls:
Thanks @skyfrench - are we ready to merge this? Does anyone want to validate the Kafka plugin first?

hir12111:
I have tested it using a Kafka server and it works fine, I'm happy with this getting merged

timcnicholls:
Sounds like it's time to press the big green button @skyfrench? Let's blow this thing and go home 😉

PR #179, opened 2019-03-14
Stored configuration sequences

ajgdls:
This PR provides a mechanism for storing arbitrarily complex configuration sequences within an FP controller 

`"store": {"index": <index name>, "value": <sequence of configuration dictionaries>}`

that can be replayed at any time by sending the configuration 

`"execute": {"index": <index name>}`

The use case for this is to be able to tear down plugin chains and rebuild them.  This involves submission of many configurations to update each connection.  With this mechanism each plugin chain can be stored by and index and then executed with a single configuration item.  This will be required at Diamond for pulling in and out compression and gap filling plugins.

The mechanism used to store and replay is the same as for loading json config files from the command line, with the sequences of configurations stored as JSON.

NOTE: The reason for storing the sequences within the FP itself is to eliminate inconsistencies between FP applications and adapters which we have previously suffered from when applications are being re-loaded.

Open to discussion...

p.s. Also added the single configuration option to be able to disconnect a whole plugin chain without needing to know the setup of the chain.

### DIFF
--- a/frameProcessor/src/FrameProcessorController.cpp
+++ b/frameProcessor/src/FrameProcessorController.cpp
@@ -33,6 +33,7 @@ const std::string FrameProcessorController::CONFIG_PLUGIN_NAME       = "name";
 const std::string FrameProcessorController::CONFIG_PLUGIN_INDEX      = "index";
 const std::string FrameProcessorController::CONFIG_PLUGIN_LIBRARY    = "library";
 const std::string FrameProcessorController::CONFIG_PLUGIN_CONNECTION = "connection";
+const std::string FrameProcessorController::CONFIG_PLUGIN_DISCONNECT_ALL = "all";
 
 const int FrameProcessorController::META_TX_HWM = 10000;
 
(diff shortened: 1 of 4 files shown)


frameProcessor/src/FrameProcessorController.cpp
```diff
@@ -499,12 +557,24 @@ void FrameProcessorController::configurePlugin(OdinData::IpcMessage& config, Odi
...
+    {
+      OdinData::IpcMessage pluginConfig(config.get_param<const rapidjson::Value&>(FrameProcessorController::CONFIG_PLUGIN_DISCONNECT));
+      if (pluginConfig.has_param(FrameProcessorController::CONFIG_PLUGIN_CONNECTION) &&
+          pluginConfig.has_param(FrameProcessorController::CONFIG_PLUGIN_INDEX)) {
+        std::string index = pluginConfig.get_param<std::string>(FrameProcessorController::CONFIG_PLUGIN_INDEX);
+        std::string cnxn = pluginConfig.get_param<std::string>(FrameProcessorController::CONFIG_PLUGIN_CONNECTION);
+        this->disconnectPlugin(index, cnxn);
+      }
+    }
+    catch (OdinData::IpcMessageException& e)
```
timcnicholls: Why is the disconnectAll mechanism only reached via catching an exception?

ajgdls: Added comments to the code, see below.

dw96:
This looks very welcome for the Hexitec project. There are currently 7 plugins available for the FP, where a typical plugin chain would look like this:

1) Reorder 2) Threshold 3) Calibration 4) Charged Sharing Correction 5) Next Frame 6) Histogram 7) HDF

The plugin at 4) Charge Sharing can be one out of three choices (Addition, Discrimination or None, i.e. it's skipped) and 5) Next Frame may be loaded or skipped. This means there are typically 5 different chains that the user would be interested in employing. Understandably, designing a UI that can currently go from one chain to another (and teardown the existing chain before loading the next) is a bit cumbersome. With the proposed PR, would it be possible to load these 5 chains using a Json file and then simply go from one to another issuing a "disconnect all" in between? Would there be any limit on how many chains can be loaded? I suspect there may be future use cases where the number of configurations chains may grow beyond 5..

timcnicholls:
@dw96 Yes, that's it exactly what it's intended to do.

frameProcessor/src/FrameProcessorController.cpp
```diff
@@ -392,6 +398,58 @@ void FrameProcessorController::configure(OdinData::IpcMessage& config, OdinData:
...
+    }
+  }
+
+  // Check if we are being asked to execute a previously stored configuration object
+  if (config.has_param(FrameProcessorController::CONFIG_EXECUTE)) {
+    OdinData::IpcMessage storeConfig(config.get_param<const rapidjson::Value&>(FrameProcessorController::CONFIG_EXECUTE));
+
+    if (storeConfig.has_param(FrameProcessorController::CONFIG_INDEX)) {
+      std::string index = storeConfig.get_param<std::string>(FrameProcessorController::CONFIG_INDEX);
+      if (stored_configs_.count(index) > 0){
```
MattTaylorDLS: needs space between ( and {

ajgdls:
Hi @dw96 you could even add the disconnect all configuration as the first for each of your saved chains, and so a single configuration message would tear them down and re-build, instead of needing to send the disconnect all explicitly.  There is no limit, they are stored as a string indexed map and so as long as you can think of a unique index for each one you can have as many as your memory permits.

ajgdls:
Pushed format update because I want that green tick from @MattTaylorDLS ...

dw96:
> 
> 
> Hi @dw96 you could even add the disconnect all configuration as the first for each of your saved chains, and so a single configuration message would tear them down and re-build, instead of needing to send the disconnect all explicitly. There is no limit, they are stored as a string indexed map and so as long as you can think of a unique index for each one you can have as many as your memory permits.

Ahh, I didn't realise I could combine them into one message. -Thank you.

PR #181, opened 2019-03-15
Set high water mark on meta writer receive socket to reasonable value

MattTaylorDLS:
The default value is 1000. Elsewhere in odin, we've set the default value to 10000 to cope with higher frame rates, so setting the meta writer in-line with the others


GDYendell:
Can you squash these two commits?

MattTaylorDLS:
yep

PR #182, opened 2019-03-25
Add file_prefix to metalistener adapter and change from status to config

MattTaylorDLS:
During integration with ADOdin, discovered that the file_prefix should be on the request_configuration message instead of status, and that a change is needed in the adapter to make it work as desired.


ajgdls:
@MattTaylorDLS New release of odin-data required for this?  Or do we just re-release as a DLS patch?

MattTaylorDLS:
Just release as a DLS patch i think

PR #184, opened 2019-03-27
Flatten compression params

ajgdls:
This should be the last PR required for complete functionality of GapFill and compression within Excalibur (and other detectors in the future).
Added @timcnicholls as I enjoy getting that token tick :)

### DIFF
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -358,11 +358,9 @@ void FileWriterPlugin::requestConfiguration(OdinData::IpcMessage& reply)
 
     // Add the dataset compression
     reply.set_param(get_name() + "/dataset/" + iter->first + "/" + FileWriterPlugin::CONFIG_DATASET_COMPRESSION, (int)iter->second.compression);
-    if ((int)iter->second.compression == blosc) {
-      reply.set_param(get_name() + "/dataset/" + iter->first + "/" + FileWriterPlugin::CONFIG_DATASET_BLOSC_COMPRESSOR, (int)iter->second.blosc_compressor);
-      reply.set_param(get_name() + "/dataset/" + iter->first + "/" + FileWriterPlugin::CONFIG_DATASET_BLOSC_LEVEL, (int)iter->second.blosc_level);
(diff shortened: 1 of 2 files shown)


frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -358,11 +358,9 @@ void FileWriterPlugin::requestConfiguration(OdinData::IpcMessage& reply)
 
     // Add the dataset compression
     reply.set_param(get_name() + "/dataset/" + iter->first + "/" + FileWriterPlugin::CONFIG_DATASET_COMPRESSION, (int)iter->second.compression);
-    if ((int)iter->second.compression == blosc) {
-      reply.set_param(get_name() + "/dataset/" + iter->first + "/" + FileWriterPlugin::CONFIG_DATASET_BLOSC_COMPRESSOR, (int)iter->second.blosc_compressor);
-      reply.set_param(get_name() + "/dataset/" + iter->first + "/" + FileWriterPlugin::CONFIG_DATASET_BLOSC_LEVEL, (int)iter->second.blosc_level);
-      reply.set_param(get_name() + "/dataset/" + iter->first + "/" + FileWriterPlugin::CONFIG_DATASET_BLOSC_SHUFFLE, (int)iter->second.blosc_shuffle);
-    }
+    reply.set_param(get_name() + "/dataset/" + iter->first + "/" + FileWriterPlugin::CONFIG_DATASET_BLOSC_COMPRESSOR, (int)iter->second.blosc_compressor);
```
MattTaylorDLS: This to me seems a little unfortunate that it's "polluting" the FileWriterPlugin with blosc specific parameters, but i guess there's no way around it.

ajgdls: It isn't great, but in the long run we will create the Excalibur adapter to manage all of this; I don't know if we will need to expose any of them at that point as they could all be configuration options (build time).
Currently all dataset objects have these parameters even if they were being hidden from the configuration control interface, and so I have made a direct mapping of dataset parameters to configuration parameters.  They are ignored if Blosc is not chosen.

ajgdls:
Thanks everyone for getting these through so quickly.

PR #185, opened 2019-04-02
Kafka Producer plugin

hir12111:
- Implementation of a plugin that allows sending frames to one or more Kafka servers asynchronously.
- Unit tests
- Kafka cmake file
- Travis dependency installer

A few decisions were made in this implementation:

 - librdkafka C library was used instead of the C++ version as it is more flexible.
 - The message header is json (for more flexibility and easy to parse), the message structure is: [2 bytes header size] + [json header] + [frame data]
 - The header contains the following keys: "data_size", "data_type", "frame_number", "dims", "frame_offset", "acquisition_id", "compression" and optionally "parameters" (object containing frame parameters)
 - The plugin is optional
 - The message is sent without blocking (or waiting for ack)
 - Default partition is automatic partitioning (RD_KAFKA_PARTITION_UA)
 - It allows choosing which dataset will be published, defaults to "data"
 - It allows choosing in which topic will be published, defaults to "data"
 - If many topics or datasets are required, the plugin could be instantiated many times with different configuration.

I've already tested it with Tom in our kafka cluster and everything works fine.
EDIT: it has also been tested with a real Eiger detector

### DIFF
--- a/frameProcessor/include/KafkaProducerPlugin.h
+++ b/frameProcessor/include/KafkaProducerPlugin.h
@@ -20,8 +20,8 @@ using namespace log4cxx::helpers;
 #define KAFKA_ERROR_BUFFER_LEN 512
 #define KAFKA_MAX_MSG_LEN 512
 #define KAFKA_LINGER_MS 1000
-#define KAFKA_DATASET_TOPIC_SEPARATOR ":"
 #define KAFKA_DEFAULT_DATASET "data"
+#define KAFKA_DEFAULT_TOPIC   "data"
 #define KAFKA_POLLING_MS 5000
 #define KAFKA_MESSAGE_MAX_BYTES "134217728"
 #define MSG_HEADER_FRAME_SIZE_KEY "data_size"
@@ -52,6 +52,8 @@ namespace FrameProcessor {
 
     void requestConfiguration(OdinData::IpcMessage &reply);
 
+    void status(OdinData::IpcMessage &status);
+
     void process_frame(boost::shared_ptr <Frame> frame);
 
     bool reset_statistics();
@@ -69,12 +71,12 @@ namespace FrameProcessor {
 
     void configure_kafka_topic(std::string topic_name);
 
-    void configure_kafka_partition(int32_t partition);
+    void configure_partition(int32_t partition);
 
     void configure_dataset(std::string dataset);
 
     void *create_message(boost::shared_ptr<Frame> frame,
-            size_t &nbytes);
+                         size_t &nbytes);
 
     void enqueue_frame(boost::shared_ptr<Frame> frame);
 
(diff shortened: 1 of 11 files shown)


ulrikpedersen:
Thanks @hir12111! 

We want to test this with at least one of our detector systems (preferably Eiger) before merging this PR.

hir12111:
I was using Stroustrup code style in my IDE, unfortunately there isn't any predefined code style that matches this project, so I fixed it manually in the last commit

hir12111:
I tested it with the Eiger simulator and it worked fine, then, Matt and I went to VMXi and tested it with a real Eiger, the test worked and the Kafka server got the data without problems. 
In future we might try doing load testing,  by pushing as many frames as possible (once SC gets the servers ready)

frameProcessor/src/KafkaProducerPlugin.cpp
```diff
@@ -0,0 +1,419 @@
...
+    // Make sure statistics are updated
+    poll_delivery_message_report_queue();
+
+    reply.set_param(get_name() + "/" + KafkaProducerPlugin::CONFIG_SENT,
+                    frames_sent_);
+    reply.set_param(get_name() + "/" + KafkaProducerPlugin::CONFIG_LOST,
+                    frames_lost_);
+    reply.set_param(get_name() + "/" + KafkaProducerPlugin::CONFIG_ACK,
+                    frames_ack_);
+  }
```
GDYendell: The above config items should be added to the readback here.

hir12111:
I've fixed what Gary mentioned in the last commit, now all the configuration parameters are included in requestConfiguration and the stats are in status.
I have also included a new parameter for configuring the topic name (independently of the dataset name), it defaults to "data" if not set.

MattTaylorDLS:
Did we find out if these are the things that the intended kafka consumers are going to want? Is there a specification drawn up somewhere?

ulrikpedersen:
> Did we find out if these are the things that the intended kafka consumers are going to want? Is there a specification drawn up somewhere?

Not yet. That's a bit of a bee's nest to dive into and I need to clear some of my own backlog before I can properly dedicate some time to this...

frameProcessor/include/KafkaProducerPlugin.h
```diff
@@ -0,0 +1,132 @@
+//
+// Created by hir12111 on 25/03/19.
```
skyfrench: Replace fed ID with author name

frameProcessor/src/KafkaProducerPluginLib.cpp
```diff
@@ -0,0 +1,16 @@
+//
+// Created by hir12111 on 25/03/19.
```
skyfrench: Fed ID -> name

frameProcessor/src/KafkaProducerPlugin.cpp
```diff
@@ -0,0 +1,428 @@
...
+#include "rapidjson/writer.h"
+#include "rapidjson/stringbuffer.h"
+
+#include "KafkaProducerPlugin.h"
+#include "version.h"
+
+namespace FrameProcessor {
+
+  // Callback function used to report status of a delivered message
+  static void kafka_message_callback(rd_kafka_t *kafka_producer,
```
skyfrench: For all functions should there be more detailed comments? Detailed comment specifying parameters and return in source code, and single line / trailing comment in header for methods?

Issue #189, opened 2019-04-11
Do we still use the old client python code?


ajgdls:
There are some old client python files in odin-data.  This is forcing odin-data to depend on npyscreen.  I notice in at least one file there is a hardcoded path to my DLS home directory to load a percival library.

I'm guessing no-one ever uses any of this now and so I think we should clean it up and remove these kind of dependencies.

MattTaylorDLS:
I dont use them

timcnicholls:
Agreed, we should have a good spring clean of that entire package.

ajgdls:
Directory tools/python/odin_data/frame_processor

I think we should eliminate the whole directory.  There is a file frame_processor.py that is a test harness for a FrameProcessor application but I'm not sure we need that any more as we have the real frame processor application with the dummy plugin?

GDYendell:
Fixed by #191

PR #191, opened 2019-04-15
Remove old FP client file no longer used and delete npyscreen dependency

ajgdls:
Reference issue #189


GDYendell:
In #189 you suggested removing the entire directory. Is there any reason not to do that?
If not you should probably at least delete the `__main__.py`, which imports and runs this client.

ajgdls:
I had suggested we remove the entire directory, but I felt that specifically for this PR I only wanted to clean out the npyscreen reference.  However you are right there is no reason not to remove the entire directory so I'll add it to this PR.

ajgdls:
Good spot, removed those scripts.

ajgdls:
@GDYendell squashed and merged.

Issue #186, opened 2019-04-11
Processes throw generic exception if an incorrectly formatted JSON config file is specified


timcnicholls:
Currently, both the frameReceiver and the frameProcessor throw an unhelpful exception error message and terminate if started with a JSON config file that contains syntax errors:
```
ERROR FR.App null - Generic exception during frame receiver run: rapidjson assertion thrown
```
```
ERROR FP.App null - Caught unhandled exception in FrameProcessor, application will terminate: rapidjson assertion thrown
```
This can be resolved by changing the calls to `param_doc.Parse()` in the relevant app startups to something akin to:
```
if (param_doc.Parse(json.c_str()).HasParseError())
{
   std::stringstream msg;
   msg << "Parsing JSON configuration failed at offset " 
     << param_doc.GetErrorOffset() << ": " 
     << rapidjson::GetParseError_En(param_doc.GetParseError());
   throw OdinData::OdinDataException(msg.str());
}
```
There's a wider discussion about whether the app should terminate at this point or report and error and remain responsive to interrogation by a controlling client. Either way, the exception reporting can be improved.

ajgdls:
@timcnicholls that will be a huge improvement, especially if the error message really explains what the problem with the JSON file is!
I've spent a *lot* of time looking for problems only to find out I have an extra (or missing) comma in the JSON configuration file.

timcnicholls:
Yeah @ajgdls that's exactly what burnt me too. The byte offset is a start but hard to translate into e.g. a line number. Given that the parser takes the entire file as a string, I was wondering about also printing out e.g +/- x characters either side of the offset.

GDYendell:
Could it search for +/- x `\n`s either side of the offset?

timcnicholls:
That's a good idea @GDYendell, I don't see why not.

PR #192, opened 2019-04-17
print JSON parse error on app start-up

skyfrench:
Issue #186

### DIFF
--- a/common/include/CMakeLists.txt
+++ b/common/include/CMakeLists.txt
@@ -10,7 +10,8 @@ SET(HEADERS ClassLoader.h
         logging.h
         OdinDataException.h
         SegFaultHandler.h
-        SharedBufferManager.h)
+        SharedBufferManager.h
+        stringparse.h)
 SET(RAPIDJSON_INCLUDE_DIR rapidjson)
 SET(ZMQ_INCLUDE_DIR zmq)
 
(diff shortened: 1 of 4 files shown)


skyfrench:
> Is the limit a character limit, so for example you will either print up to the previous and next line or up to +- 15 characters (whichever is minimum)?

Yes. I wanted to protect against somebody having a JSON config file formatted all on one line, for example. Without the character limit it worked nicely printing the whole line; but I had nicely formatted JSON files to test with ...

ajgdls:
OK I see.  That looks fine, do you think +- 15 characters will normally get the whole line - difficult question to answer I guess (would it work for the current Excalibur config files)?
Are you able to get the line number also by counting \n characters?

skyfrench:
The character limit of 15 wouldn't capture the decoder_path in very many cases I don't suppose, but perhaps most other things?

Current example print out looks like this (when typo of ";" not ":"):

11:22:28,038 FR.App         ERROR - Frame receiver run failed: Parsing JSON configuration failed at offset 24: Missing a colon after a name of object member.     "decoder_type"; "Excalibur",

I could add line number. 

What would the most useful content of the message be?

ajgdls:
Most of what you have looks fine.  I think a line number would be good for finding quickly the issue within the file.
You could update the message to be
Parse JSON failed at line [n]: Missing a colon after a name of object member. "decoder_type"; "Excalibur",

ajgdls:
Great!  It looks like you might need to re-base this branch to be able to get it merged due to conflicts in the CMakeLists.txt.

PR #197, opened 2019-05-23
Changed UID Offset Plugin to generic Parameter Offset Plugin

MattTaylorDLS:


GDYendell:
Git has not tracked the rename of the cpp file. If you care about the history of that you could add an additional commit where you only rename the files.

MattTaylorDLS:
Weird that it managed to track the rename of the other two. I'm not so fussed about the history or it so happy for its history to be lost

frameProcessor/src/ParameterAdjustmentPlugin.cpp
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  // Apply any parameter adjustments
+  if (current_parameter_adjustments_.size() != 0)
+  {
+    std::map<std::string, int64_t>::iterator iter;
+    for (iter = current_parameter_adjustments_.begin(); iter != current_parameter_adjustments_.end(); ++iter) {
+      uint64_t param_value = frame->get_frame_number() + iter->second;
+      frame->meta_data().set_parameter<uint64_t>(iter->first, param_value);
+    }
```
skyfrench: I was expecting this to adjust the parameter value by adding the parameter adjustment to the current value of the parameter? This sets the new parameter value to the frame number + the adjustment?

frameProcessor/src/ParameterAdjustmentPlugin.cpp
```diff
@@ -0,0 +1,150 @@
...
+ * Perform processing on the frame. For the ParameterAdjustmentPlugin class we
+ * adjust confiugred parameters by the configured amount, adding the parameter if needed
+ *
+ * \param[in] frame - Pointer to a Frame object.
+ */
+void ParameterAdjustmentPlugin::process_frame(boost::shared_ptr<Frame> frame)
+{
+  if (frame->get_frame_number() == first_frame_number_)
+  {
+    // If at first frame, set the current adjustments to the the configured ones to apply from now on
```
timcnicholls: Why does this need to latch the configured adjustment in when the first frame number is seen? What happens if that isn't know at first, or the first frame is dropped for some reason?

frameProcessor/src/ParameterAdjustmentPlugin.cpp
```diff
@@ -0,0 +1,150 @@
...
+    LOG4CXX_DEBUG(logger_, "Setting current parameter adjustments to configured values");
+    current_parameter_adjustments_ = configured_parameter_adjustments_;
+  }
+
+  // Apply any parameter adjustments
+  if (current_parameter_adjustments_.size() != 0)
+  {
+    std::map<std::string, int64_t>::iterator iter;
+    for (iter = current_parameter_adjustments_.begin(); iter != current_parameter_adjustments_.end(); ++iter) {
+      uint64_t param_value = frame->get_frame_number() + iter->second;
```
timcnicholls: Can parameters only be adjusted relative the the current frame number, rather than by an absolute amount?

frameProcessor/include/ParameterAdjustmentPlugin.h
```diff
@@ -49,14 +48,14 @@ class UIDAdjustmentPlugin : public FrameProcessorPlugin
 
   /** Pointer to logger */
   LoggerPtr logger_;
-  /** UID adjustment currently in use **/
-  uint64_t current_uid_adjustment_;
-  /** UID adjustment configured to be used from next first frame onwards **/
-  uint64_t configured_uid_adjustment_;
+  /** Parameter adjustments currently in use **/
+  std::map<std::string, int64_t> current_parameter_adjustments_;
+  /** Parameter adjustments configured to be used from next first frame onwards **/
```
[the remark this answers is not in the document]
MattTaylorDLS: It is a tricky concept to explain in a limited number of words!

frameProcessor/src/ParameterAdjustmentPlugin.cpp
MattTaylorDLS: Yes I think this is what we decided in the meeting that we had. This plugin will just add the configured number to the frame_number to get a new number which it sets to the parameter.

frameProcessor/src/ParameterAdjustmentPlugin.cpp
MattTaylorDLS: Because we need some way of configuring it, but not have it affect any frames that are still passing through the chain. As there's no way to tell what 'acquisition' you're on from within a plugin, we have to say only apply it from the next frame onwards.
This is the way it used to work for the old UID plugin, and works for the Offset plugin.
]If the first frame is dropped, at the time, we said that's just unfortunate.
However, Gary and I have discussed this, and have thought that it might not be as solid as we'd like, and so discussed the possibililty of adding another parameter, or member to the frame to designate a scan id (which IS a separate thing from the acquisition ID !). We can maybe discuss this next week in the meeting.

frameProcessor/src/ParameterAdjustmentPlugin.cpp
MattTaylorDLS: Yes I think this is what we discussed in the meeting, that this plugin would just set a number based on the frame_number. Though I'm open to suggestions to other ways it could work

frameProcessor/include/ParameterAdjustmentPlugin.h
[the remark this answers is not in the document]
ajgdls: Is the first frame a restart of the detector?  Could the sentence become
"Parameter adjustments configured to be used from next detector restart" ?

frameProcessor/src/ParameterAdjustmentPlugin.cpp
ajgdls: NO NO NO NO NO NO NO NO

MattTaylorDLS: Give in to the dark side of multiple scan identifiers!

frameProcessor/include/ParameterAdjustmentPlugin.h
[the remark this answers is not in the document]
MattTaylorDLS: yeah that sounds better

frameProcessor/src/ParameterAdjustmentPlugin.cpp
GDYendell: Could we have a more structured acquisition ID instead? So if we get `scan123` and then `scan123.1`, the UID would see the `.1` and update its value, but the writer would see the `.1` to mean it isn't a new acquisition it is just a second batch and keep writing to the same file.

I am not sure if this is better or worse...

MattTaylorDLS:
I've removed the concept of having a first frame that the configured offset will apply from, so that now the offset applies as soon as it is configured.
I've also made it so that you can specify an input parameter, such that you can add the offset to the input parameter, and output it either on top, or to another named parameter. If you don't specify the input parameter, it will add it to the frame_id.
One question is: Currently, this assumes all input and output parameters are int_64t. Do we want to handle the case where the input parameter is not an int64_t? We'll just then add the configured value to the parameter casted as an int64_t and then write out to an int64_t parameter

PR #198, opened 2019-05-23
Updated default FrameMetaData constructor arguments (data type & compression)

skyfrench:
The default constructor for FrameMetaData currently sets the following:

        data_type_(raw_8bit),
        compression_type_(no_compression),

This should be changed to:

        data_type_(raw_unknown),
        compression_type_(unknown_compression),

so that if this constructor is used, nothing is assumed about the data or compression prior to the explicit set calls.


ajgdls:
Given that we have 80% approval from 2 weeks shouldn't we merge this?

skyfrench:
I was holding out for 100% :)

timcnicholls:
Pinging @ANeaves to do the honourable thing :wink:

Issue #36, opened 2017-06-22
Make acquisition ID a first class buffer parameter


timcnicholls:
A number of detectors implementing plugins for odin-data make use of an 'acquisition ID' (approx equivalent to a run number) to correctly handle  file writing operations during pipelined acquisition sequences. This ID may or may not originate from the detector itself. 

At present, the is no core header information passed in a shared memory buffer between FR and FP; each detector defines a FrameHeader structure in its common definitions. odin-data should define a common header struct, which would be inserted into the start of each frame header. This would include the acquisition ID as a field; on a system where it is not needed/used, the field can be left at a default value (-1?).

timcnicholls:
After discussion on the meaning and data type for acquisition_id, decision was taken to NOT make this a first class parameter and establish a core header. Instead, the decoder plugins for each detector that requires the concept should implement the appropriate type in their buffer headers, to be extracted, parsed and formatted as appropriate into the string acquisition_id field in the FP Frame object.

PR #199, opened 2019-06-05
integration testing

skyfrench:
I've made a stab at some integration testing for discussion.

Added 'integrationTest' to the test directory.

Created two new executables: 'odinData' and 'frameTests'

bin/odinData:

- configured with an ini file (specifying receiver, processor and simulator options)
- launches a frameReceiver and a frameProcessor
- runs a frameSimulator
- writes frames to hdf5 file

bin/frameTests

- boost test framework
- configured with an ini file (specifying expected contents of hdf5 file)
- compares hdf5 properties with expected properties

These two run after the frameReceiver and frameProcessor tests.

Notes:
- It has currently just been tested against excalibur (see [excalibur-detector #36](https://github.com/dls-controls/excalibur-detector/pull/36)).
- The configuration files for the specific detector are in the detector repository (e.g. excalibur-detector/data/test/integrationTest/config); but the odin-data repository has had to be 'polluted' with some knowledge of excalibur (.travis.yml, .travis-install-excalibur.sh and additional dependencies, travis now installs excalibur).
- Would imagine similar files are added to each detector we want to include in the standard integration testing; perhaps also a dummy detector?

Hopefully this isn't a complete swing and a miss and is a starting point for what we want ...


skyfrench:
> * I was a bit confused by the naming and structure of the test apps. I think `odinData` is misleading as it makes it sound like the core application for the framework, which it isn't (yet 😉). I also didn't quite understand why it's factored into two executables, one driving by the unit test framework. I'd float the idea that the whole integration test (`odinDataTest`??) - run the apps and check the output file - would be better served by a single executable.

I've renamed the test app 'odinDataTest' as suggested. This now launches the frame test app at the end. I factored it into two executables as it seemed to me that for code development (rather than the CI testing on travis) it might be useful to be able to run one and not the other. 'odinDataTest' was written as a way to quickly run up the three applications (receiver, processor & simulator) with a set of configurations. Then 'frameTests' was written as an application to quickly test that the contents of a HDF5 file match expectation.

> * I think we should avoid polluting this repo with any external detector dependencies. Otherwise we'll end up with an increasingly long list of dependencies, not all under the control of the core odin-data team, all of which have the potential to cause the odin-data CI to fail independent of any core changes. Those detector repos should test their own specific functionality and plugins by running this integration test app from their own CI jobs ,with their own configuration, against a specific release of odin-data. We should then, as part of odin-data, implement sufficiently capable dummy plugins (e.g. for receiving generic streams of frames from UDP or 0MQ) that allow a generic CI test to properly test odin-data. I'd be happy to contribute a UDP decoder and processing plugin pair at the very least.

The latest commits remove the excalibur test (and associated files). The CI test in excalibur-detector now runs these applications instead. Once we've got suitable dummy plugins and a matching simulator we can run that as part of the odin-data CI test as suggested.

> * We have a ticket open (my task) to migrate away from INI-style config files to JSON, so perhaps you could steal a march on that and lead the way into a brighter future?

I'll make a start ... 

> * I was caught out by the env variable substitution in the config files. If I run without INSTALL_PREFIX or EXCALIBUR_PREFIX set, I get a strange error output:
> 
> ```
> 0 [0x7f9ecd439d00] DEBUG Test.App null - Reading config file test_config/excalibur.ini
> 1 [0x7f9ecd439d00] ERROR Test.App null - Caught unhandled exception in FrameTestApp, application will terminate: basic_string::_S_construct null not valid
> terminate called after throwing an instance of 'std::logic_error'
>   what():  basic_string::_S_construct null not valid
> Aborted (core dumped)
> ```
This error is now caught!

> * Elsewhere we use string substitution in the CMake install step to write paths explicitly into installed config files, which we could also consider using here. Although having env variable substitution is quite nice too...

I've left the environment variable substitution in currently - happy to go with whatever is preferred ...

PR #202, opened 2019-06-12
Added performance stats to the FP plugins for monitoring and debug.

ajgdls:
This is all handled in the plugin base class and so no changes required on specific plugins.
Each plugin reports the following additional status eg

`{
  'timing': {
    'mean_process': 5.0,
    'last_process': 5,
    'max_process': 9
  }
}`

I'm open to name changes for the status items.  Calling reset on the FP will now also reset these statistics.

### DIFF
--- a/frameProcessor/include/FrameProcessorPlugin.h
+++ b/frameProcessor/include/FrameProcessorPlugin.h
@@ -40,6 +40,8 @@ public:
   virtual void configure(OdinData::IpcMessage& config, OdinData::IpcMessage& reply);
   virtual void requestConfiguration(OdinData::IpcMessage& reply);
   virtual void status(OdinData::IpcMessage& status);
+  void add_performance_stats(OdinData::IpcMessage& status);
+  void reset_performance_stats();
   void version(OdinData::IpcMessage& status);
   void register_callback(const std::string& name, boost::shared_ptr<IFrameCallback> cb, bool blocking=false);
   void remove_callback(const std::string& name);
@@ -53,6 +55,8 @@ private:
   LoggerPtr logger_;
 
   void callback(boost::shared_ptr<Frame> frame);
+
+  unsigned int elapsed_ms(struct timespec& start, struct timespec& end);
 
   /**
    * This is called by the callback method when any new frames have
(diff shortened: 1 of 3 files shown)


ajgdls:
Times are in ms.

ajgdls:
You are correct, this will add to process time.  Turning it off would require the addition of an if statement which would also add CPU cycles.
I can add some config to turn it on and odd but is there a test we can do with your specific 9000fps setup?  Can we run some acquisitions with the code absent, and then present to see if it makes a difference?  I don't want to add the config for the sake of it if the processing time is for example dwarfed by file writing or some other part of processing.

Final comment is that each plugin runs in its own thread, so as long as you have enough threads in your OS the number of plugins would not necessarily have a linear additive effect on the overall process time.  At least I think that statement is true.

GDYendell:
You could add a time measurement around the time measurement to see how long the time measurement takes.

ajgdls:
Yes I could do that.  @MattTaylorDLS would that be OK?  I can run a long test and it should come up with some reasonable estimate of the time taken to work out the time (ignoring the fact that taking the additional time will in fact take some non zero amount of time...)

GDYendell:
goto https://github.com/odin-detector/odin-data/pull/202#issuecomment-501224958

MattTaylorDLS:
I'm lost! But yes, go for it. I could also maybe test with the 9k eiger this aftenroon

timcnicholls:
You can use the compiler directive `_builtin_expect` to optimise branch prediction around the condition associated with turning the timing off, although it can be hard to predict if that helps. RapidJSON and LOG4XX have macros for that already you can find in our code base.

MattTaylorDLS:
Well, the detector only actually outputs frames at 3000Hz, but at that speed it seems to not make any difference over a 250000 frame acquisition.
Getting results like:
timing: {
  last_process: 0,
  max_process: 7,
  mean_process: 4.877923e-260
}

and

timing: {
  last_process: 0,
  max_process: 10,
  mean_process: 0.0
}

ajgdls:
Mm, I have used millisecond precision here but if all we are ever going to get is ~0.0 then perhaps I should actually use microseconds.  I think the calculation will essentially be unchanged, I just multiply by a different factor.  I think I will just try this out on my version.

ajgdls:
OK this is much better, now I'm reporting in microseconds without adding any extra complexity.  Now all of the plugins which were reporting 0 ms and 0.0 ms averages are reporting meaningful results.  For example for Excalibur 1M the blosc plugin reports a mean of 250us and max of 936us instead of all 0 ms.

ajgdls:
@MattTaylorDLS I'm going to merge this as is for now.  If in the future we think it is causing performance issues I can add the ability to turn it on/off.

MattTaylorDLS:
Yeah go for it

PR #203, opened 2019-06-12
Return strings for datatype and compression in request config

MattTaylorDLS:
The config for these was changed from int to string, but they are still being reported as ints in the request_configuration message.

### DIFF
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -354,10 +354,10 @@ void FileWriterPlugin::requestConfiguration(OdinData::IpcMessage& reply)
   std::map<std::string, DatasetDefinition>::iterator iter;
   for (iter = this->dataset_defs_.begin(); iter != this->dataset_defs_.end(); ++iter) {
     // Add the dataset type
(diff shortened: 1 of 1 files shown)


MattTaylorDLS:
> Should the blosc compressor and blosc shuffle also have similar string enum types?

Not sure, depends if we're happy that they can be enums as they're not top level, or if we want it consistent across all input parameters.

ajgdls:
I guess I vote that all enums should be consistent, we still need to set those from the control API.

MattTaylorDLS:
Would we want to do that on this PR, or create a new one?

MattTaylorDLS:
Looking at the code, the slight difference between the datatype/compression and the blosc settings, is that the datatype and compression are enums we've made up, whereas the blosc settings are values i don't think we've made up, which are set straight to the cd_values as ints for when blsoc_compress() is called

ajgdls:
Yes I see the difference.  It's just a point to note that these would be easier to understand as String enum API calls.

However this could be a different PR, it is not necessary for this one.  Although I made the comment I also approved so go ahead and merge, fine with me.

MattTaylorDLS:
I've created an issue for it (#204)

PR #210, opened 2019-07-09
Stop live view resetting endpoint if reconfigured without an endpoint

### LINKED ISSUE #209: LiveViewPlugin resets endpoint to default when any other parameter configured
I believe this logic was added to delay the endpoint initialisation and avoid processes trying to bind to the same endpoint as soon as they startup. The simplest fix would be to only set the endpoint to default if it is not already set when the first message comes in.

https://github.com/odin-detector/odin-data/blob/ec4b87c90924aabc814cbeb6d940984a54ad5eaa/frameProcessor/src/LiveViewPlugin.cpp#L161-L168

The Kafka plugin has a similar problem where really we would like to pass arguments to the constructor, but are instead passing them to configure and just restricting them from being set more than once. Perhaps we need an API that allows for a one time initial configuration.

ANeaves:
Fixes #209

### DIFF
--- a/frameProcessor/src/LiveViewPlugin.cpp
+++ b/frameProcessor/src/LiveViewPlugin.cpp
@@ -162,7 +162,7 @@ void LiveViewPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMessag
     {
       set_socket_addr_config(config.get_param<std::string>(CONFIG_SOCKET_ADDR));
     }
-    else if (is_bound_ == false) //do not default if the socket is already bound
+    else if (not is_bound_) //do not default if the socket is already bound
     {
       set_socket_addr_config(DEFAULT_IMAGE_VIEW_SOCKET_ADDR);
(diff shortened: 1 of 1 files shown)


frameProcessor/src/LiveViewPlugin.cpp
```diff
@@ -162,7 +162,7 @@ void LiveViewPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMessag
     {
       set_socket_addr_config(config.get_param<std::string>(CONFIG_SOCKET_ADDR));
     }
-    else
+    else if (is_bound_ == false) //do not default if the socket is already bound
```
GDYendell: Maybe `if (not is_bound)`? Although it sounds awkward.

ANeaves: Didn't know those sort of keywords exist in c++, neat

PR #211, opened 2019-08-01
Prevent unexpected frames being shown as error when not writing

MattTaylorDLS:


frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -693,9 +693,17 @@ bool FileWriterPlugin::frame_in_acquisition(boost::shared_ptr<Frame> frame) {
...
       std::stringstream ss;
       ss << "Unexpected acquisition ID on frame (" << frame_acquisition_ID << ")";
-      set_error(ss.str());
-      ss << " for frame " << frame->get_frame_number();
-      LOG4CXX_WARN(logger_, ss.str());
+      if (writing_)
+      {
+        set_error(ss.str());
+        ss << " for frame " << frame->get_frame_number();
+        LOG4CXX_WARN(logger_, ss.str());
```
GDYendell: `set_error` will log an error message, so you probably don't want the WARN.

I also wonder if we should just not put the frame number in the error message and always print debug with the frame number. If you run a scan at 500Hz it's going to print a lot of messages before you realise and they won't be filtered because they will all be unique. I think one message without the frame number would be sufficient to figure out the problem.

MattTaylorDLS: The problem with relying on set_error to log, is that if there is already an error there (e.g. from a previous aquisition that hasn't been cleared, then it won't even log the error, and so there won't be any indication that something went wrong in this acquisition. Having the frame number in the log message has also been useful in the past for being able to easily determine which frames were missed.

ajgdls: I think that if the error has not been cleared then there should be red indicators on the screen?  Hopefully(?) the higher level code is checking there are no errors before attempting to start.

MattTaylorDLS: yes there are the errors on the screen. at the moment the higher level code isn't checking that. Do we want to be reliant on higher level code behaving? And if it doesn't, then we don't give any indication as to why we haven't written frames we've been given?

ajgdls: I agree don't rely on anything...

GDYendell: We should probably check that the current error is equal and if not the set it, rather than just have we ever had the error. Then we know that most recent error is what is displayed on the screen. I think it is important to not log the same error over and over again.

MattTaylorDLS: It could be the case that the current error is the same though, but is from a previous acquisition, but we still want to log that this acquisition has also failed for the same reason.

GDYendell: Yes, but that error will still be displayed on the screen, so we will know that is still the reason it is not working.

MattTaylorDLS: But when Data Analysis say "What happened on acquisition xyz_123, we can't see any frames?" 2 weeks later, as they do quite a lot at the moment, there won't be any indication in the logs as to anything happening. We won't be able to tell if the frames even made it to the processor.

GDYendell: Ah I see... But, would that will be in the debug logs on graylog?

MattTaylorDLS: not if i take out the LOG_WARN that you're suggesting!

GDYendell: But the DEBUG would still be there.

MattTaylorDLS: It only does the DEBUG if it's not currently writing (and DEBUG isn't sent to graylog anyway). In the meeting we agreed that we wouldn't really care if odin wasn't set up to write, so would just log to DEBUG. If we were writing, then we should log to WARN.

GDYendell: OK I'm on board. We don't log DEBUG to graylog, so we can't send different level messages to graylog and console.

PR #212, opened 2019-08-05
Kafka config

EmmaCD:
Updated the kafka producer configuration:

* Allow frames to build up in the odin buffer when the internal kafka producer queue is full rather than dropping the frame.

* Allow selection of kafka parameters with an effect on throughput to be configurable. Default parameters achieve the highest throughput observed in testing. Parameters control:
    * The kafka producer queue : default max 200 messages or 2.1 GB, whichever reached first
    * Message batching : no message batching by default
    * Message acknowledgement : full acknowledgement by default

* Allocate memory the size of the compressed image rather than full data size

### DIFF
--- a/frameProcessor/src/KafkaProducerPlugin.cpp
+++ b/frameProcessor/src/KafkaProducerPlugin.cpp
@@ -194,6 +194,19 @@ namespace FrameProcessor {
     kafka_config = rd_kafka_conf_new();
     int status;
 
+
+    status = rd_kafka_conf_set(kafka_config,
+                               "message.send.max.retries",
+                               KAFKA_MESSAGE_MAX_RETRIES,
+                               errBuf,
+                               sizeof(errBuf));
+
+    if (status != RD_KAFKA_CONF_OK) {
+      LOG4CXX_ERROR(logger_, "Kafka configuration error while setting max retries: "
+        << errBuf);
+      return;
+    }
+
     status = rd_kafka_conf_set(kafka_config,
                                "message.max.bytes",
                                KAFKA_MESSAGE_MAX_BYTES,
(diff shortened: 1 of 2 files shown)


frameProcessor/src/KafkaProducerPlugin.cpp
```diff
@@ -149,6 +140,17 @@ namespace FrameProcessor {
...
+    if (config.has_param(CONFIG_SERVERS)) {
+      destroy_kafka();
+      configure_kafka_servers(config.get_param<std::string>(CONFIG_SERVERS));
+      need_configure_topic = true;
+    }
+
+    if (config.has_param(CONFIG_TOPIC) || need_configure_topic) {
+      configure_kafka_topic(config.get_param<std::string>(CONFIG_TOPIC));
+    }
+
```
GDYendell: Why does this need to come after the parameters? Do we need to force a reconfigure of the server if certain other things change?

EmmaCD: The producer needs to be created before the topic can be configured, so configure_kafka_servers has to come first to set the parameters and create the producer, then the topic can be configured. 

Any time kafka configuration parameters change you need to first create producer then configure the topic, but there should be no reason to change parameters after start up. 

@hir12111 might want to expand on that.

hir12111: We are assuming that everything is configured at start, the parameters are used when creating the kafka handler, that's why we need to first set the parameters (as attributes), and then configure the server so that those parameters are used.

GDYendell: If we are able to change them at runtime, they should take effect without having to configure the server explicitly. My initial thought is that it should raise an error if you try to configure them after startup, but Emilio suggests that it could be useful during commissioning to be able to tune these parameters at runtime, so maybe we should allow it. I think there should be a need_configure_server flag that is set by when the relevant parameters are changed, as with the topic.

ajgdls:
This can be re-opened when ready for approval.

PR #222, opened 2019-11-15
Update to use static dimensions for HDF5 datasets when frame count given

GDYendell:
This is to solve a problem with VDS datasets mapping to OdinData HDF5 files. If the underlying files are H5S_UNLIMITED in extent, the VDS mappings exhibit strange behaviour, or can't even be opened, via SWMR. The file has to be closed before the data can be processed. With this change, when datasets are created with a non-zero frame count, the exact size will be used for the dataset size. If frame count is set to 0, it will use the old behavior and the file will have to be manually closed when the client knows the acquisition has ended.

### DIFF
--- a/frameProcessor/src/HDF5File.cpp
+++ b/frameProcessor/src/HDF5File.cpp
@@ -30,7 +30,8 @@ HDF5File::HDF5File() :
         param_memspace_(-1),
         hdf5_error_flag_(false),
         file_index_(0),
-        use_earliest_version_(false)
+        use_earliest_version_(false),
+        unlimited_(false)
 {
   static bool hdf_initialised = false;
   this->logger_ = Logger::getLogger("FP.HDF5File");
@@ -46,6 +47,16 @@ HDF5File::HDF5File() :
 HDF5File::~HDF5File() {
   // Call to close file in case it hasn't been closed
   close_file();
+}
+
+void HDF5File::set_unlimited() {
+  if (hdf5_datasets_.empty()) {
+    LOG4CXX_DEBUG_LEVEL(1, logger_, "Setting HDF5 datasets to use unlimited");
+    unlimited_ = true;
+  }
+  else {
+    throw std::runtime_error("Datasets have already been created. Cannot set unlimited.");
+  }
 }
 
 /**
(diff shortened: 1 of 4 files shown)


ajgdls:
You know that now we have to appoint a new "code police" representative as @MattTaylorDLS has gone...

MattTaylorDLS:
Don't worry I'm still keeping an eye on it!

frameProcessor/src/HDF5File.cpp
```diff
@@ -197,24 +208,23 @@ void HDF5File::close_file() {
...
-  this->extend_dataset(dset, (frame_offset + 1) * outer_chunk_dimension);
+  if (unlimited_) {
+    this->extend_dataset(dset, (frame_offset + 1) * outer_chunk_dimension);
+  }
 
-  LOG4CXX_TRACE(logger_, "Writing frame offset=" << frame_no  <<
-      " (" << frame_offset << ")" <<
-      " dset=" <<frame.get_meta_data().get_dataset_name());
+  LOG4CXX_TRACE(logger_, "Writing frame offset=" << frame.get_frame_number()  << " (" << frame_offset <<
+                         ") dset=" <<frame.get_meta_data().get_dataset_name());
```
MattTaylorDLS: Need a space between << and frame.get_mete_data() !

GDYendell:
We tested this on I18 and were able to perform live analysis from the VDS files. The images, sum and UID were all visible and correct during the acqusition.

GDYendell:
I have already also verified that creating static datasets does not take any longer than unlimited (i.e. the dataset is always created by the time I release the mouse button, regardless of the number of frames). Because we are using the following:

- Fill Value = `H5D_FILL_TIME_IFSET` with a user-defined value
- Allocation Time = `H5D_ALLOC_TIME_INCR`
- Chunked Datasets

the fill value is written when each chunk is written, according to this page: https://support.hdfgroup.org/HDF5/doc_resource/H5Fill_Behavior.html

I am slightly confused by this. If we write the chunk with direct chunk write, it has to be the full chunk and therefore there will be no gaps and no fill value used. For the chunks we don't write at all presumably no fill value is set. However, if I Start/Stop an acquisition, the file that is created is all zeroes.

ajgdls:
@GDYendell isn't it the case that the fill values are applied by the reading application?  What is the file size in the case where you create a file, and then close it without writing any data?  Does the size match a file full of zeros, or a file with nothing but header information?

GDYendell:
@ajgdls I also assumed the fill value was dynamically generated by the reader for chunks that haven't been written to and I have not experienced anything to the contrary. That page I linked seems to suggest otherwise, though:

> User Fill Value Set In Each Chunk Accessed During Write

The size does not suggest that the fill value has actually been written to the chunks. It is just the single fill value in the header.

GDYendell:
Rebased to squash commits. I think this is ready to merge now.

PR #224, opened 2020-01-20
impose 64-byte alignment on datablocks

wnichols1:
AVX likes alignment, so this is an easy change which could be helpful in future.

### DIFF
--- a/frameProcessor/src/DataBlock.cpp
+++ b/frameProcessor/src/DataBlock.cpp
@@ -28,7 +28,7 @@ DataBlock::DataBlock(size_t block_size) :
   // Create this DataBlock's unique index
   index_ = DataBlock::index_counter_++;
   // Allocate the memory required for this data block
-  block_ptr_ = memalign(alignment, block_size);
+  posix_memalign(&block_ptr_, alignment, block_size);
 }
 
 /**
@@ -77,7 +77,7 @@ void DataBlock::resize(size_t block_size)
     // Free the current allocation first
     free(block_ptr_);
     // Allocate the new number of bytes
-    block_ptr_ = memalign(alignment, block_size);
+    posix_memalign(&block_ptr_, alignment, block_size);
     // Record our new size
     allocated_bytes_ = block_size;
   }
--- a/frameProcessor/src/DataBlock.cpp
+++ b/frameProcessor/src/DataBlock.cpp
@@ -28,7 +28,11 @@ DataBlock::DataBlock(size_t block_size) :
   // Create this DataBlock's unique index
   index_ = DataBlock::index_counter_++;
   // Allocate the memory required for this data block
-  posix_memalign(&block_ptr_, alignment, block_size);
+  int rc = posix_memalign(&block_ptr_, alignment, block_size);
+  if(rc)
+  {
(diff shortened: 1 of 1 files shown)


frameProcessor/src/DataBlock.cpp
```diff
@@ -7,11 +7,13 @@
 
 #include <DataBlock.h>
 #include "DebugLevelLogger.h"
+#include <malloc.h>
 
 namespace FrameProcessor
 {
 int DataBlock::index_counter_ = 0;
 
+static const int alignment = 32;
```
timcnicholls: Wouldn't alignment be 64 in this case for AVX for have I misunderstood the `memalign()` arguments?

frameProcessor/src/DataBlock.cpp
```diff
@@ -26,7 +28,7 @@ DataBlock::DataBlock(size_t block_size) :
   // Create this DataBlock's unique index
   index_ = DataBlock::index_counter_++;
   // Allocate the memory required for this data block
-  block_ptr_ = malloc(block_size);
+  block_ptr_ = memalign(alignment, block_size);
```
timcnicholls: The man page for this states `memalign()` is obsolete and should be replaced by `posix_memalign()` or `aligned_alloc()`

frameProcessor/src/DataBlock.cpp
wnichols1: no you are right - that was a mistake in the commit.

wnichols1: I have altered the commit to 64.

frameProcessor/src/DataBlock.cpp
wnichols1: I think memalign is safe and viable in spite of what the manual says. posix_memalign is more complicated, and could go wrong as it takes a pointer to a pointer... but I will change it if you prefer it. I don't mind.

timcnicholls: Why would it be more likely to go wrong? It's passing a pointer to a pointer, this case the address of the `block_ptr_` member of `DataBlock`, which will be valid if the constructor succeeded.

If you don't wish to use `posix_memalign()`, `aligned_alloc()` is available and doesn't require `void **ptr`. It imposes the need to ensure that `block_size` is a multiple of `alignment` but you probably want that anyway for AVX and efficient storage.

I don't feel a strong desire to avoid `memalign()` but it seems prudent to me given it is labelled as obsolete.

Not a comment on your PR, but I also noticed that the original `DataBlock` implementation has no check that the memory allocation succeeds, i.e. `block_ptr_` isn't checked to be non-null. Is this done elsewhere @ajgdls ?

wnichols1: ok then.

ajgdls: @timcnicholls you are right the original DataBlock implementation was not handling memory allocation failures.
The posix_memalign call returns one of two errors if it fails, and this sort of failure is catastrophic to the application (it just won't work) and so probably now is a good time to fix that.
@wnichols1 should we throw an exception on a failure?  We should print an error log message at the very least.  I know it isn't your code that has introduced this issue but as you are changing the call it would be a good time to improve the error handling for this kind of failure.

PR #227, opened 2020-02-19
Filepath update

ajgdls:
I've added a mechanism to this PR for setting a nack with rejection message to an IpcMessage object.
I've currently got the rejection message set as a parameter in the object.
I've then implemented checks for the file path to check if the path exists, check if the path is a valid path, and check if the application user has permission to write to the path.

### DIFF
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -532,7 +532,7 @@ void FileWriterPlugin::configure_file(OdinData::IpcMessage& config, OdinData::Ip
         if (eaccess(file_path.c_str(), W_OK)){
           // Return code other then zero is a failure
           std::stringstream ss;
-          ss << "Path is not writable by the user: " << file_path;
+          ss << "User does not have write permissions for directory: " << file_path;
           LOG4CXX_ERROR(logger_, ss.str());
           reply.set_nack(ss.str());
         } else {
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -531,23 +531,26 @@ void FileWriterPlugin::configure_file(OdinData::IpcMessage& config, OdinData::Ip
         // Check directory has write permission
         if (eaccess(file_path.c_str(), W_OK)){
           // Return code other then zero is a failure
-          LOG4CXX_ERROR(logger_, "HDF5 - Path is not writable by the user: " << file_path);
-          reply.set_nack("HDF5 - Path is not writable by the user: " + file_path);
-          reply.set_msg_type(OdinData::IpcMessage::MsgTypeNack);
+          std::stringstream ss;
+          ss << "Path is not writable by the user: " << file_path;
+          LOG4CXX_ERROR(logger_, ss.str());
+          reply.set_nack(ss.str());
         } else {
           // All checks passed, we can write to this location
           this->next_acquisition_->file_path_ = config.get_param<std::string>(FileWriterPlugin::CONFIG_FILE_PATH);
           LOG4CXX_DEBUG_LEVEL(1, logger_, "Next file path changed to " << this->next_acquisition_->file_path_);
         }
       } else {
-        LOG4CXX_ERROR(logger_, "HDF5 - Path is not a directory: " << file_path);
-        reply.set_nack("HDF5 - Path is not a directory: " + file_path);
-        reply.set_msg_type(OdinData::IpcMessage::MsgTypeNack);
+        std::stringstream ss;
+        ss << "Path is not a directory: " << file_path;
+        LOG4CXX_ERROR(logger_, ss.str());
+        reply.set_nack(ss.str());
       }
     } else {
-      LOG4CXX_ERROR(logger_, "HDF5 - Invalid path requested: " << file_path);
-      reply.set_nack("HDF5 - Invalid path requested: " + file_path);
-      reply.set_msg_type(OdinData::IpcMessage::MsgTypeNack);
+      std::stringstream ss;
+      ss << "Invalid path requested: " << file_path;
+      LOG4CXX_ERROR(logger_, ss.str());
+      reply.set_nack(ss.str());
     }
   }
   if (config.has_param(FileWriterPlugin::CONFIG_FILE_NAME)) {
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -521,8 +521,34 @@ void FileWriterPlugin::configure_file(OdinData::IpcMessage& config, OdinData::Ip
   LOG4CXX_DEBUG_LEVEL(1, logger_, "Configure file name and path");
   // Check for file path and file name
   if (config.has_param(FileWriterPlugin::CONFIG_FILE_PATH)) {
-    this->next_acquisition_->file_path_ = config.get_param<std::string>(FileWriterPlugin::CONFIG_FILE_PATH);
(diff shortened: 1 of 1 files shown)


common/src/IpcMessage.cpp
```diff
@@ -213,6 +213,20 @@ bool IpcMessage::has_param(const std::string& param_name) const
...
+//!
+//! This method can be used to set this IpcMessage up as a rejection (nack) message.
+//! A reason is supplied and set as a parameter called 'rejetion'
+//!
+//! \return void
+
+void IpcMessage::set_nack(const std::string& reason)
+{
+  this->set_msg_type(MsgTypeNack);
+  this->set_param("rejection", reason);
```
GDYendell: Is "rejection" a special string that needs to be recognised by clients / the control server?

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -506,8 +506,34 @@ void FileWriterPlugin::configure_file(OdinData::IpcMessage& config, OdinData::Ip
...
+    // Check path exists
+    boost::system::error_code ec;
+    if (boost::filesystem::exists(p, ec)){
+      // Check path is a directory
+      if (boost::filesystem::is_directory(p, ec)){
+        // Check directory has write permission
+        if (eaccess(file_path.c_str(), W_OK)){
+          // Return code other then zero is a failure
+          LOG4CXX_ERROR(logger_, "HDF5 - Path is not writable by the user: " << file_path);
+          reply.set_nack("HDF5 - Path is not writable by the user: " + file_path);
```
GDYendell: I think it would be better to only contruct these strings once.

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -506,8 +506,34 @@ void FileWriterPlugin::configure_file(OdinData::IpcMessage& config, OdinData::Ip
...
+          reply.set_msg_type(OdinData::IpcMessage::MsgTypeNack);
+        } else {
+          // All checks passed, we can write to this location
+          this->next_acquisition_->file_path_ = config.get_param<std::string>(FileWriterPlugin::CONFIG_FILE_PATH);
+          LOG4CXX_DEBUG_LEVEL(1, logger_, "Next file path changed to " << this->next_acquisition_->file_path_);
+        }
+      } else {
+        LOG4CXX_ERROR(logger_, "HDF5 - Path is not a directory: " << file_path);
+        reply.set_nack("HDF5 - Path is not a directory: " + file_path);
```
GDYendell: You have used the same prefix for all these messages, but it is not used throughout the code. Is it necessary to have 'HDF5 -' at the front of every message? I think they give enough information to know what is wrong, and in the logs we will have the file name.

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -506,8 +506,34 @@ void FileWriterPlugin::configure_file(OdinData::IpcMessage& config, OdinData::Ip
...
+        reply.set_msg_type(OdinData::IpcMessage::MsgTypeNack);
+      }
+    } else {
+      LOG4CXX_ERROR(logger_, "HDF5 - Invalid path requested: " << file_path);
```
GDYendell: Should these messages be propagated to the error string for the client?

timcnicholls: Are there some places like this in the code where you've got an explicit call set `set_msg_type()` for a NACK without calling `set_nack()`

common/src/IpcMessage.cpp
timcnicholls: I didn't get chance to respond to your question on Slack last night @ajgdls but this touches on what I would have said. I think adding a string message on NACKs is fine - there is already an _ad-hoc_ implementation of this used elsewhere, e.g. the [FR RX thread ](https://github.com/odin-detector/odin-data/blob/master/frameReceiver/src/FrameReceiverRxThread.cpp#L319) :
```
        rx_reply.set_msg_type(IpcMessage::MsgTypeNack);
        rx_reply.set_msg_val(msg_val);
        rx_reply.set_param("error", std::string("Unexpected message type from main thread"));
        rx_channel_.send(rx_reply.encode());
```
My (slight) preference is for that error field to be "error" rather than "rejection" but otherwise it's the same mechanism. I'd be happy to use your `set_nack()` method wherever this has been done in the FR.

frameProcessor/src/FileWriterPlugin.cpp
ajgdls: So far there are no nacks anywhere in any of the plugins as far as I can tell.  Currently I think most messages were coming back as invalid.  I think we need a new task to go through and rectify this everywhere.
I'm happy to do that for the HDF5 plugin as part of this PR but I wanted first some confirmation that people are happy with the idea of nacking a configuration with a specific rejection message (ie I want it to be consistent everywhere).

So the main questions are:
1) Do we keep the new method in IpcMessage for nacking a message?
2) Is a parameter called "rejection" suitable, or should the IpcMessage have a proper member variable that is part of its core?
3) Do we stop people from being able to set nack with set_msg_type() and force them to use the method with a message?

frameProcessor/src/FileWriterPlugin.cpp
ajgdls: Agreed

frameProcessor/src/FileWriterPlugin.cpp
ajgdls: Agreed

frameProcessor/src/FileWriterPlugin.cpp
timcnicholls: I've partly responded in comments above, but to answer directly:
1. I'm happy for there to be a specific method `set_nack(std::string&)` in IpcMessage for NACking, as you've proposed. 
2. I prefer the message string parameter to be `error` or `reason` rather than `rejection`, mostly stylistically, but essentially to be a noun rather than a verb. I'm not sure it's necessary to implement a required member variable since the JSON message is more freely formatted.
3. Fairly hard to prevent `set_msg_type()` being used without breaking current code or raising an exception. You could, in that method, if the type is NACK, construct an empty or 'unknown' response in the absence of an explicit value?

timcnicholls: I also wonder, given this PR is doing two things, if we should separate out the `set_nack()` feature from the file path checking into two PRs? Otherwise we're hiding a new feature of IpcMessage behind something FP specific?

ajgdls: OK great, so I will set the string to be called "error" as you already have in the FR.  I will also thoroughly go through the HDF5 FP plugin and ensure it correctly responds to all requests.
@GDYendell you OK with those actions?

ajgdls:
These are the responses now passed back:

Invalid path:

`{u'msg_val': u'configure', u'params': {u'error': u'Invalid path requested: /tmp2'}, u'id': 0, u'msg_type': u'nack', u'timestamp': u'2020-04-29T14:42:05.002894'}`

Wrong permissions:

`{u'msg_val': u'configure', u'params': {u'error': u'Path is not writable by the user: /dls_sw/prod'}, u'id': 0, u'msg_type': u'nack', u'timestamp': u'2020-04-29T14:42:37.023998'}`

Path is not a directory:

`{u'msg_val': u'configure', u'params': {u'error': u'Path is not a directory: /home/gnx91527/test'}, u'id': 0, u'msg_type': u'nack', u'timestamp': u'2020-04-29T14:45:11.639923'}`

GDYendell:
Could I suggest you change the wrong permissions message to "User does not have permissions for directory: %s"? It is a little ambiguous currently if the %s is the user or the path.

PR #229, opened 2020-03-06
Perform checks of dropped frames or low buffers

ajgdls:
FP adapter can use inter adapter comms to check the free buffers and if any frames dropped in the FR adapter.
FP adapter will then reject any attempt to start a new file write until free buffers is above 5% and there are no dropped frames.
Resetting FR should clear dropped frames, free buffers would be resolved by fixing whatever has gone wrong.

### DIFF
--- a/tools/python/odin_data/frame_processor_adapter.py
+++ b/tools/python/odin_data/frame_processor_adapter.py
@@ -16,7 +16,6 @@ from tornado.ioloop import IOLoop
 FP_ADAPTER_KEY = 'fr_adapter_name'
 FP_DEFAULT_ADAPTER_KEY = 'fr'
 PCT_BUFFER_FREE_KEY = 'buffer_threshold'
-DEFAULT_PCT_BUFFER_FREE = 5.0
 
 def bool_from_string(value):
     bool_value = False
@@ -49,12 +48,13 @@ class FrameProcessorAdapter(OdinDataAdapter):
 
         self._fr_adapter = None
 
-        self._fr_pct_buffer_threshold = DEFAULT_PCT_BUFFER_FREE
-        if PCT_BUFFER_FREE_KEY in kwargs:
-            try:
-                self._fr_pct_buffer_threshold = float(kwargs[PCT_BUFFER_FREE_KEY])
-            except:
-                logging.error("Could not set the buffer threshold to: {}".format(kwargs[PCT_BUFFER_FREE_KEY]))
+        # Set the bufer threshold check, default to 0.0 (no check) if the option
+        # does not exist or is badly formatted
+        self._fr_pct_buffer_threshold = 0.0
+        try:
+            self._fr_pct_buffer_threshold = float(self.options.get(PCT_BUFFER_FREE_KEY, 0.0))
+        except ValueError:
+            logging.error("Could not set the buffer threshold to: {}".format(kwargs[PCT_BUFFER_FREE_KEY]))
 
         self._param = {
             'config/hdf/acquisition_id': '',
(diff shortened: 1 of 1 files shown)


tools/python/odin_data/frame_processor_adapter.py
```diff
@@ -167,6 +191,40 @@ def put(self, path, request):  # pylint: disable=W0613
...
+                for fr in frs:
+                    try:
+                        frames_dropped = fr['frames']['dropped']
+                        empty_buffers = fr['buffers']['empty']
+                        mapped_buffers = fr['buffers']['mapped']
+                        if frames_dropped > 0:
+                            valid_check = False
+                            reason = "Frames dropped [{}] on at least one FR".format(frames_dropped)
+                        pct_free = float(empty_buffers) / (float(empty_buffers+mapped_buffers)) * 100.0
+                        if pct_free < 5.0:
```
timcnicholls: Can this threshold be exposed as an option? There may be a reason to choose a different value, based on the performance of a particular deployment of odin-data in a system.

ajgdls: Done

ajgdls:
In answer to your larger question, I agree I think it would be a good idea to have these kind of dependencies controlled by the higher "detector" level adapter.
However I know getting around to implement something like that is not going to be high on the list for a while.  So I don't mind if you want to reject this change, we need to have something in place at DLS now so I can merge this locally at DLS until such a time that we can pull it into the aforementioned detector adapter.

tools/python/odin_data/frame_processor_adapter.py
ajgdls: Yes

timcnicholls:
No, I have no reason to reject this outright. Please do merge it. If this ever makes it into a higher-level adapter we can remove it. 

You could also only do the test if the config file option is present (have it default to 0 or something to detect default ) so that other systems can choose not to check it?

tools/python/odin_data/frame_processor_adapter.py
```diff
@@ -39,6 +43,19 @@ def __init__(self, **kwargs):
...
+            self._fr_adapter_name = kwargs[FP_ADAPTER_KEY]
+
+        self._fr_adapter = None
+
+        self._fr_pct_buffer_threshold = DEFAULT_PCT_BUFFER_FREE
+        if PCT_BUFFER_FREE_KEY in kwargs:
+            try:
+                self._fr_pct_buffer_threshold = float(kwargs[PCT_BUFFER_FREE_KEY])
+            except:
+                logging.error("Could not set the buffer threshold to: {}".format(kwargs[PCT_BUFFER_FREE_KEY]))
```
GDYendell: What error are you catching here? Could it be more specific?

timcnicholls: At this point, once the superclass init function is called, the config parameters are available in `self.options`, which is the nominal way adapters get their config options. I'd suggest at this point, if the parameter is not present, the buffer threshold check is not performed.

ajgdls: OK @timcnicholls so you think we can do away with the defaults and instead use the option existence as a check vs no check.
@GDYendell I think the specific exception here could be if someone set that option to be a non float parsable value (like a word instead of a number).  I can probably specialize that exception check.

timcnicholls: Yes that's right. The `ApiAdapter` constructor parses `kwargs` into `self.options` as a dict. You can use the dict `get` method to retrieve the value or a default. For instance, I've implemented this in a recent change to the EXCALIBUR adapter (on 24bit-mode branch) to allow the low-level detector class to be used:
```
# Determine if the raw (low-level) or high-level detector class is to be used
use_raw_detector = bool(self.options.get('raw_detector', False))
```
You're right to trap `ValueError` if the option isn't parseable to a float.

ajgdls: Changes made

PR #230, opened 2020-03-12
File Writing Metrics

GDYendell:
This adds a timer around certain hdf5 calls to log the time taken and to print an error if it is not reset within a certain time (i.e. if the main thread is stuck in the hdf5 call).

There is also some not entirely related tidying up.

This does not yet publish the metrics over the meta channel. I think the meta needs some refactoring before doing this. I have made a start here https://github.com/dls-controls/odin-data/tree/meta-refactor on the C++ side, but stopped at the point that I also needed to change the python library.

### DIFF
--- a/frameProcessor/src/HDF5File.cpp
+++ b/frameProcessor/src/HDF5File.cpp
@@ -25,13 +25,15 @@ herr_t hdf5_error_cb(unsigned n, const H5E_error2_t* err_desc, void* client_data
   return 0;
 }
 
-HDF5File::HDF5File() :
+HDF5File::HDF5File(const HDF5ErrorDefinition_t& hdf5_error_definition) :
         hdf5_file_id_(-1),
         param_memspace_(-1),
         hdf5_error_flag_(false),
         file_index_(0),
         use_earliest_version_(false),
-        unlimited_(false)
+        unlimited_(false),
+        watchdog_timer_(hdf5_error_definition.callback),
+        hdf5_error_definition_(hdf5_error_definition)
 {
   static bool hdf_initialised = false;
   this->logger_ = Logger::getLogger("FP.HDF5File");
@@ -120,8 +122,10 @@ void HDF5File::clear_hdf_errors()
  * \param[in] use_earliest_version - Whether to use the earliest version of HDF5 library
  * \param[in] alignment_threshold - Chunk threshold
  * \param[in] alignment_value - Chunk alignment value
+ *
+ * \return - The duration of the H5Fcreate call
  */
-void HDF5File::create_file(std::string filename, size_t file_index, bool use_earliest_version, size_t alignment_threshold, size_t alignment_value)
+size_t HDF5File::create_file(std::string filename, size_t file_index, bool use_earliest_version, size_t alignment_threshold, size_t alignment_value)
 {
   // Protect this method
   boost::lock_guard<boost::recursive_mutex> lock(mutex_);
@@ -153,7 +157,11 @@ void HDF5File::create_file(std::string filename, size_t file_index, bool use_ear
   // Creating the file with SWMR write access
   LOG4CXX_INFO(logger_, "Creating file: " << filename);
   unsigned int flags = H5F_ACC_TRUNC;
+
+  watchdog_timer_.start_timer("H5Fcreate", hdf5_error_definition_.create_duration);
   this->hdf5_file_id_ = H5Fcreate(filename.c_str(), flags, fcpl, fapl);
+  size_t create_duration = watchdog_timer_.finish_timer();
+
   ensure_h5_result(this->hdf5_file_id_, "Failed to create HDF5 file");
   if (this->hdf5_file_id_ < 0) {
     // Close file access property list
@@ -173,26 +181,45 @@ void HDF5File::create_file(std::string filename, size_t file_index, bool use_ear
 
   file_index_ = file_index;
 
+  return create_duration;
 }
 
 /**
  * Close the currently open HDF5 file.
+ *
+ * \return - The hdf5 write metric with the durations of the write and flush calls
  */
-void HDF5File::close_file() {
+size_t HDF5File::close_file() {
   // Protect this method
   boost::lock_guard<boost::recursive_mutex> lock(mutex_);
+
+  size_t close_duration = 0;
   if (this->hdf5_file_id_ >= 0) {
-    ensure_h5_result(H5Fclose(this->hdf5_file_id_), "H5Fclose failed to close the file");
+    watchdog_timer_.start_timer("H5Fclose", hdf5_error_definition_.close_duration);
+    hid_t status = H5Fclose(this->hdf5_file_id_);
+    close_duration = watchdog_timer_.finish_timer();
+    ensure_h5_result(status, "H5Fclose failed to close the file");
     this->hdf5_file_id_ = -1;
   }
+
+  return close_duration;
 }
 
 /**
  * Write a frame to the file.
  *
- * \param[in] frame - Reference to the frame.
+ * \param[in] frame - Reference to the frame
+ * \param[in] frame_offset - The offset in the file to write the frame into
+ * \param[in] outer_chunk_dimension - The size of the outermost dimension of a chunk
+ * \param[in] call_durations - Struct containing hdf5 call durations - write and flush will be updated
+ *                             with the durations of the H5DOwrite_chunk and H5Dflush calls
  */
-void HDF5File::write_frame(const Frame& frame, hsize_t frame_offset, uint64_t outer_chunk_dimension) {
+void HDF5File::write_frame(
+    const Frame& frame,
+    hsize_t frame_offset,
+    uint64_t outer_chunk_dimension,
+    HDF5CallDurations_t& call_durations
+  ) {
   // Protect this method
   boost::lock_guard<boost::recursive_mutex> lock(mutex_);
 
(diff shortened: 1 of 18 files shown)


frameProcessor/src/HDF5File.cpp
```diff
@@ -31,7 +36,8 @@ HDF5File::HDF5File() :
         hdf5_error_flag_(false),
         file_index_(0),
         use_earliest_version_(false),
-        unlimited_(false)
+        unlimited_(false),
+        watchdog_timer_()
```
ajgdls: I'm not sure this line is necessary?

GDYendell:
> At the risk of being annoying by suggesting extra work wink can I suggest you consider using the existing timer functionality available in `IpcReactor` for this, rather than introducing `boost::asio` as another dependency?

I did wonder if I should use it, but it seemed like it did a lot more than I needed in this situation. Do you mean I should register my timers with the existing instance on the FrameProcessorController or that I should create a new isntance just for HDF5File to use.

frameProcessor/src/HDF5File.cpp
GDYendell: Isn't this required to create the WatchdogTimer instance? it will be undefined otherwise, right?

timcnicholls:
> I did wonder if I should use it, but it seemed like it did a lot more than I needed in this situation. Do you mean I should register my timers with the existing instance on the FrameProcessorController or that I should create a new instance just for HDF5File to use?

`IpcReactor` certainly does more than you need but I don't think should be any more complicated to use with a single timer than `asio`. You could run the timers in the controller but it's probably simpler to create a new instance in the watchdog, otherwise you've got to make a reference to the controller's reactor available to the plugin, make sure calls crossing threads are safe etc.

GDYendell:
@timcnicholls IpcReactor slots in pretty cleanly in place of io_service / deadline_timer and it seems to work the same.

frameProcessor/src/HDF5File.cpp
GDYendell: Alan explained that you don't have to explicitly call a default constructor in the initialiser list, but it also does no harm.

GDYendell:
There are decisions to be made over the time thresholds for various messages. Currently, it is using the following:

The threshold for warning when printing call time:
- write: 10ms (This does get printed sometimes when writing to my local SSD)
- flush: 10ms
- open: 1 second

And the threshold for the watchdog timeout error message (for any call) is 5 seconds.

Any thoughts on this @ulrikpedersen , @ajgdls ?

timcnicholls:
> There are decisions to be made over the time thresholds for various messages.

Can the thresholds be configurable? Seems that normal behaviour might vary from one system to another... Also can the stats about these operations be made available in the status report of the FileWriter plugin? I reckon they'd be useful metrics even in systems not using the meta channel.

GDYendell:
I think yes to both. We can configure thresholds on a per-acquisition basis. Would you want the stats to be updated during an acquisition or just once after it is finished?

timcnicholls:
I think it would be good to see the stats refreshed during the acquisition, especially if you're calculating average and max values for each. Then a polling client can watch the behaviour in real time

GDYendell:
I think I can get the ability to configure the timings in on this PR.

To make statistics available during an acquisition I can use the logic for publishing meta data that I have moved to the meta-refactor branch. That needs quite a lot more work for unrelated reasons, so I am going to make a future PR for that, which will probably be after our shutdown.

The key thing we would like to get in for the next release is the logged error when the file writer is completely stuck.

Does that seem reasonable?

timcnicholls:
> Does that seem reasonable?

That's fine by me @GDYendell

GDYendell:
Do we need to be able to separately define warning and error (watchdog) duration thresholds, or could we do something like error_duration = 10 * warning_duration?

If we have both, that will create 8 new configuration parameters (2 each for open, write, flush, close), which seems very bloated.

timcnicholls:
Seems sensible. You could have single config parameter for the multiplier, which reduces a bit, if anyone feels a constant isn't good enough...

GDYendell:
Currently the tests hang in write_parameter after a3e9beb. I don't understand why.

GDYendell:
It doesn't happen on travis...

GDYendell:
Currently all that happens if the watchdog times out is that an error is logged. It would be good if this could propagate to the error status of the FP. Possibly by adding a callback reference to the HDF5ErrorDefinitions_t.

GDYendell:
I have tried writing parameter datasets within a full system and it works fine. My feeling now is that I have broken the tests rather than the code...

GDYendell:
Fixed the tests. It seems it is a bad idea to use shared pointers within a test fixture.

GDYendell:
@timcnicholls @ajgdls @ulrikpedersen @skyfrench 

This has been a noisy PR, so I imagine you have started ignoring it now. This is now ready for review / testing when you get a chance.

Without changing any runtime configuration, this will simply print debug messages (after function exit) saying how long certain calls take in HDF5File.

If any of the *_error_durations are set, e.g.  `fp/config/hdf/file/write_error_duration = 1000`
it will then a) log an error in a separate thread if the call takes longer than the given value (in milliseconds) and b) upgrade the function exit log message to warning level if the call takes longer than 0.1 * the error duration. This fraction can easily be made a config option, but my feeling is it does not need to be. For example (if I set the create_error_duration unreasonably low):

```
/scratch/work/tools/RHEL7-x86_64/odin-data/frameProcessor/src/HDF5File.cpp:158:
 10:45:34,752 FP.HDF5File    INFO  - Creating file: /tmp/test_000001.h5
/scratch/work/tools/RHEL7-x86_64/odin-data/frameProcessor/src/WatchdogTimer.cpp:50:
 10:45:34,752 FP.WatchdogTimer DEBUG - H5Fcreate | Registering 5ms watchdog timer
/scratch/work/tools/RHEL7-x86_64/odin-data/frameProcessor/src/WatchdogTimer.cpp:104:
 10:45:34,757 FP.WatchdogTimer ERROR - H5Fcreate | Watchdog timed out
/scratch/work/tools/RHEL7-x86_64/odin-data/frameProcessor/src/WatchdogTimer.cpp:76:
 10:45:34,810 FP.WatchdogTimer WARN  - H5Fcreate | Call took 58518us
```

Status items have been added to mirror the existing process_frame timings (in fact, these have also been update to use the new shared logic) for specific HDF5 calls. These are updated for every call. For example (timings in microseconds as before):

```
GET fp/status/hdf/timing/0

{
    "value": {
        "last_close": 420356,
        "last_create": 998,
        "last_flush": 149,
        "last_process": 1809,
        "last_write": 749,
        "max_close": 420356,
        "max_create": 1955,
        "max_flush": 316,
        "max_process": 89455,
        "max_write": 88359,
        "mean_close": 210434,
        "mean_create": 1476,
        "mean_flush": 136,
        "mean_process": 1622,
        "mean_write": 730
    }
}
```

This does not yet publish the on the meta channel. I would like to refactor the meta writer so that things such as this are implemented in the base class. Currently any shared meta data has to be implemented in each detector-specific writer. I am working on this in a separate branch.

timcnicholls:
@GDYendell Haven't tested this myself but looks good and the configuration logic looks sensible and neat. I notice you chose fully implementing mutexes in IpcReactor to avoid the issues we discussed last week. I have no objection to this per se but are you happy there are no negative side effects on e.g. the frameReceiver? I guess the timer shutdown delay in tests is avoided by having a fast heartbeat timer in the Watchdog?

GDYendell:
I wouldn't like to claim there are definitely no side effects, but I don't see why it would cause any issues and it seems to work completely fine in my test system.

I think we should definitely give this a good run against the lab 29 Excalibur and the system tests before merging.

timcnicholls:
HI @gdy - no problem. If you're happy with the testing you've done I have no reason to complain :smile: Do you want to leave the PR open until we can test on lab 29 (which may well be some time, I don't know if remote testing is feasible currently) or merge onto master and test before releasing? I'd probably favour the latter so we don't leave PRs dangling too long...

Re the shutdown delay - I think I misunderstood the original issue about his you described last week when testing. I thought the wait in poll was slowing down termination of each test.

GDYendell:
That is a good point... I will try and run the system tests this week and then we can merge and fix after if needed.

> I thought the wait in poll was slowing down termination of each test.

Ah. Yes, there was a delay closing and opening each file. But, now that the tick is 1ms I think it is not noticeable. The tests still complete basically instantaneously, as before.

ajgdls:
I'll check with Scott, if the detector is on then remote testing will be fine, as I can set up the IOCs easily.  It just depends on whether the detector has been powered down.

timcnicholls:
I can also test it with the single-FEM EXCALIBUR slice we develop firmware on if that helps.

ajgdls:
It does look like I can't access anything in the lab network....

timcnicholls:
> It does look like I can't access anything in the lab network....

I'd be surprised if the detector head is on - I think it needs an N2 cylinder for the dry air supply to the head?

GDYendell:
I think the build failures were an existing problem that were further provoked by these changes. I believe 98f32c3 fixes it and makes some other improvements.

GDYendell:
I have run about a million frames through this now on the I07 1M at 100Hz, 300K of which were on the very latest commit (d76b111), so I am relatively confident it doesn't break anything.

ajgdls:
Hi @GDYendell so now it builds OK and all tests pass on Travis repeatedly?  You mentioned to me about parameter failures on the previous released version when you built with optimisation, you have now resolved all of those issues in this PR too?

GDYendell:
I think so. I have re-run the build on this commit three times now and it passed those builds. I think the main problem was that num_frames was not defined on the DatasetDefinition in the parameter test, so it would use a random bit of memory. Usually this ends up being quite a big number.

ajgdls:
Yes OK I see, makes sense.  Looks good, we should definitely report these numbers out through the EPICS ADOdin interface.

ajgdls:
Although you've tested it yourself against the hardware, as I have no hardware to test it with (and I trust you of course!) I think I'll tick it so we can get it merged sooner rather than later.

PR #231, opened 2020-03-12
set the default acquid on frameProcessor's frames to '' again

wnichols1:
setting it to zero was probably done to account for a bug elsewhere; agreed in software meeting 12/3/20.


timcnicholls:
Ah, so this affects detector-specific plugins differently, depending on whether they copy the frame metadata from the incoming shared buffer (like PERCIVAL) or construct it from scratch (like EXCALIBUR)...

wnichols1:
I suspect that someone was running some python script and the default value for an int (0) got turned into a string or something....

timcnicholls:
I think that change crept in when we did a big refactor of the `Frame` implementation last year and the metadata was separated out into a different class. At that point the incoming shared memory buffers get wrapped in a frame object with meta data explicitly initialised.

PR #232, opened 2020-03-16
Fix a bug that was stopping the message type and value propagating

ajgdls:
Finally decided to fix this, it's been bugging people for ages...

The message type and value is no longer logged as illegal by every plugin on every configure message after this fix.  Whoop whoop.


timcnicholls:
Oh I dunno, seems like a complex and messy change to me :wink:

PR #233, opened 2020-03-17
IpcMessage Nack

ajgdls:
Related to #227 I will remove these commits from the specific File Writing check PR.
This PR adds the IpcMessage set_nack method which allows an IpcMessage to have its type set to nack with an appropriate error message.


ajgdls:
As @skyfrench is on holiday I think I've removed her approval from these non-controversial (I hope) PRs.

PR #235, opened 2020-03-20
Rename members

wnichols1:


ajgdls:
I wanted to see what other people think about the name changes.  I think in general if the names changes make sense that is OK, but the addition of the "m_" might make this inconsistent with the rest of odin data so I was waiting to see what @skyfrench and @timcnicholls said about this PR.

timcnicholls:
The convention elsewhere is to use trailing prefixes on member variables. @skyfrench should be consulted about this PR as she is the core author of the frameSimulator

timcnicholls:
I've added reviewers now

wnichols1:
that's great thank you very much.

I meant to refer to these actually which are on master:
FrameSimulatorPluginUDP.h:        std::vector<struct sockaddr_in> m_addrs;
FrameSimulatorPluginUDP.h:        int m_socket;
FrameSimulatorPluginUDP.h:        pcap_t *m_handle;

thanks!

w.

________________________________

skyfrench:
No major objections to name changes - but would it make more sense for the change to add a trailing _ to match the rest of odin data, rather than m_?

wnichols1:
That's YOUR decision folks. In my view I'll stick with m_ because it's clear and an established convention. This trailing underscore is just a fashion. But I will accept the decision of the group.

timcnicholls:
Trailing underscores are no more of a fashion than `m_` prefixes. Both are stylistic conventions with arguments for and against. `m_` is a partial implementation of Hungarian notation. Trailing underscores are the convention in the [Google C++ style guide](https://google.github.io/styleguide/cppguide.html), which is where we took it from. I agree with @skyfrench and suggest that consistency with an existing code base is more important than the choice of style.

wnichols1:
ok Tim!

wnichols1:
at your request I have taken out my nice m_ ! I hope we can approve this now.

wnichols1:
ok thanks folks! very good. Can you merge this please? I don't have permissions!

PR #236, opened 2020-03-24
Add total buffer count to FR

ajgdls:
It turns out that the mapped buffer count I was using to calculate the pct free was not counting the buffers in the FP applications as I had thought.
I have added a new total buffers item to the status and I am using that to calculate the percentage of free buffers.
I've added a check for dividing by 0 in the calculation.

### DIFF
--- a/frameReceiver/src/FrameReceiverController.cpp
+++ b/frameReceiver/src/FrameReceiverController.cpp
@@ -629,6 +629,9 @@ void FrameReceiverController::configure_buffer_manager(OdinData::IpcMessage& con
       LOG4CXX_DEBUG_LEVEL(1, logger_, "Configured frame buffer manager of total size " <<
           max_buffer_mem << " with " << buffer_manager_->get_num_buffers() << " buffers");
 
+      // Record the total number of buffers in the system here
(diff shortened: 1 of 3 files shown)


ajgdls:
@GDYendell will test these changes using the same applications that he was able to crash it with previously (by starving the system of buffers and killing FP applications).

frameReceiver/src/FrameReceiverController.cpp
```diff
@@ -629,6 +629,9 @@ void FrameReceiverController::configure_buffer_manager(OdinData::IpcMessage& con
       LOG4CXX_DEBUG_LEVEL(1, logger_, "Configured frame buffer manager of total size " <<
           max_buffer_mem << " with " << buffer_manager_->get_num_buffers() << " buffers");
 
+      // Record the total number of buffers in the system here
```
timcnicholls: Can you move assigning to 'total_buffers_' above the immediately preceding debug log call and avoid calling 'get_num_buffers()' twice?

ajgdls: Yes good idea.

PR #237, opened 2020-03-31
Reorder packets

wnichols1:
ok this is far from finished


ulrikpedersen:
Hi William,
I'm closing this pull request for a couple of reasons:

* We would like to have a discussion and agreement on the direction of travel between Odin core developers **before** issuing code/pull requests.
  - The natural place to have this is the developers meeting
* Pull Requests really should be pretty much complete, rather than Work In Progress as your message seems to indicate.
* Pull Requests should come with a full explanation of the feature/fix.

wnichols1:
I had discussed it with people on slack, and Tim Nichols said I could give it a go. Since we aren't having any dev meetings, the slack channel is the standin for it.
I know it was a bit tentative, but the PR gives people a chance to see the diffs quite easily, and it was a step towards the finished solution.

kind regards,
will.
________________________________

PR #238, opened 2020-04-22
Updated frame writing timeout from an info log message to an error condition.

ajgdls:
A timeout should put the file writer into an error state to ensure somebody notices it and has to clear the error.  Currently it only logged an info message which is quite missable.  JIRA ticket BC-1266.


ajgdls:
Removed @ulrikpedersen as 3 ticks are enough for me :)

PR #239, opened 2020-05-14
Perform a calculation to make sure file sizes are correct in block mode.

ajgdls:
In the current version when running in block mode the total number of expected frames is set as the dataset size for each file, instead of block sizes.  This PR is to fix that bug.

### DIFF
--- a/frameProcessor/src/Acquisition.cpp
+++ b/frameProcessor/src/Acquisition.cpp
@@ -267,7 +267,7 @@ void Acquisition::create_file(size_t file_number) {
     // Calculate the number of frames required for this dataset in the case that
     // the acquisition is using block mode.
     int frames_per_file = blocks_per_file_ * frames_per_block_;
-    if (frames_per_file > 2){
+    if (frames_per_file > 1){
       if (file_number * frames_per_file > frames_to_write_){
         // This is the final file creation which may contain less than a full block of frames
         dset_def.num_frames = frames_to_write_ % frames_per_file;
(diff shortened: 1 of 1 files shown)


frameProcessor/src/Acquisition.cpp
```diff
@@ -264,6 +264,18 @@ void Acquisition::create_file(size_t file_number) {
...
+    int frames_per_file = blocks_per_file_ * frames_per_block_;
+    if (frames_per_file > 1){
+      if (file_number * frames_per_file > frames_to_write_){
+        // This is the final file creation which may contain less than a full block of frames
+        dset_def.num_frames = frames_to_write_ % frames_per_file;
+      } else {
+        // This is not the final file, it will contain a full block of frames
+        dset_def.num_frames = frames_per_file;
+      }
+    }
```
GDYendell, line 278: Could the `dset_def.num_frames = frames_to_write_;` from above go into the else of this block with a comment? I think it would be clearer.

ajgdls: I'm not sure moving it made it any clearer, but I want a tick and I've added a comment which does explain what the line is doing.

GDYendell: Is that because you thought it was clear before or because you think it still isn't clear?

ajgdls:
Thanks all for the fast response, the beamline who are relying on this will be happy at the turnaround (I hope).

PR #240, opened 2020-05-18
Using file_number to work out data size is incorrect.

ajgdls:
The wrap needs to be calculated using the file_number and concurrent_processes and this
value can be used to decide exactly how many frames of data belong in any
individual file.

### DIFF
--- a/frameProcessor/src/Acquisition.cpp
+++ b/frameProcessor/src/Acquisition.cpp
@@ -270,9 +270,10 @@ void Acquisition::create_file(size_t file_number, HDF5CallDurations_t& call_dura
 
     // Calculate the number of frames required for this dataset in the case that
     // the acquisition is using block mode.
+    int wrap = (file_number/concurrent_processes_)+1;
     int frames_per_file = blocks_per_file_ * frames_per_block_;
     if (frames_per_file > 1){
-      if (file_number * frames_per_file > frames_to_write_){
+      if (wrap * frames_per_file > frames_to_write_){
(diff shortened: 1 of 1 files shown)


ajgdls:
PR #239 did not work for all cases, it was using file_number to decide how many frames are required in any given file.  However, the acquistion class in each FP knows how many frames it is going to get only (not the whole total) and so it needs to work out a wrap number and use this instead.

frameProcessor/src/Acquisition.cpp
```diff
@@ -270,9 +270,10 @@ void Acquisition::create_file(size_t file_number, HDF5CallDurations_t& call_dura
 
     // Calculate the number of frames required for this dataset in the case that
     // the acquisition is using block mode.
+    int wrap = (file_number/concurrent_processes_)+1;
     int frames_per_file = blocks_per_file_ * frames_per_block_;
     if (frames_per_file > 1){
-      if (file_number * frames_per_file > frames_to_write_){
+      if (wrap * frames_per_file > frames_to_write_){
```
GDYendell, line 276: Is it possible to have `wrap * frames_per_file > frames_to_write_` but not `frames_per_file > 1`?

GDYendell: Ignore.

ajgdls: OK, I would say the logic is different so the if still needs to be there.  I guess you figured that out.

Issue #213, opened 2019-08-08
Obtuse Error Message when a frame with an unexpected dataset name is received


ANeaves:
If the FileWriterPlugin receives a frame that has a dataset name that isn't one of the ones it's been configured to receive, the error message produced comes from Acquisition.cpp, and looks like the following:
`FP.Acquisition ERROR - Out of Range exception: map::at`
which stops the frame from being saved to a file, but is not very clear what the reason is. I had to dig into the Acquisition code to find out that it was the dataset name issue causing the error.

ajgdls:
See PR #241

PR #241, opened 2020-05-21
Improved handling of specific error case where dataset has not been created.

ajgdls:
See JIRA ticket BC-1319.

### DIFF
--- a/frameProcessor/src/Acquisition.cpp
+++ b/frameProcessor/src/Acquisition.cpp
@@ -414,66 +414,78 @@ bool Acquisition::check_frame_valid(boost::shared_ptr<Frame> frame)
 {
   bool invalid = false;
   const FrameMetaData frame_meta_data = frame->get_meta_data();
-  DatasetDefinition dataset = dataset_defs_.at(frame_meta_data.get_dataset_name());
+  DatasetDefinition dataset;
+  try
+  {
+    dataset = dataset_defs_.at(frame_meta_data.get_dataset_name());
 
-  // Check if frame compression is set to unknown and raise an error.
-  // Otherwise verify the compression is consistent with the dataset definition.
-  CompressionType frame_compression_type = frame_meta_data.get_compression_type();
-  if (frame_compression_type == unknown_compression) {
-    std::stringstream ss;
-    ss << "Invalid frame: Frame has unknown compression for dataset " << dataset.name;
-    last_error_ = ss.str();
-    LOG4CXX_ERROR(logger_, last_error_);
-    invalid = true;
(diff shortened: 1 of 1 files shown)


ajgdls:
Resolves issue #213

frameProcessor/src/Acquisition.cpp
```diff
@@ -414,66 +414,78 @@ bool Acquisition::check_frame_valid(boost::shared_ptr<Frame> frame)
...
-          " for dataset " << dataset.name <<
-          " (0: Unknown, 1: None, 2: LZ4, 3: BSLZ4, 4: Blosc)";
-    last_error_ = ss.str();
-    LOG4CXX_ERROR(logger_, last_error_);
-    invalid = true;
-  }
+  DatasetDefinition dataset;
+  try
+  {
+    dataset = dataset_defs_.at(frame_meta_data.get_dataset_name());
```
GDYendell, line 420: Isn't this the only line that needs to be in the try block? Where else could out_of_range exception happen?

ajgdls: The point is if you can't get a valid dataset here then none of the rest of this code block makes any sense and would simply fail later on as there are several assumptions that the dataset is not null.  So by putting the whole block in the try catch none of that extra code is called if the exception happens here.

GDYendell: Can you just return in the catch?

ajgdls: You can but I really don't like doing that.
Do you want other catches to be added?

GDYendell: I just thought that smaller try blocks were good because you can tell where the exception is. What do you mean other catches?

ajgdls: Catch different exceptions.  Anyway how about I catch it (small like you want), and then do not proceed if it fails.  I will add a check, I still like returns at the end of the method otherwise it isn't always easy to trace but I will shorten the try catch.

ajgdls: Wait here....

GDYendell: Another option is using `find` to check for the key instead of try catch.

ajgdls:
Well that took a long time, I'm not sure why...  @GDYendell I have tightened the try catch around the specific failure point.  I know you already approved but would you mind casting an eye to confirm it is what you were asking for?

Thanks.

GDYendell:
I think it is much clearer now :)

PR #242, opened 2020-07-02
Meta Writer Refactor

GDYendell:
This is a PR of two parts:

- Update the frameProcessor to publish file writing metrics over the meta data channel
- A big refactor of the meta writer application to
    - Make the base class fully functional and handle the file writing metrics with no detector version required
    - Allow it to be easily extended for detector-specific meta data

Most of the logic has been moved out of EigerMetaWriter into the base class. There is an accompanying eiger-detector PR here: https://github.com/dls-controls/eiger-detector/pull/8

The application can be run without specifying a writer class and the base class will be used instead. This will listen for messages from the FileWriterPlugin (the file writing metrics). If it does get detector specific messages it will ignore them.

I think this is pretty much ready, but there are a few outstanding questions and more things that could either be resolved in this PR or afterwards:
- It might need to be extended to fulfill requirements for tristan, which I believe has some more exotic meta data
- There are two TODOs that I don't think are urgent: 
    - Fixed in 79a9b89 ~~There is some logic to count the number of status requests before deleting an acquisition - this is to make sure that a client (our areaDetector driver) sees the acquisition in the finished state and updates accordingly. I think this is messy and it would be better to have some kind of command to clear the old acquisitions out, but this might cause problems if there are multiple clients for some reason. EDIT: I think this can be fixed by only removing finished acquisitions if there is more than one. This is a fix I patched on VMXi for a problem with the meta seemingly hanging acquisitions less than a second in duration because the IOC status wouldn't update in time to see it had finished.~~
    - It does not support rewinding. I don't think it will be a big change to make this work, but I don't know if it is required.
- I would have liked to add some tests, but they are not currently built into the module and implementing this would require some further discussion
- In the version information (shown below) there is no way to tell what actual detector module is. Should it use (e.g.) eiger-detector as the key so that this is clearer? Also, if the base class MetaWriter class is used, the writer dict will just be a duplication of odin-data.

```
"odin-data": {
    "full": "1.4.0+88.g5a4838a.dirty",
    "major": "1",
    "minor_version": "4",
    "patch_version": "0",
    "short_version": "1.4.0"
},
"writer": {
    "full": "1-7-1+3.g59eaaff.dirty",
    "major": "1",
    "minor": "7",
    "patch": "1",
    "short": "1.7.1"
}
```
- I am not sure now if I needed to copy in that versioneer file
- I have updated the numpy version requirement to 1.14+. Hopefully this is OK. If not I could work around it.
- I think the parameter names in the meta messages from FrameProcessorController to MetaWriter should be renamed to match what they are currently being used for. E.g. `parameter` stores the unique message type, `plugin` stores the class name that created the message.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 /external
-/build*
-/prefix
+/*build*
+/*prefix/
 *.pyc
 /tools/python/venv/
 /.settings
@@ -23,3 +23,6 @@ tools/imagej/plugins/Odin-Data/build/**
 *.jar
 tools/imagej/odin-imagej
 tools/imagej/build/
+
+tools/python/build/
+tools/python/dist/
--- a/common/include/CMakeLists.txt
+++ b/common/include/CMakeLists.txt
@@ -7,6 +7,7 @@ SET(HEADERS ClassLoader.h
         IpcMessage.h
         IpcReactor.h
         IVersionedObject.h
+        Json.h
         logging.h
         OdinDataException.h
         SegFaultHandler.h
@@ -18,4 +19,3 @@ SET(ZMQ_INCLUDE_DIR zmq)
 INSTALL(FILES ${HEADERS} DESTINATION include)
 INSTALL(DIRECTORY ${RAPIDJSON_INCLUDE_DIR} DESTINATION include)
 INSTALL(DIRECTORY ${ZMQ_INCLUDE_DIR} DESTINATION include)
-
--- a/common/include/Json.h
+++ b/common/include/Json.h
@@ -1,0 +1,67 @@
+/*
+ * Json.h
+ *
+ *  Created on: 10 Dec 2020
+ *      Author: Gary Yendell
+ */
+
+#ifndef JSON_H_
+#define JSON_H_
+
+#include <string>
+#include <vector>
+
+#include "rapidjson/document.h"
+
+namespace OdinData
+{
+
+class JsonDict
+{
+public:
+  JsonDict();
+
+  /** Add a scalar value to the underlying rapidjson::Document
+   *
+   * \param[in] key - The key of the item to add
+   * \param[in] value - The value of the item to add
+   */
+  template <typename T> void add(const std::string& key, T value)
+  {
+    rapidjson::Value json_value(value);
+    add(key, json_value);
+  }
+
+  void add(const std::string& key, const std::string&  json_value);
+
+  /** Add a vector value to the underlying rapidjson::Document
+   *
+   * \param[in] key - The key of the item to add
+   * \param[in] value - The value of the item to add
+   */
+  template <typename T> void add(const std::string& key, std::vector<T> value)
+  {
+    rapidjson::Value json_value;
+    json_value.SetArray();
+    typename std::vector<T>::iterator it;
+    for (it = value.begin(); it != value.end(); it++) {
+      json_value.PushBack(*it, document_.GetAllocator());
+    }
+    add(key, json_value);
+  }
+
+  /** Return a json representation of the dictionary
+   *
+   */
+  std::string str() const;
+
+private:
+  void add(const std::string& key, rapidjson::Value& json_value);
+
+  // RapidJSON document object
+  rapidjson::Document document_;
+};
+
+} /* namespace OdinData */
+
+#endif /* JSON_H_ */
--- a/common/src/Json.cpp
+++ b/common/src/Json.cpp
@@ -1,0 +1,54 @@
+/*
+ * Json.cpp
+ *
+ *  Created on: 10 Dec 2020
+ *      Author: Gary Yendell
+ */
+
+#include "rapidjson/writer.h"
+#include "rapidjson/stringbuffer.h"
+
+#include "Json.h"
+
+namespace OdinData
+{
+
+JsonDict::JsonDict()
+{
+  document_.SetObject();
+}
+
+/** Add a key value pair to the underlying rapidjson::Document
+ *
+ * \param[in] key - The key of the item to add
+ * \param[in] value - The value of the item to add
+ */
+void JsonDict::add(const std::string& key, rapidjson::Value& json_value)
+{
+  rapidjson::Value json_key(key.c_str(), document_.GetAllocator());
+  document_.AddMember(json_key, json_value, document_.GetAllocator());
+}
+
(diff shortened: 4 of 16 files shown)


ajgdls:
@GDYendell do you agree the best thing for me to do is to try updating the Tristan system to work with this?  The changes here are substantial and I couldn't just look at them and say "yes this will all be fine with Tristan".
You aren't expecting this to be wildly changed again I assume (before I embark on those Tristan changes)?

GDYendell:
@ajgdls Yeah I think the best thing to do is just look at the new EigerMetaWriter and create the equivalent message type handler methods for Tristan.

I have no plans to change this much at all. I am hoping comments here only lead to minor fixes / additions.

One thing I am thinking about doing based on testing this against the new Eiger9M is making a stop() method that the detector class can override to say it is not ready to stop and, in the Eiger case, linking this to the end message from the detector. It seems that for very short acquisitions without a pause after arming the global header comes after the data frames and is lost. This shouldn't require any additional changes for you compared to this version (unless you want to make use of it).

MattTaylorDLS:
"It seems that for very short acquisitions without a pause after arming the global header comes after the data frames and is lost" - this happens on the other Eigers as well, and can cause issues, the eiger data path (fan->fp->fw-meta) relies on receiving the header before any data frames.
On the MX beamlines, we got GDA to wait until the fan ready flag was set (Which is set when the header is received), before triggering the detector.

GDYendell:
Hi @MattTaylorDLS !

Oh OK thanks. This was just in my manual testing, so it is good to know GDA will work around it. Other than the the fan logging warnings and the global header meta data not being in the file everything seemed to work OK. What problems might occur?

MattTaylorDLS:
Hi!

My memory is a little hazy but I think either the frames won't be written to file until the header is sent and processed. That or the meta frames. I'll have a look through the code again to see if it jogs my memory.

MattTaylorDLS:
Think i've remembered - The fan needs to receive the global header first in order to get the acquisition ID from it, which it then attaches to all the frames it sends out. And the FR/FP needs this, or it'll throw the frame away because either a) it has no ID on it, or b) it has the ID from the previous acquisition.

There's also a "danger" on stopping an acquisition based on seeing the End of Series message. Because again it could arrive out of turn, and arrive before some of the last frames come in so you'll drop frames at the end of the acquisition. The frames arrive out of turn because of the way we connect to the Eiger - with 4 zmq channels, as connecting with just 1 didn't seem to get the full speed out of the 40gb connection, I guess cause it was split into 4 x 10gb wires. If you now have a single 40gb connection, then it might be worth checking whether you can get away with a single zmq connection to the eiger, in which case you shouldn't receive frames out of order. It might be the case with the 9M that you don't need the full speed, so you might be ok with 1 connection anyway.

GDYendell:
> The fan needs to receive the global header first in order to get the acquisition ID

The fan was updated to be configured with the acquisition ID to insert on the outgoing messages (when the Eiger2 turned up and the header appendix was missing). We never stopped using this method. 

> There's also a "danger" on stopping an acquisition based on seeing the End of Series message. Because again it could arrive out of turn, and arrive before some of the last frames come in so you'll drop frames at the end of the acquisition.

That is good to know. I think I am going to go with FPs finished AND end message received, so that should be safe.

MattTaylorDLS:
Ah yes forgot we'd done that. Might be ok then, I've just had a quick look through and I can't see anything else. Probably best to wait for it if you can though.

Yeah think that should be safe for the end message.

ajgdls:
OK @GDYendell first question:

I need to create a dataset for each FP to store the time slice information.  Can I 
a) Find out how many FPs I've got programmatically to populate my DETECTOR_DATASETS (best option)
b) Pass in the value from command line options
c) None of the above (worst option)

GDYendell:
The `MetaListener` is given a list of data endpoints which it then counts and stores as `_process_count` for the lifetime of the application. It passes this number to the writer class when it creates an instance, so it could be used in the `TristanMetaWriter` `__init__` to append to as required DETECTOR_DATASETS (probably before you call `super.__init__`). Would that do what you want?

ajgdls:
Yes I think so - thanks.  I'll give that a try.

ajgdls:
I've been working with this and Tristan and I think there are a couple of issues I might need to suggest updates for.

1) Tristan has 0 frames to write when started (unlimited) as we never know in advance the number of frames.  This results in errors for the frame statistics who have caching turned on.
2) Tristan also has 3 frame types (event ID, timestamp, energy) and I'm getting 3 messages the same all going into the meta writer same dataset.  I could handle these as a special case in the Tristan specific part, I would need to just make sure I get hold of the dataset name.
3) Cache appears to be all or nothing.  For Tristan I think nothing is risky as we could be writing many thousands of individual values per second, but I will probably try this out.  I cannot set the cache to all for Tristan as we don't know the number of frames in advance (see 1).  I think it would be good to be able to create a cache for blocks of data which when filled up (or timer expires) get flushed to disk.

I'll make some updates to the branch and if I think they are working I'll push them somewhere for @GDYendell to check.

ajgdls:
@GDYendell did I give you the branch that I had developed?

GDYendell:
I don't think so

GDYendell:
Two Eiger beamlines at DLS are using this and are happy with it. Alan is working on a tristan meta branch on top of this one, so we can resolve the rest of the outstanding points in the initial description in that PR. Given that, I think this is ready to merge if people are happy with what is here so far. I have rebased on the latest master and it still works OK with the simulator. Can we discuss this in the next meeting?

GDYendell:
Rebased and added a couple of commits, including fixing the version information. It now appears like this:
```
{
    "value": [
        {
            "eiger-detector": {
                "full": "1-9-0beta2+1.gcfb2c1f",
                "major": "1",
                "minor": "9",
                "patch": "0",
                "short": "1.9.0"
            },
            "odin-data": {
                "full": "1.5.0+36.g05e15ec",
                "major": "1",
                "minor": "5",
                "patch": "0",
                "short": "1.5.0"
            }
        }
    ]
}
```

GDYendell:
This is ready now @timcnicholls

timcnicholls (approved):

timcnicholls:
Thanks @GDYendell - I have no objections to this at all. If @ajgdls is happy we can merge.

ajgdls (approved):
I'm good with those extra changes, I agree with the version information choices.

PR #244, opened 2020-08-07
rename some variables to conform to convention, and fix a few comments

wnichols1:
I did bring this up at the dev meeting a month ago and got approval to pr it.

### DIFF
--- a/frameProcessor/include/Acquisition.h
+++ b/frameProcessor/include/Acquisition.h
@@ -94,16 +94,16 @@ public:
   size_t concurrent_rank_;
   /** Number of frames to write consecutively in a file */
   size_t frames_per_block_;
-  /** Number of blocks to write in a file  */
+  /** Number of blocks to write in a file; 0=put all blocks in same file  */
   size_t blocks_per_file_;
   /** HDF5 call error definitions */
   const HDF5ErrorDefinition_t& hdf5_error_definition_;
 
 private:
   /** The current file that frames are being written to */
-  boost::shared_ptr<HDF5File> current_file;
+  boost::shared_ptr<HDF5File> current_file_;
   /** The previous file that frames were written to, held in case of late frames  */
-  boost::shared_ptr<HDF5File> previous_file;
+  boost::shared_ptr<HDF5File> previous_file_;
   /** Most recently generated error message */
   std::string last_error_;
 };
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -129,8 +129,8 @@ void FileWriterPlugin::process_frame(boost::shared_ptr<Frame> frame)
   boost::mutex::scoped_lock cflock(close_file_mutex_);
   boost::lock_guard<boost::recursive_mutex> lock(mutex_);
 
-  // Check the frame against the current acquisition and start a new one if
-  // the frame object contains a different acquisition ID from the current one
+  // check it matches the current (or next) acquisition.
+  // frames that don't match are dropped / ignored.
   if (frame_in_acquisition(frame)) {
 
     if (writing_) {
@@ -762,12 +762,13 @@ bool FileWriterPlugin::reset_statistics()
   return true;
 }
 
-/** Check if the frame contains an acquisition ID and start a new file if it does and it's different from the current one
+/** This function checks the acquisition id of the frame matches that of the current acquisition,
+ * subject to two caveats:
+ *  i) if the frame-aid is "", it matches anything
+ *  ii) this function has the side-effect of moving from the current-acq to
+ *        the next-acq if that helps us match the frame.
  *
- * If the frame object contains an acquisition ID, then we need to check if the current acquisition we are writing has
- * the same ID. If it is different, then we close the current file and create a new one and start writing.
- * If we are not currently writing then we just create a new file and start writing.
- *
+ *  The function will set an error if the frame does not match a writing acquisition. 
  * \param[in] frame - Pointer to the Frame object.
  */
 bool FileWriterPlugin::frame_in_acquisition(boost::shared_ptr<Frame> frame) {
--- a/frameProcessor/src/Acquisition.cpp
+++ b/frameProcessor/src/Acquisition.cpp
@@ -165,10 +165,10 @@ ProcessFrameStatus Acquisition::process_frame(boost::shared_ptr<Frame> frame, HD
       // or if no master frame has been defined. If either of these conditions
       // are true then increment the number of frames written.
       if (master_frame_.empty() || master_frame_ == frame_dataset_name) {
-        size_t dataset_frames = current_file->get_dataset_frames(frame_dataset_name);
+        size_t dataset_frames = current_file_->get_dataset_frames(frame_dataset_name);
         frames_processed_++;
         LOG4CXX_TRACE(logger_, "Master frame processed");
-        size_t current_file_index = current_file->get_file_index() / concurrent_processes_;
+        size_t current_file_index = current_file_->get_file_index() / concurrent_processes_;
         size_t frames_written_to_previous_files = current_file_index * frames_per_block_ * blocks_per_file_;
         size_t total_frames_written = frames_written_to_previous_files + dataset_frames;
         if (total_frames_written == frames_written_) {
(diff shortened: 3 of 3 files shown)


frameReceiver/src/FrameReceiverUDPRxThread.cpp
```diff
@@ -135,3 +135,4 @@ void FrameReceiverUDPRxThread::handle_receive_socket(int recv_socket, int recv_p
 
   FrameDecoder::FrameReceiveState frame_receive_state = frame_decoder_->process_packet(bytes_received, recv_port, &from_addr);
 }
+
```
timcnicholls, line 138: Is this a deliberate change?

frameProcessor/include/Acquisition.h
```diff
@@ -90,15 +90,15 @@ class Acquisition : public MetaMessagePublisher
   size_t concurrent_processes_;
   /** Rank of this file writer */
   size_t concurrent_rank_;
-  /** Number of frames to write consecutively in a file */
+  /** Number of frames to write consecutively in a file; 0= */
```
ajgdls, line 93: I think this is missing some text

frameReceiver/src/FrameReceiverUDPRxThread.cpp
wnichols1: no

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -120,8 +120,7 @@ void FileWriterPlugin::process_frame(boost::shared_ptr<Frame> frame)
   boost::mutex::scoped_lock cflock(close_file_mutex_);
   boost::lock_guard<boost::recursive_mutex> lock(mutex_);
 
-  // Check the frame against the current acquisition and start a new one if
-  // the frame object contains a different acquisition ID from the current one
+  // Check the frame against the current acquisition and only process it if it does.
```
ajgdls, line 123: Only process if it does what?

frameProcessor/src/HDF5File.cpp
```diff
@@ -189,7 +189,7 @@ void HDF5File::close_file() {
  *
  * \param[in] frame - Reference to the frame.
  */
-void HDF5File::write_frame(const Frame& frame, hsize_t frame_offset, uint64_t outer_chunk_dimension) {
+void HDF5File::write_frame(const Frame& frame, hsize_t frame_offset, uint64_t outer_dim_qty) {
```
ajgdls, line 192: I'm not sure about renaming away from chunks.  Chunks have a very specific meaning in HDF5, are we sure this is still clear?  Actually I think it is reasonably clear in this context.

ajgdls:
General comment, there are conflicts that will need to be resolved.

timcnicholls:
The PR is reporting conflicts. Your clone needs to be updated and your branch rebased from master possibly?

frameReceiver/src/FrameReceiverUDPRxThread.cpp
timcnicholls: I suggest you undo this commit then to avoid any confusion on history.

frameProcessor/src/HDF5File.cpp
GDYendell: These changes seem to assume that outer_chunk_dimension is always 1, but it is not - it is configurable.

I think the terms chunk and dimension are important to have in this context - they both have specific meaningful hdf5 definitions. outer_chunk_dimension_size might be a more complete description, but that is a bit excessive.

frameProcessor/include/Acquisition.h
wnichols1: ok fixed.

frameProcessor/src/HDF5File.cpp
wnichols1: well you are welcome to object; but I was confused so I renamed it. My thought was that outer_chunk_dimension looks like an index telling you which IS the outer_dimension, but it's always dset.dataset_dimensions[0] that is augmented. The variable is there to tell you how much to increment the outer dimension instead. That is why I renamed it. I could also rename it "outer_chunk_dimension_quantity" if you prefer. I think the fact we are using chunks is assumed.

GDYendell: Maybe we could add some more detail to the parameter description in the doc string?

frameReceiver/src/FrameReceiverUDPRxThread.cpp
wnichols1: yes sorry; I have amended the commit.

frameProcessor/src/HDF5File.cpp
wnichols1: yes ok I will do that.

frameProcessor/src/FileWriterPlugin.cpp
wnichols1: I have to disagree with the original comment, which is why I changed this. The doco for bool FileWriterPlugin::frame_in_acquisition(boost::shared_ptr<Frame> frame) is wrong in my view.
Because the aqu id MUST match the id on the current / next acqu for it to be saved. So I will
review and  rewrite these comments.

GDYendell: That is true - it either matches current or next or it's an error. That doc string should be updated and this comment could say something more generic like "Check if acquisition ID on frame is valid"

frameProcessor/src/HDF5File.cpp
```diff
@@ -199,21 +203,22 @@ void HDF5File::write_frame(const Frame& frame, hsize_t frame_offset, uint64_t ou
       << "] name [" << frame.get_meta_data().get_dataset_name() << "]");
   HDF5Dataset_t& dset = this->get_hdf5_dataset(frame.get_meta_data().get_dataset_name());
 
-  // We will need to extend the dataset in 1 dimension by the outer chunk dimension
-  // For 3D datasets this would normally be 1 (a 2D image)
-  // For 1D datasets this would normally be the quantity of data items present in a single chunk
+  // We will need to extend the dataset in 1 dimension by the outer_dim_qty
+  // For 3D datasets outer_dim_qty == 1 (a 2D image)
+  // For 1D datasets this would normally be the quantity of data items present in a single frame
```
wnichols1, line 208: I think we can remove this comment as it's now done above.

wnichols1:
I've updated that - does it look ok? I am going to wait a week before resolving the conflicts because Elettra are currently having problems building and I wish to keep percival repos constant. I will squash everything into a single commit.

frameProcessor/src/HDF5File.cpp
```diff
@@ -210,14 +210,17 @@ size_t HDF5File::close_file() {
  *
  * \param[in] frame - Reference to the frame
  * \param[in] frame_offset - The offset in the file to write the frame into
- * \param[in] outer_chunk_dimension - The size of the outermost dimension of a chunk
+ * \param[in] outer_dim_qty How much to increase the dimensionality of the dataset to
+ *              create space for the new frame. Dimension[0] is incremented this amount.
+ *              For a 1D dataset, this is the number of new items, for a 3d dataset, it will
+ *              be 1 to create a new frame at [f, :, :].
```
GDYendell, line 216: I don't the changes to this method make things clearer. Fundamentally, this parameter is the size of the outermost dimension of a chunk, as the original description stated. The primary use is for the `offset` parameter to `H5DOwrite_chunk`, which is determined directly from the size of a chunk in the dataset. Its use as the amount the dataset is extended by is completely secondary.

wnichols1: ok no problem I will remove it then!

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -762,12 +762,13 @@ bool FileWriterPlugin::reset_statistics()
...
 }
 
-/** Check if the frame contains an acquisition ID and start a new file if it does and it's different from the current one
- *
- * If the frame object contains an acquisition ID, then we need to check if the current acquisition we are writing has
- * the same ID. If it is different, then we close the current file and create a new one and start writing.
- * If we are not currently writing then we just create a new file and start writing.
+/** This function checks the acquisition id of the frame matches that of the current acquisition,
+ * subject to two caveats:
+ *  i) if the frame-aid is "", it matches anything
```
ajgdls, line 767: Sorry, should this be frame-id ?

wnichols1: no. we are comparing acquisition ids.

ajgdls: Oh I see frame-aid stands for frame acquisition ID.  OK then.

timcnicholls (approved):

wnichols1:
can someone merge this onto master now? I can't.

timcnicholls:
> can someone merge this onto master now? I can't.

Done

PR #245, opened 2020-09-18
removing dependency on blosc

wnichols1:
i) to test this, I ran cmake -DBLOSC_ROOT_DIR=/garbagesdkfjs/prefix and then make, and it built ok.
ii) Note that the enum for compression type is in FrameProcessorDefinitions.h:24, and the compression-type is set in the frame's metadata in the compression plugin. It should still be valid without blosc being installed.

### DIFF
--- a/frameProcessor/src/HDF5File.cpp
+++ b/frameProcessor/src/HDF5File.cpp
@@ -8,7 +8,6 @@
 #include "HDF5File.h"
 
 #include <hdf5_hl.h>
-#include <blosc.h>
 #include "logging.h"
 #include "DebugLevelLogger.h"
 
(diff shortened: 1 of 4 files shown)


frameProcessor/src/HDF5File.cpp
```diff
@@ -8,10 +8,13 @@
 #include "HDF5File.h"
 
 #include <hdf5_hl.h>
-#include <blosc.h>
 #include "logging.h"
 #include "DebugLevelLogger.h"
 
+#ifdef BLOSC_FOUND
```
ajgdls, line 14: Do we need this ifdef if you have verified we do not have a BLOSC dependency?  It would be nice to eliminate the connection between this plugin and BLOSC altogether if there is no real connection.

wnichols1:
I saw this in the top CMakeLists
find_package(Blosc REQUIRED)

frameProcessor/src/HDF5File.cpp
wnichols1: ok I did it a different way now. how is that?

ajgdls (approved):
I like these changes, completely removing the BLOSC dependency from the file writer but keeping a link to ensure the version is correct.

GDYendell (approved):

GDYendell:
Can this be squashed before merging?

skyfrench (approved):

wnichols1:
ok it's one commit - can i go in now?

wnichols1:
it seems some versions of blosc have int blosc_compcode_to_compname(int, const char**), while others want a char**.

timcnicholls (approved):

GDYendell:
I assume @ulrikpedersen is OK with it as he added reviewers, so I think so this is ready

wnichols1:
ya I am updating software on a Desy pc so I have a chance to test it on a system which has no blosc.

ulrikpedersen:
I merged this one as review had happened, everyone approved, but no one had clicked the merge button. 
Thus bringing up my github activity charts a little with minimal efforts 😆 
(hope I didn't screw anything up!)

PR #249, opened 2020-10-07
Fix RX thread startup wait loop under optimised release builds

### LINKED ISSUE #248: Release builds of the frameReceiver don't correctly initialise
Reported by @wnichols1 on PERCIVAL on DLS JIRA ticket [DAQ-3190](https://jira.diamond.ac.uk/browse/DAQ-3190).

frameReceiver builds with the CMake Release configuration (i.e. optimisation level 3) do not correctly start. At the initialisation of the RX thread, the shared buffer precharge mechanism does not appear to operate and the process hangs.

timcnicholls:
The `FrameReceiverRxThread::start()` method has a wait loop which blocks until the thread is running and initialised. When compiled as a CMake release build, i.e. with -O3 optimisation, the loop was being optimised in such a way that the polling in the loop did not detect the thread state change. This has been fixed by modifying the loop to have a bounded timeout and a short sleep each on pass.

Resolves #248

### DIFF
--- a/frameReceiver/include/FrameReceiverRxThread.h
+++ b/frameReceiver/include/FrameReceiverRxThread.h
@@ -78,13 +78,12 @@ private:
   unsigned int           tick_period_ms_;      //!< Receiver thread tick timer period
 
   boost::shared_ptr<boost::thread> rx_thread_; //!< Pointer to RX thread
-  
   IpcChannel             rx_channel_;          //!< Channel for communication with the main thread
   std::vector<int>       recv_sockets_;        //!< List of receive socket file descriptors
 
   bool                   run_thread_;          //!< Flag signalling thread should run
-  bool                   thread_running_;      //!< Flag singalling if thread is running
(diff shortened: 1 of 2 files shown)


ajgdls (approved):

GDYendell:
I had a similar problem with the WatchdogTimer and I fixed it just by making the flag volatile to prevent optimisation breaking things (addcbad). Is this case different such that it needs the extra logic?

timcnicholls:
@GDYendell `volatile` does indeed fix the problem, however it highlighted an unbounded wait loop with no sleep state or timeout, so the additional modifications are just to make that loop a better citizen. 

I found that adding the sleep/timeout without `volatile` also avoided the loop being optimised, but it is better to do both.

GDYendell:
Sounds reasonable! I may copy this logic across. Perhaps there could be a MakeSureWorkerThreadStarts helper class.

GDYendell (approved):

PR #251, opened 2020-10-29
provide FrameDecoder with its own logger

wnichols1:
as agreed in meeting 29/10

### DIFF
--- a/frameReceiver/src/FrameDecoder.cpp
+++ b/frameReceiver/src/FrameDecoder.cpp
@@ -16,27 +16,29 @@ using namespace FrameReceiver;
 //! for base configuration parameters and variables.
 //!
 FrameDecoder::FrameDecoder() :
-     logger_(0),
+     logger_(Logger::getLogger("FR.FrameDecoder")),
      enable_packet_logging_(FrameReceiver::Defaults::default_enable_packet_logging),
      frame_timeout_ms_(FrameReceiver::Defaults::default_frame_timeout_ms),
      frames_timedout_(0),
      frames_dropped_(0)
- {
- };
+{
+};
+
+//! This overload is scheduled for deletion; logger is unused.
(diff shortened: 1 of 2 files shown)


frameReceiver/src/FrameDecoder.cpp
```diff
@@ -16,27 +16,26 @@ using namespace FrameReceiver;
...
 FrameDecoder::FrameDecoder() :
-     logger_(0),
      enable_packet_logging_(FrameReceiver::Defaults::default_enable_packet_logging),
      frame_timeout_ms_(FrameReceiver::Defaults::default_frame_timeout_ms),
      frames_timedout_(0),
      frames_dropped_(0)
- {
- };
+{
+   logger_ = Logger::getLogger("FR.FrameDecoder");
```
timcnicholls, line 24: This should be done in the member initialisation list to be consistent with the other classes, rather than in the body of the constructor.

timcnicholls (changes requested):

frameReceiver/src/FrameDecoder.cpp
wnichols1: ok done

ajgdls:
To smooth the way for the API change, is it worth overloading the init method to already have the version in place without the logger pointer?  I don't know if it is worth doing so I'll approve anyway.

ajgdls (approved):

timcnicholls:
> To smooth the way for the API change, is it worth overloading the init method to already have the version in place without the logger pointer? I don't know if it is worth doing so I'll approve anyway.

That's a good idea @ajgdls then decoder plugins can start to support the new API without a step change in compatibility.

@wnichols1 are you happy to add a further `init` method without the `LoggerPtr` argument?

wnichols1:
ok done.

GDYendell (approved):

timcnicholls (approved):

Issue #256, opened 2020-12-10
FrameProcessor classes should not explicitly set trace logging level


timcnicholls:
Multiple classes in the FrameProcessor explicitly set their own logging level to TRACE. This should not be done programmatically, but left to the control of the logging configuration file.

Reported by @wnichols1

timcnicholls:
Closed by #263

PR #263, opened 2020-12-14
Removed hardcoded log level set calls.

ajgdls:
Resolves issue #256


timcnicholls (approved):
Didn't realise there were so many!

wnichols1:
I said there were 8 at the meeting. In fact there were double that! For simple-minded people like myself counting goes 1,2,3, a lot.

ajgdls:
It was clearly something that was simply copied across into every new file, and we missed that when the refactor took place last year (or the year before).

GDYendell (approved):

PR #264, opened 2020-12-15
WIP Possible RPM implementation.  No dependencies added yet.

ajgdls:
Hello, this isn't finished - I just wanted some opinions on the possible implementation of RPMs using cpack.  I can close this PR and re-open new version if we decide this is the way to go.

Note I haven't yet added dependency checking yet, I wanted to get some thoughts before proceeding too far.

Also as William Nichols is the one who requested this I would like to add him to review but I'm not sure I can do it, is that right?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,17 +59,6 @@ SET(CMAKE_SKIP_BUILD_RPATH  FALSE)
 # When building, don't use the install RPATH already (but later on when installing)
 SET(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE) 
 
-SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
-
-# Add the automatically determined parts of the RPATH which point to directories outside the 
-# build tree to the install RPATH
-SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
-
-# The RPATH to be used when installing, but only if it's not a system directory
-LIST(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/lib" isSystemDir)
-IF("${isSystemDir}" STREQUAL "-1")
-   SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
-ENDIF("${isSystemDir}" STREQUAL "-1")
 
 # Find and add external packages required for application and test
 find_package( Boost 1.41.0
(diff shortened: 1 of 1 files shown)


timcnicholls:
Looks like a sensible approach, especially since it's nicely integrated with CMake. I don't think you can include @wnichols1 as a reviewer, but with him tagged he should see this and be able to comment?

wnichols1:
My experiments seem to show that CPack packages up the binaries from the build directory, and these are different from the ones in the install dir. How do we get it to use the other ones?

ajgdls:
Aren't they only different according to their RPATH settings?

wnichols1:
probably yes, rpath is certainly what we care about. I don't think cpack rewrites the rpath when it runs. So there are two hacks to this problem:
1) set CMAKE_BUILD_RPATH to be appropriate for an installed version
2) tell cpack to package up files it finds in the prefix directory, and ignore the projects. I have got this working - you create an install component like this:
INSTALL(FILES ${CMAKE_INSTALL_PREFIX}/bin/basic
COMPONENT c2
DESTINATION bin
)
and tell cpack to package this component. Then it grabs the install-versions of the files; but it's not pretty.

wnichols1:
For those new to this area, the RPATH is a path specified in an exe telling it where to find the shared libs it needs - we rely on it quite a lot. CMake sets the rpath differently in its build vs its install directories, and uses CMAKE_BUILD_RPATH vs CMAKE_INSTALL_RPATH.
 To install our stuff to /opt/odin/bin and /opt/odin/lib, we will need to set /opt/odin/lib as the rpath in the exes, but we probably don't want this in other versions.
  Currently we put things like /dls_sw/prod/tools in the rpath, but we can't do that any more because it won't work at Desy.

ajgdls:
In the file I committed there was an if clause.  If the RPM directory was set then the executable RPATH cmake commands are not executed.  Can we rely on the build RPATH values as they are relative?  I think for rpm installation relative would be OK if we install all things odin into (eg /opt/odin) ?

wnichols1:
yes, relative rpaths seem to work. However Cmake won't let you set the rpath, only append to the front of it, so it's a bit messy but probably viable.
How about linking statically? I always think that is the simplest. Then there would be 0 shared libraries to worry about.

wnichols1:
I think we need to state WHICH build-systems we are going to make these rpms on, and make sure we have compatibility.
Are we going to build them at github or dls or what?

ajgdls:
Regarding the RPATH (I'll let Ulrik comment on the build system) - I just unpacked the rpm I created using this branch and there is NO RPATH specified for the binaries, which is what I had expected given the updates I made to the file.

PR #265, opened 2021-01-20
End of acquisition

ajgdls:
Currently this just contains the changes to set and act on an EOA flag.  I will look at a control layer implementation for setting up an EOA frame.
This is just for initial review.

### DIFF
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -149,6 +149,14 @@ void FileWriterPlugin::process_frame(boost::shared_ptr<Frame> frame)
       } else if (status == status_invalid) {
         this->set_error("Frame invalid");
         this->set_error(current_acquisition_->get_last_error());
+      } else {
+        // Check to see if this frame is tagged as the last frame in an acquisition
+        if (frame->get_meta_data().get_end_of_acquisition()){
+          LOG4CXX_INFO(logger_, "End of acquisition attached to frame, stopping writer");
+          stop_acquisition();
+          // Prevent the timeout from closing the file as it's just been closed. This will also stop the timer if it's running
+          timeout_active_ = false;
+        }
       }
 
       // Push frame to any registered callbacks
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -141,8 +141,6 @@ void FileWriterPlugin::process_frame(boost::shared_ptr<Frame> frame)
       // Check if this acquisition is complete and stop
       if (status == status_complete) {
         stop_acquisition();
-        // Prevent the timeout from closing the file as it's just been closed. This will also stop the timer if it's running
-        timeout_active_ = false;
       } else if (status == status_complete_missing_frames) {
         LOG4CXX_INFO(logger_, "Starting close file timeout as received last frame but missing some frames");
         start_close_file_timeout();
@@ -170,9 +168,6 @@ void FileWriterPlugin::process_end_of_acquisition()
   if (writing_) {
     LOG4CXX_INFO(logger_, "End of acquisition frame received, stopping writer");
     stop_acquisition();
-    // Prevent the timeout from closing the file as it's just been closed.
-    // This will also stop the timer if it's running.
-    timeout_active_ = false;
   }
 }
 
@@ -880,6 +875,8 @@ void FileWriterPlugin::stop_acquisition() {
       }
     }
   }
+  // Prevent the timeout from closing the file as it's just been closed. This will also stop the timer if it's running
+  timeout_active_ = false;
 }
 
 /**
--- a/frameProcessor/src/FileWriterPlugin.cpp
+++ b/frameProcessor/src/FileWriterPlugin.cpp
@@ -165,7 +165,7 @@ void FileWriterPlugin::process_frame(boost::shared_ptr<Frame> frame)
  * Checks we are writing. If we are in writing mode then the acquisition is
  * stopped and the timeout_active_ flag is set to false.
  */
-void FileWriterPlugin::end_of_acquisition()
+void FileWriterPlugin::process_end_of_acquisition()
 {
   if (writing_) {
     LOG4CXX_INFO(logger_, "End of acquisition frame received, stopping writer");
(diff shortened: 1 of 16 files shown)


timcnicholls:
Looks good to me. Can the EOA frame be empty or does it have to be a valid dataset that the file writer is expecting? I was just wondering about the mechanism in the case e.g. where you want upstream plugins to ensure that have pushed all of their datasets out first before setting this flag.

GDYendell:
It will have to pass this check `if (frame_in_acquisition(frame))` (FileWriterPlugin:136), so I think it has to be an expected frame.

ajgdls:
I could add a case outside of that check to simply look for the tag, but what currently happens if we send an empty frame I wonder, does it error the FP in some way?

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -150,6 +150,14 @@ void FileWriterPlugin::process_frame(boost::shared_ptr<Frame> frame)
       } else if (status == status_invalid) {
         this->set_error("Frame invalid");
         this->set_error(current_acquisition_->get_last_error());
+      } else {
+        // Check to see if this frame is tagged as the last frame in an acquisition
+        if (frame->get_meta_data().get_end_of_acquisition()){
+          LOG4CXX_INFO(logger_, "End of acquisition attached to frame, stopping writer");
+          stop_acquisition();
+          // Prevent the timeout from closing the file as it's just been closed. This will also stop the timer if it's running
+          timeout_active_ = false;
```
GDYendell, line 159: Calling stop and setting timeout_active happens three times now. Could it just be set inside stop_acquisition?

GDYendell:
We agreed in a meeting that it would be better to have a special EOA Frame instead of a flag on a normal Frame, so that the same thing can be done in all cases, but it doesn't necessarily need to be done to merge this first version.

ajgdls:
I've created an EndOfAcquisitionFrame object that contains no data.  The question is how do we process this through the chain?  I think many plugins just expect frames to contain the right kind of stuff, and this frame looks like a frame but with zero size data.

I'm wondering if we should have a new method build into the process chain, so that for this special kind of frame we don't even attempt to process, we simply end up calling the new method.
Basically, outside the control of plugin specific implementation:
1 We check incoming frames
2 If the frame is standard we call process
3 If the frame is end of acquisition we do not call process, instead we call "end_of_acquisition" or similar
4 Plugins have a default "end_of_acquisition" which does nothing
5 Plugins can override end_of_acquisition and push extra frames if they wish or do anything else
6 The plugin will automatically then push end of acquisition frames

We could then also have another method which creates the end_of_acquisition frame and pushes so a plugin can just call that method instead of creating the frame object manually.

Thoughts @GDYendell @timcnicholls ?

GDYendell:
I think that makes sense. If we made a `process_frame(EndOfAcquisitionFrame frame)` overload, would that be called instead of the more generic `process_frame(Frame frame)` if the frame happened to be an `EndOfAcquisitionFrame` (assuming EndOfAcquisitionFrame is a child class of Frame)?

I am failing to search for this because just comes up with results for overloading a normal method in a child class...

timcnicholls:
I like your proposal @ajgdls, I think that makes sense. I favour the `end_of_acquisition` mechanism rather than overloading `process_frame` - explicit is better than implicit here I think.

ajgdls:
I have now made the client updates to allow injecting EOA into the plugin chain.  I think this is ready for a review to see what you think.  @timcnicholls it would be good if you could try out the injection from a client to check it does what you would expect.  @GDYendell if you would review it I would be grateful.

timcnicholls:
> I have now made the client updates to allow injecting EOA into the plugin chain. I think this is ready for a review to see what you think. @timcnicholls it would be good if you could try out the injection from a client to check it does what you would expect. @GDYendell if you would review it I would be grateful.

Will do @ajgdls . I'm tied up with appraisal reviewing most of the rest of this week but I'll look as soon as I can early next.

frameProcessor/src/SharedMemoryController.cpp
```diff
@@ -289,4 +290,19 @@ void SharedMemoryController::status(OdinData::IpcMessage& status)
 
 }
 
+/**
+ * Create an EndOfAcquisitionFrame object and inject it into the plugin chain
+ */
+void SharedMemoryController::injectEOA()
```
GDYendell, line 296: I guess this class was missed when we converted everything to use snake_case

frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -160,6 +160,22 @@ void FileWriterPlugin::process_frame(boost::shared_ptr<Frame> frame)
   }
 }
 
+/** Process an EndOfAcquisitionFrame.
+ *
+ * Checks we are writing. If we are in writing mode then the acquisition is
+ * stopped and the timeout_active_ flag is set to false.
+ */
+void FileWriterPlugin::end_of_acquisition()
```
GDYendell, line 168: Could this be renamed to `process_(eoa|end_of_acquisition)`? `end_of_acquisition` is not an action, so it isn't clear what it does.

frameProcessor/src/SharedMemoryController.cpp
ajgdls: It's a good point.  I'm happy to refactor that class but I don't want to do it as part of this PR, I would prefer to keep this to the point and then open a different issue for the refactor.

frameProcessor/src/FileWriterPlugin.cpp
ajgdls: I sort of prefer "process_eoa" but I think it should probably be "process_end_of_acquisition".  I'll change it to that.

frameProcessor/src/FileWriterPlugin.cpp
GDYendell: This still applies

GDYendell (approved):

timcnicholls (approved):
Christian is still integrating this PR with his HEXITEC plugin base, which is the use case we have for e.g. using EoA to flush histograms out of the pipeline. Since we've sat on this for months, I've also tested it briefly and confirmed that sending the EoA config command does trigger the mechanism correctly, so am happy to approve it. If anything else crops up I suggest we could resolve it through further PRs rather than delaying any further.

PR #267, opened 2021-02-19
Added webp repsponse type to live view to fix bug on Firefox

### LINKED ISSUE #225: Live View does not work in Firefox 68, potentially other versions as well
Live view has stopped working in Firefox, likely since version 65. When requesting an image, it receives a HTTP error code 406 - Not Acceptable.

This is because Firefox has changed the default Accept types in it's headers when requesting an image. The response type produced by the Live View Adapter for the image is `image/*`. Since Firefox 65, the default img accept types are `image/webp` or `*/*`, neither of which now match the response type.

Possible solutions are to support the `image/webp` type, or to find a way to set the accept header for the image so that it does not default to the types set by Firefox.

ANeaves:
Closes #225

### DIFF
--- a/tools/python/odin_data/live_view_adapter.py
+++ b/tools/python/odin_data/live_view_adapter.py
@@ -58,7 +58,7 @@ class LiveViewAdapter(ApiAdapter):
 
         self.live_viewer = LiveViewer(endpoints, default_colormap)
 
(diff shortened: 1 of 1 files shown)


timcnicholls (approved):
Tested with Firefox and recent odin-control - works fine

PR #270, opened 2021-03-18
Clean up frameReceiver and frameProcessor startup error handling and reporting

timcnicholls:
This PR addresses issues in the error handling and logging in the frame receiver and processor startup where the command line arguments are incorrect.

At present, the behaviour of the two applications is inconsistent. The FR doesn't report an error at all, while the frame processor buries the message in a traceback handler as an uncaught exception propagates up the call stack. For instance:
```
$ bin/frameReceiver --config
log4cxx: No appender could be found for logger (FR.App).
log4cxx: Please initialize the log4cxx system properly.
$
```
and
```
$ bin/frameProcessor --config
log4cxx: No appender could be found for logger (FP.App).
log4cxx: Please initialize the log4cxx system properly.
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::program_options::invalid_command_line_syntax> >'
  what():  the required argument for option '--config' is missing
Caught signal 6 (SIGABRT)
stack trace:
bin/frameProcessor(_ZN8OdinData17print_stack_traceEP8_IO_FILEj+0xca)[0x49d6f7]
bin/frameProcessor(_ZN8OdinData13abort_handlerEiP9siginfo_tPv+0xc6)[0x49d825]
/lib64/libpthread.so.0(+0xf630)[0x7f2b86722630]
/lib64/libc.so.6(gsignal+0x37)[0x7f2b856653d7]
/lib64/libc.so.6(abort+0x148)[0x7f2b85666ac8]
/aeg_sw/tools/CentOS7-x86_64/gcc/9-3-0/prefix/lib64/libstdc++.so.6(+0x99286)[0x7f2b85fb1286]
/aeg_sw/tools/CentOS7-x86_64/gcc/9-3-0/prefix/lib64/libstdc++.so.6(+0xa5346)[0x7f2b85fbd346]
/aeg_sw/tools/CentOS7-x86_64/gcc/9-3-0/prefix/lib64/libstdc++.so.6(+0xa53b1)[0x7f2b85fbd3b1]
/aeg_sw/tools/CentOS7-x86_64/gcc/9-3-0/prefix/lib64/libstdc++.so.6(__cxa_rethrow+0x49)[0x7f2b85fbd699]
bin/frameProcessor(main+0x13a1)[0x4a9ea7]
/lib64/libc.so.6(__libc_start_main+0xf5)[0x7f2b85651555]
bin/frameProcessor[0x46f0e9]
$
```
The frame processor also has a logic error where, if an exception is correctly caught during argument parsing, the application continues to attempt to initialise but crashes as the number of iothreads default value has not been set.

To resolve the problems this PR makes a number of changes:

1. In both apps, define a default log4cxx basic configuration as early as possible in the application initialisation to log messages can still be emitted to stdout before the log4cxx config file is parsed and applied.
2. Refactor the FP startup process to be more symmetric with the FR by simplifying `main` to parse arguments and then instantiate a `FrameProcessorApp` object to run the application.
3. If the FP argument parsing fails and doesn't yield a sensible default configuration that would result in a runnable application, quit cleanly with a non-zero exit code.

```
$ bin/frameReceiver --config
4 [0x7f61d2c408c0] ERROR FR.App null - Got exception during command line argument parsing: the required argument for option '--config' is missing
$ echo $?
1
$
```
and
```
$ bin/frameProcessor --config
3 [0x7f8a73ffe8c0] ERROR FP.App null - Got exception during command line argument parsing: the required argument for option '--config' is missing
$ echo $?
1
$
```


frameProcessor/include/FrameProcessorApp.h
```diff
@@ -0,0 +1,50 @@
...
+private:
+
+  LoggerPtr logger_;                    //!< Log4CXX logger instance pointer
+  std::string json_config_file_;        //!< Full path to JSON configuration file
+  static boost::shared_ptr<FrameProcessorController> controller_; //!< FrameProcessor controller object
+  boost::program_options::variables_map vm_; //!< Boost program options variable map
+};
+
+} // namespace FrameProcessor
+#endif /* FRAMEPROCESSOR_APP_H */
```
GDYendell, line 50: No newline

frameProcessor/include/FrameProcessorApp.h
```diff
@@ -0,0 +1,50 @@
...
+
+  void checkNoClientArgs();
+
+  void run();
+
+private:
+
+  LoggerPtr logger_;                    //!< Log4CXX logger instance pointer
+  std::string json_config_file_;        //!< Full path to JSON configuration file
+  static boost::shared_ptr<FrameProcessorController> controller_; //!< FrameProcessor controller object
```
GDYendell, line 45: Why is this static (out of curiosity)?

GDYendell (approved):
All looks very reasonable to me

frameProcessor/include/FrameProcessorApp.h
timcnicholls: Thanks - now fixed

frameProcessor/include/FrameProcessorApp.h
timcnicholls: Ah good spot. It's copied from the FR where it is done so that the interrupt signal handler can invoke the `stop` method on the controller from global scope without an object reference, i.e. https://github.com/odin-detector/odin-data/blob/253d1ed4cf3038d318571ddc096ce2ef3f8cc3eb/frameReceiver/src/FrameReceiverApp.cpp#L418-L429 

Maybe we need to add a handler to the FP and call `shutdown()` on the FP controller? @ajgdls thoughts ??

timcnicholls: OK, I've tried implementing a signal handler for the FP in the same way, where `FrameProcessorApp::stop()` calls `controller_->shutdown()` , which returns OK but the application doesn't quit, presumably as not all threads have been joined. Since the FP terminates cleanly on Ctrl-C without the handler anyway, and the `static` definition is benign, I propose to leave this as it is, unless @ajgdls objects.

ajgdls (approved):
I like it

Issue #271, opened 2021-03-19
Frame receiver no longer sets UDP receive socket buffer size correctly


timcnicholls:
A regression entered the FR code with commit ae5d58e027f01ec20ba7dd7a305b0f4d75d52008, merged with the integration test PR #215 and released with 1.5.0 where the `setsockopt` option name `SO_RCVBUF` was replaced with `SO_REUSEADDR`: 

https://github.com/odin-detector/odin-data/blob/38f84d84be01b167e0909862c45f654f96d35d93/frameReceiver/src/FrameReceiverUDPRxThread.cpp#L51-L64

This prevents the UDP receive thread from setting the receive buffer size on the packet receive socket. This should be reverted.

Reported by @wnichols1

timcnicholls:
Fixed by #272

PR #272, opened 2021-03-19
Restore set/get of SO_RCVBUF socket option in frame receiver UDP RX thread

timcnicholls:
This PR addresses #271 and fixes the regression in setting the receive buffer size on the UDP socket in the receive thread of the frame receiver.

The readback/reporting of the buffer size now also includes a divide by two to account for the kernel allocation behaviour, so that the true buffer size available is reported in matches the requested size.

### DIFF
--- a/frameReceiver/src/FrameReceiverUDPRxThread.cpp
+++ b/frameReceiver/src/FrameReceiverUDPRxThread.cpp
@@ -49,7 +49,7 @@ void FrameReceiverUDPRxThread::run_specific_service(void)
     }
 
     // Set the socket receive buffer size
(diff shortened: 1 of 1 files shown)


ajgdls (approved):

ulrikpedersen (approved):
That is indeed the change we've discussed on slack and in meeting yesterday 👍

GDYendell:
Is it possible to use both of these options?

timcnicholls:
I'm not sure what you mean @GDYendell ?

GDYendell:
Sorry. Is it possible to keep SO_REUSEADDR enabled rather than replacing it?

timcnicholls:
SO_REUSEADDR has no utility on a UDP socket unless you're attempting multicast, to allow multiple processes to bind to the same port. The TIME_WAIT issue only affects closing TCP sockets.

GDYendell:
Ahh. Yes OK, I didn't spot that this is only in the UDP child class.

PR #278, opened 2021-03-26
Migrate odin-data CI tests from travis to Github actions

timcnicholls:
This PR migrates the odin-data CI workflow from travis to Github actions. A simple action workflow, which checks out, builds, installs and tests odin-data is implemented, along with associated fixes (CMake rules, library link order, test job config) required to make the workflow operate on the newer Ubuntu OS provided by the GH CI runners.


ulrikpedersen:
Looks good. All tests are passing so there is a nice green tickbox. What happens when a test fails? Do we get a nice'n easy view of which one?

ulrikpedersen (commented):
You run the tests with `--log_level=all` which produces 1000s of lines of output with the test failed/passed messages somewhere in between. I think that's more like debug output. Is that a concious decision to enable here?

GDYendell:
Can we apt install the kafka libraries so that it can build the plugin? Do we get rid of the travis-install* scripts if we are going to rely on apt install for packages?

timcnicholls:
> You run the tests with `--log_level=all` which produces 1000s of lines of output with the test failed/passed messages somewhere in between. I think that's more like debug output. Is that a concious decision to enable here?

It was just copying what the old Travis job did, but I can easily turn that off.

timcnicholls:
> Can we apt install the kafka libraries so that it can build the plugin? Do we get rid of the travis-install* scripts if we are going to rely on apt install for packages?

Good spot. I will look into that now. Yes, in principle we can delete all of the travis scripts once we're happy.

timcnicholls:
> Looks good. All tests are passing so there is a nice green tickbox. What happens when a test fails? Do we get a nice'n easy view of which one?

Yes, if a step in the workflow fails (e.g. non-zero exit code on unit test job) that is flagged in the workflow output and you can inspect the output. I don't have an example of a single test case failing, but you can see an example of how a step-level failure looks in an earlier run here: https://github.com/odin-detector/odin-data/runs/2167908992?check_suite_focus=true

.github/workflows/test_odin_data.yml
```diff
@@ -0,0 +1,59 @@
...
+          INSTALL_PREFIX: ${{ env.INSTALL_PREFIX }}
+        run: |
+          cd build
+          bin/frameReceiverTest --log_level=all
+      - name: Run frame processor unit tests
+        env:
+          INSTALL_PREFIX: ${{ env.INSTALL_PREFIX }}
+        run: |
+          cd build
+          bin/frameReceiverTest --log_level=all
```
ajgdls, line 44: I think that should say bin/frameProcessorTest

ajgdls (changes requested):

timcnicholls:
OK I've address the issues raised. Thanks for the feedback. Concerning suppressing the logging output from the unit test jobs. That works fine, but it highlights slightly contradictory output from the FP file writer test cases, where expected errors are logged in deliberate test cases  just before the job finishes and says "No errors detected": 

https://github.com/odin-detector/odin-data/runs/2226288858?check_suite_focus=true 

This is correct behaviour - before it was hidden by the reams of other log messages - but is this a potential source of confusion?

ulrikpedersen:
> > Looks good. All tests are passing so there is a nice green tickbox. What happens when a test fails? Do we get a nice'n easy view of which one?
> 
> Yes, if a step in the workflow fails (e.g. non-zero exit code on unit test job) that is flagged in the workflow output and you can inspect the output. I don't have an example of a single test case failing, but you can see an example of how a step-level failure looks in an earlier run here: https://github.com/odin-detector/odin-data/runs/2167908992?check_suite_focus=true

Cool 👍  

I think maybe a nice-to-have future feature would be a nicely formatted, generated unittest and maybe coverage report straight from the github actions. I'm sure there must be some ready-made github actions out there that do that... Of course that's feature-creep and not for this PR! 😃

ulrikpedersen:
> OK I've addressed the issues raised. Thanks for the feedback. Concerning suppressing the logging output from the unit test jobs. That works fine, but it highlights slightly contradictory output from the FP file writer test cases, where expected errors are logged in deliberate test cases just before the job finishes and says "No errors detected":
> 
> https://github.com/odin-detector/odin-data/runs/2226288858?check_suite_focus=true
> 
> This is correct behaviour - before it was hidden by the reams of other log messages - but is this a potential source of confusion?

oh yeah, that is a bit confusing but hard to see how to "fix" as it is correct behaviour - becasue we're testing error cases. If we tweak the output then we'd lose a good trace when a "real" error do happen, right?

timcnicholls:
> EDIT: maybe we could add a log message in the test framework to make it clear that the following trace output is testing an error case and expected?

How about this:
```
Running 32 test cases...
**********************************************************************
*                                                                    *
* *** Starting HDF5File unit tests - ERROR messages are expected *** *
*                                                                    *
**********************************************************************

ERROR - HDF5 Function Error: (Failed to create HDF5 file) in /aeg_sw/work/users/tcn45/develop/projects/odin/odin-data/frameProcessor/src/HDF5File.cpp:163: size_t FrameProcessor::HDF5File::create_file(std::string, size_t, bool, size_t, size_t)
...
<snip>
...
*** No errors detected
```

timcnicholls:
> Perhaps the 'fix' is int the last line? `This warning is for project developers. Use -Wno-dev to suppress it.`

This only seems to happen on the CMake version on the test runners. My Ubuntu machine doesn't do the same. But I've fixed this by renaming `cmake/FindLog4CXX.cmake` to `cmake/FindLOG4CXX.cmake` so that the package name is capitalised consistently.

ulrikpedersen:
> > Perhaps the 'fix' is int the last line? `This warning is for project developers. Use -Wno-dev to suppress it.`
> 
> This only seems to happen on the CMake version on the test runners. My Ubuntu machine doesn't do the same. But I've fixed this by renaming `cmake/FindLog4CXX.cmake` to `cmake/FindLOG4CXX.cmake` so that the package name is capitalised consistently.

Ah OK, I expected this was a complex CMake code change 🤣

GDYendell (approved):

ulrikpedersen (approved):

ajgdls (approved):

PR #279, opened 2021-04-07
Metadata cache and status update

ajgdls:
Created a cache object that grows dynamically for use within the meta writer application.
Added unit test for the object.
Provide status for previously completed acquisitions.

### DIFF
--- a/tools/python/odin_data/meta_writer/hdf5dataset.py
+++ b/tools/python/odin_data/meta_writer/hdf5dataset.py
@@ -40,14 +40,14 @@ class HDF5UnlimitedCache(object):
             upper_index = index + self._block_size
             current_block_size = self._block_size
             if upper_index > self._highest_index:
-                upper_index = self._highest_index
-                current_block_size = upper_index - index + 1
+                upper_index = self._highest_index + 1
+                current_block_size = upper_index - index
             self._logger.debug("[{}] Flushing block {} to index {}:{}".format(self._name, block, index, upper_index))
             self._logger.debug("[{}] Current block size:{}".format(self._name, current_block_size))
             current_block = self._blocks[block]
             self._logger.debug("[{}] Block slice to write:{}".format(self._name, current_block[0:current_block_size]))
-            h5py_dataset.resize(upper_index+1, axis=0)
-            h5py_dataset[index:upper_index+1] = current_block[0:current_block_size]
+            h5py_dataset.resize(upper_index, axis=0)
+            h5py_dataset[index:upper_index] = current_block[0:current_block_size]
 
 
 class HDF5Dataset(object):
--- a/tools/python/odin_data/meta_writer/hdf5dataset.py
+++ b/tools/python/odin_data/meta_writer/hdf5dataset.py
@@ -18,7 +18,11 @@ class HDF5UnlimitedCache(object):
 
     def new_array(self):
         self._logger.debug("[{}] Appending new numpy array of {} to cache in unlimited mode".format(self._name, self._block_size))
-        return np.full(self._block_size, self._fillvalue, self._dtype)
+        data = {
+            'active': True,
+            'data': np.full(self._block_size, self._fillvalue, self._dtype)
+        }
+        return data
 
     def add_value(self, value, offset):
         if offset > self._highest_index:
@@ -27,27 +31,34 @@ class HDF5UnlimitedCache(object):
         if block_index not in self._blocks:
             self._logger.debug("New cache block required")
             self._blocks[block_index] = self.new_array()
-        self._blocks[block_index][value_index] = value
+        self._blocks[block_index]['active'] = True
+        self._blocks[block_index]['data'][value_index] = value
 
     def flush(self, h5py_dataset):
         # Loop over the blocks and write each one to the dataset
         self._logger.debug("[{}] Highest recorded index: {}".format(self._name, self._highest_index))
         index = 0
         dataset_size = len(self._blocks) * self._block_size
+        self._logger.info("[{}] Flushing: total number of blocks to check: {}".format(self._name, len(self._blocks)))
         h5py_dataset.resize(dataset_size, axis=0)
+        active_blocks = 0
         for block in self._blocks:
-            index = block * self._block_size
-            upper_index = index + self._block_size
-            current_block_size = self._block_size
-            if upper_index > self._highest_index:
-                upper_index = self._highest_index + 1
-                current_block_size = upper_index - index
-            self._logger.debug("[{}] Flushing block {} to index {}:{}".format(self._name, block, index, upper_index))
-            self._logger.debug("[{}] Current block size:{}".format(self._name, current_block_size))
-            current_block = self._blocks[block]
-            self._logger.debug("[{}] Block slice to write:{}".format(self._name, current_block[0:current_block_size]))
-            h5py_dataset.resize(upper_index, axis=0)
-            h5py_dataset[index:upper_index] = current_block[0:current_block_size]
+            if self._blocks[block]['active']:
+                active_blocks += 1
+                index = block * self._block_size
+                upper_index = index + self._block_size
+                current_block_size = self._block_size
+                if upper_index > self._highest_index:
+                    upper_index = self._highest_index + 1
+                    current_block_size = upper_index - index
+                self._logger.debug("[{}] Flushing block {} to index {}:{}".format(self._name, block, index, upper_index))
+                self._logger.debug("[{}] Current block size:{}".format(self._name, current_block_size))
+                current_block = self._blocks[block]['data']
+                self._logger.debug("[{}] Block slice to write:{}".format(self._name, current_block[0:current_block_size]))
+                h5py_dataset.resize(upper_index, axis=0)
+                h5py_dataset[index:upper_index] = current_block[0:current_block_size]
+                self._blocks[block]['active'] = False
+        self._logger.info("[{}] Flushing complete - flushed {} active blocks".format(self._name, active_blocks))
 
 
 class HDF5Dataset(object):
--- a/tools/python/odin_data/meta_writer/hdf5dataset.py
+++ b/tools/python/odin_data/meta_writer/hdf5dataset.py
@@ -3,6 +3,53 @@ import logging
 import numpy as np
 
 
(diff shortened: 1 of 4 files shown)


GDYendell:
Do we need to update eiger-detector to work with this?

ajgdls:
No due to the way EigerMetaWriter is constructed I think it should work without change.

tools/python/odin_data/meta_writer/meta_listener.py
```diff
@@ -227,22 +229,39 @@ def status(self, request):
...
 
+    def clear_writer(self, name):
+        """Clear the specified writer"""
+        if name in self._writers:
+            self._logger.info("Removing writer: {}".format(name))
+            writer = self._writers[name]
+            self._status_dict[name] = writer.status()
+            if writer.finished:
+                del self._writers[name]
+
```
GDYendell, line 261: As discussed, we don't think we need this

tools/python/odin_data/meta_writer/meta_listener.py
```diff
@@ -314,8 +333,14 @@ def create_new_writer(self, writer_name):
         # Now create new acquisition
         self._logger.info("Creating new writer %s", writer_name)
         self._writers[writer_name] = self._writer_class(
-            writer_name, DEFAULT_DIRECTORY, self._process_count
+            writer_name, DEFAULT_DIRECTORY, self._process_count, self._data_endpoints
         )
+        self._writers[writer_name].register_owner(self)
```
GDYendell, line 338: As discussed, we don't think we need this

tools/python/odin_data/meta_writer/meta_writer.py
```diff
@@ -115,15 +117,18 @@ def __init__(self, name, directory, process_count):
         self._writers_finished = False
         self._detector_finished = True  # See stop_when_detector_finished
 
+    def register_owner(self, owner):
+        self._owner = owner
+
```
GDYendell, line 122: As discussed, we don't think we need this

GDYendell (commented):
A first pass. I am confused about the status still, so I will have another look at that.

I also think that we should combine HDF5UnlimitedCache with HDF5Dataset, so that we either have no cache and write directly to the dataset, or we use your new cache logic.

tools/python/odin_data/meta_writer/hdf5dataset.py
```diff
@@ -1,12 +1,135 @@
...
+        else:
+            self._shape = (block_size,) + shape
+        self._logger = logging.getLogger("HDF5UnlimitedCache")
+        self._logger.debug(
+            "Creating unlimited HDF5UnlimitedCache for {}".format(self._name)
+        )
+        self._blocks[0] = self.new_array()
+        self._stale_time = stale_time
+
+    def new_array(self):
```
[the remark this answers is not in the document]
ajgdls: Eliminated the method altogether.

tools/python/odin_data/meta_writer/hdf5dataset.py
```diff
@@ -121,55 +242,110 @@ def prepare_data(self, data):
...
-        if self._cache is not None:
-            self._logger.debug(
-                "%s | Writing cache to dataset: %s",
-                self.name,
-                np.array2string(self._cache, threshold=10),
-            )
-            self._h5py_dataset[...] = self._cache
-
-        self._h5py_dataset.flush()
+        if self._h5py_dataset is not None:
```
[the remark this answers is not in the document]
ajgdls: I did see this failing.  We should investigate separately, I will try to reproduce.

tools/python/odin_data/meta_writer/hdf5dataset.py
```diff
@@ -1,12 +1,135 @@
...
+        self._logger.debug(
+            "[{}] Appending new numpy array of {} to cache in unlimited mode".format(
+                self._name, self._block_size
+            )
+        )
+        data = {
+            "active": True,
+            "flush_time": time(),
+            "data": np.full(self._shape, self._fillvalue, self._dtype),
+        }
```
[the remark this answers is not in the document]
ajgdls: It can't be a tuple (as it needs to be changed) so I've created a class.

tools/python/odin_data/meta_writer/hdf5dataset.py
```diff
@@ -1,12 +1,123 @@
...
+            if self._blocks[block].active:
+                active_blocks.append(block)
+                index = block * self._block_size
+                upper_index = index + self._block_size
+                current_block_size = self._block_size
+                if upper_index > self._highest_index:
+                    upper_index = self._highest_index + 1
+                    current_block_size = upper_index - index
+                self._logger.debug(
+                    "[{}] Flush block {} [{}:{}] = {}".format(
```
GDYendell, line 86: Flush -> Flushing

tools/python/odin_data/meta_writer/hdf5dataset.py
```diff
@@ -1,12 +1,123 @@
...
+                self._name, len(self._blocks)
+            )
+        )
+        # Record the number of active blocks
+        active_blocks = []
+        for block in self._blocks:
+            if self._blocks[block].active:
+                active_blocks.append(block)
+                index = block * self._block_size
+                upper_index = index + self._block_size
```
GDYendell, line 80: `block_start` and `block_stop` maybe? To match the slicing syntax

tools/python/odin_data/meta_writer/hdf5dataset.py
```diff
@@ -1,12 +1,123 @@
...
+                        block,
+                        index,
+                        upper_index,
+                        np.array2string(self._blocks[block].data, threshold=10),
+                    )
+                )
+                current_block = self._blocks[block].data
+                if upper_index > h5py_dataset.len():
+                    h5py_dataset.resize(upper_index, axis=0)
+                h5py_dataset[index:upper_index] = current_block[0:current_block_size]
```
GDYendell, line 97: I would split everything from 79-97 into a `write_block_to_dataset(block_index, dataset)` method. I am starting to lose track of what we are doing here.

tools/python/odin_data/meta_writer/hdf5dataset.py
```diff
@@ -1,12 +1,123 @@
...
+            self._blocks[block_index] = HDF5CacheBlock(
+                self._shape, self._fillvalue, self._dtype
+            )
+        self._blocks[block_index].active = True
+        self._blocks[block_index].data[value_index] = value
+
+    def purge_blocks(self):
+        # Store to mark blocks for deletion
+        purged_blocks = []
+        for block_index, block in list(self._blocks.items()):
```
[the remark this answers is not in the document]
ajgdls: You do need it or it will fail in later Python versions I think?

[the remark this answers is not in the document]
ajgdls: "Iterating views while adding or deleting entries in the dictionary may raise a RuntimeError or fail to iterate over all entries."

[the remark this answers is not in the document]
GDYendell: Oh sorry, yes because you are removing items from it! Could you add a comment  e.g. "make a copy to modify the original"

ajgdls:
I've just realised that by removing the code that was allowing a writer to call into the parent to clean itself up after is has been closed, the meta writer has reverted to not allowing the same acquisition ID to be executed twice in a row.
@GDYendell although you thought the clear writer request from into the parent from the writer was messy, it is the only way to guarantee the writer is correctly removed from the store as soon as it is finished (simply leaving behind the status object in case a client is monitoring).  So I need to undo the changes you have requested unless we come up with another way of doing this because not being able to start the same acquisition twice is a show stopper for any manual execution.

tools/python/odin_data/meta_writer/meta_listener.py
ajgdls: @GDYendell , I've just been testing and I think we do either need this or some other mechanism to ensure a writer is cleaned out as soon as it is finished.  It should simply leave behind its final status dict.

ajgdls:
I think if we call clear_writers after every meta message that will also pick up as soon as the writer is finished.  This removes the need for the writer to know about the listener class and doesn't add too much overhead.
I'll try this before reverting any of the changes.

ajgdls:
OK @GDYendell pushed the updates...

GDYendell:
Looks good! I tried running against the eiger simulator and it works fine. I set the block size to 3 so I could see it working:
```
[D 210510 18:08:18 hdf5dataset:244] offset | Writing cache to dataset
[D 210510 18:08:18 hdf5dataset:68] [offset] Highest recorded index: 9
[D 210510 18:08:18 hdf5dataset:72] [offset] Flushing: total number of blocks to check: 4
[D 210510 18:08:18 hdf5dataset:103] [offset] Writing block 0 [0:3] = [0 1 2]
[D 210510 18:08:18 hdf5dataset:103] [offset] Writing block 1 [3:6] = [3 4 5]
[D 210510 18:08:18 hdf5dataset:103] [offset] Writing block 2 [6:9] = [6 7 8]
[D 210510 18:08:18 hdf5dataset:103] [offset] Writing block 3 [9:10] = [ 9 -1 -1]
[D 210510 18:08:18 hdf5dataset:86] [offset] Flushing complete - flushed 4 blocks: [0, 1, 2, 3]
```

ajgdls:
OK great, so I think it's ready now, I've just updated the unit test to work with the recent changes.  Any time you want to approve it I'm happy.  If @timcnicholls has no comments I will probably remove from the approval list.

ajgdls:
The failure is in the slack notification.

ajgdls:
Any ideas @GDYendell or @timcnicholls 

[ERROR] Secret `SLACK_WEBHOOK` is missing. Please add it to this action for proper execution.
Refer https://github.com/rtCamp/action-slack-notify for more information.

ajgdls:
Oh I see, OK I can do that.  I guess I'll close this one down and re-open from odin-detector.

GDYendell (approved):

ajgdls:
OK @timcnicholls I'll merge this one this time and then from now on I'll create any PRs directly from branches on odin-detector.

PR #283, opened 2021-06-14
Meta writer fixes

GDYendell:
I have pre-emptively rebased on end-of-acquisition because there were conflicts to resolve. Alan's commits will disappear from this when #265 is merged and I rebase this on master.

This contains two main changes, plus some other tidying and error handling:
- Add a workaround for receiving messages for a frame from the file writer before the matching message from the detector process plugin
- Add the option to define static datasets by defining the shape in the `HDF5Dataset` definition
  - Add a command line argument --sensor-shape so that the EigerMetaWriter can define the shape of some of its datasets at creation time, so that they are SWMR-readable


tools/python/odin_data/meta_writer/hdf5dataset.py
```diff
@@ -214,7 +215,19 @@ def write(self, data):
...
+        elif self.shape != data.shape:
+            # If the dataset cannot be resized, then the data shape must match
+            self._logger.error(
+                "%s | Dataset shape %s does not match data shape %s",
+                self.name,
+                self.shape,
+                data.shape,
+            )
+            return
+
```
ajgdls, line 230: Is this always going to be true?  Could you have a smaller shape that fits inside the dataset - this wouldn't need a resize but the sizes wouldn't match.

GDYendell: In that case I don't think you would want to define the exact initial shape, you would just make it resizable. Is that not true for some case?

ajgdls: I guess that is true.  I was just commenting on the change in functionality that has been implemented - I have no use cases either way as I use a cache for all of the datasets I'm interested in.

ajgdls (approved):

PR #292, opened 2022-01-25
Add an option for setting HDF5 file postfix.

ajgdls:
For a case where data is stored into file with geographical separation instead of round robin of frames it makes sense for the file names to be labelled with slightly different names and all retain the same numbering (e.g.):
file_A_000001.h5
file_B_000001.h5
This change allows the specification of a file postfix (_A or _B in the example above) placed before the automatic numbering.

### DIFF
--- a/frameProcessor/src/Acquisition.cpp
+++ b/frameProcessor/src/Acquisition.cpp
@@ -46,6 +46,8 @@ Acquisition::Acquisition(const HDF5ErrorDefinition_t& hdf5_error_definition) :
         frames_processed_(0),
         total_frames_(0),
         frames_to_write_(0),
+        starting_file_index_(0),
+        use_file_numbers_(true),
         use_earliest_hdf5_(false),
         alignment_threshold_(1),
         alignment_value_(1),
@@ -328,6 +330,9 @@ void Acquisition::validate_dataset_definition(DatasetDefinition definition) {
  * \param[in] concurrent_processes - The number of processors
  * \param[in] frames_per_block - The number of frames per block
  * \param[in] blocks_per_file - The number of blocks per file
+ * \param[in] starting_file_index - The first number in the indexing of filenames
+ * \param[in] use_file_numbers - Use file numbers in the file name (or not)
+ * \param[in] file_postfix - An additional string placed before any numbering
  * \param[in] file_extension - The file extension to use
  * \param[in] use_earliest_hdf5 - Whether to use an early version of hdf5 library
  * \param[in] alignment_threshold - Alignment threshold for hdf5 chunking
@@ -340,6 +345,8 @@ bool Acquisition::start_acquisition(
     size_t concurrent_processes,
     size_t frames_per_block,
     size_t blocks_per_file,
+    uint32_t starting_file_index,
+    bool use_file_numbers,
     std::string file_postfix,
     std::string file_extension,
     bool use_earliest_hdf5,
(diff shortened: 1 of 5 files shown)


timcnicholls (changes requested):
No issues with the PR, although the `Acquisition::generate_filename()` method docstring could benefit from updating to mention the postfix also. 

I'l also like to suggest that we follow up on the long-standing request to make the fixed format file numbering optional too 😄

ajgdls:
This would be a good time to make that change!  Is it a boolean flag that you would like to be able to set (turn on or off numbering)?  Also I remember someone (maybe William) suggesting the file numbering should start from a configured number so we could have the files starting with index 0 instead of 1.  Shall I stick that in too?

timcnicholls:
> This would be a good time to make that change! Is it a boolean flag that you would like to be able to set (turn on or off numbering)? Also I remember someone (maybe William) suggesting the file numbering should start from a configured number so we could have the files starting with index 0 instead of 1. Shall I stick that in too?

I think having both ia boolean enable and a start index would be perfect thanks. We can make them default to the current behaviour (enabled/starting at 1) so it doesn't surprise anyone.

ajgdls:
There is now a starting index configuration parameter and a numbering on/off configuration parameter.  The defaults are for indexing to be 0 indexed (which really is the most used indexing) instead of 1 indexed, and for the numbers to be turned on by default.

ajgdls:
Yay finally I believe the tests are all passing, and I have implemented the additional config items for setting the starting file index and turning off file numbering.  @timcnicholls @GDYendell  I think this is now ready for review again.

timcnicholls (approved):

frameProcessor/src/Acquisition.cpp
```diff
@@ -639,16 +654,23 @@ std::string Acquisition::get_meta_header() {
...
   {
-    generated_filename << configured_filename_ << "_" << number_string << file_extension_;
+    generated_filename << configured_filename_ << file_postfix_;
   }
   else if (!acquisition_id_.empty())
   {
-    generated_filename << acquisition_id_ << "_" << number_string << file_extension_;
+    generated_filename << acquisition_id_ << file_postfix_;
+  }
+  if (!generated_filename.str().empty()){
```
GDYendell, line 669: Would it be better to have an else with a default name if the `configured_filename` and `acquisition_id` are both empty and then get rid of this check? I think this method should always return `<something>.h5`.

ajgdls: I haven't changed the behaviour, the method specifically mentions returning an empty string if filename and acqID are not defined.
Later in the code (in start_acquisition method) the assigned variable from this return value is checked for empty string so I think this is doing what it was designed to do.

GDYendell: Hmm, OK. Maybe it is how it is used in `get_file` that is wrong then: https://github.com/odin-detector/odin-data/blob/master/frameProcessor/src/Acquisition.cpp#L554

ajgdls: @GDYendell , in the start_acquisition method if the filename is empty then the writer never starts running and so this `get_file` method would never be called I think:

```
  // Generate the filename
  filename_ = generate_filename(concurrent_rank_);

  if (filename_.empty()) {
    last_error_ = "Unable to start writing - no filename to write to";
    LOG4CXX_ERROR(logger_, last_error_);
    return false;
  }

```
Although it isn't the most elegant code it is functional and I don't think I want to change that behaviour as part of this PR.  Are you satisfied with that?  We could raise an issue to mention that this needs cleaning up?

GDYendell: Sounds good to me

ajgdls: OK then I'll merge this and raise an issue :+1:

PR #293, opened 2022-03-16
Fix bug in live view plugin

ajgdls:
The time_last_frame_ variable was not initialised when the plugin was loaded, which could result in no frames being sent out if the FP was immediately set with a frame_frequency of 0 and a non-zero per_second value.


timcnicholls (changes requested):
I have this same fix on a branch I'm currently working on 👍 

There was a change in the default initialisation policy for `boost::posix_time` at some point which caused this to appear. You have to be careful with the initial value though, as it depends on the granularity of the clock. There is a special constant `min_date_time` which is guaranteed to be less than any clock time value. I used it in the constructor member initialisation list:

```
LiveViewPlugin::LiveViewPlugin() :
    publish_socket_(ZMQ_PUB),
    is_bound_(false),
    time_last_frame_(boost::posix_time::min_date_time)
{
```

timcnicholls (approved):

PR #294, opened 2022-03-16
Allow targeting of frames for specific plugin chains

ajgdls:
Reason:
Xspress detector batches spectra into blocks of 256.  This single "frame" is written to HDF5.  However, the individual incoming MCAs need to be pushed to the live view plugin as they arrive.  We do not want to push them to the FileWriter plugin because the FP will drop into an error state as these individual frames are not expected.

Solution:
Overload the push method and allow a plugin to target a chain for a specific frame or frames.  This could be useful also for sending differing types of frames to different processing plugin chains.

### DIFF
--- a/frameProcessor/include/FrameProcessorPlugin.h
+++ b/frameProcessor/include/FrameProcessorPlugin.h
@@ -54,6 +54,7 @@ public:
 
 protected:
   void push(boost::shared_ptr<Frame> frame);
+  void push(const std::string& plugin_name, boost::shared_ptr<Frame> frame);
 
 private:
   /** Pointer to logger */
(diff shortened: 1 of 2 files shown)


timcnicholls (commented):
This seems like a great idea.

How is the selected plugin chain related to the plugin naming and chain configuration at run time, where the naming in the config passed to the FP is arbitrary. Is this coupling the run-time configuration to a compile-time name, i.e. `plusin_name`??

timcnicholls:
> @timcnicholls do you think there is anything further we should do to enforce this or are you happy with the current open implementation?

I think it's fine as is. The point you made in discussing it at the meeting was entirely valid - there are plenty of other examples in the FP where we couple configuration naming to code.

GDYendell (approved):

PR #297, opened 2022-05-16
Implement frame objects containing multiple datasets in outer chunk dimension

ajgdls:
This PR provides a mechanism for writing frame objects that contain multiple dataset entries in the outer chunking dimension of the frame.
The specific case that requires this is for the Xspress detector which contains arrays of 4096 elements.  In the current codebase there are two options:
1) Write out chunks of 1x4096 elements
2) Create sudo frames of 256x4096 elements to write reasonable chunks but the resulting dataset would be <n>x256x4096
Neither of these options are appropriate for these datasets.

This PR allows a frame to have set its outer chunk dimension and therefore each frame object can contain multiple individual datasets (in our case 256x4096).  As these frames are written the dataset size jumps by the number of the outer chunking dimension which is a correct representation of the data.

The specific example for Xspress is below:
Previously as data is recorded the dataset would grow as follows:
1x256x4096
2x256x4096
3x256x4096 ...

With this change the dataset grows as follows:
256x4096
512x4096
768x4096 ...

The default behaviour of the frame objects is to have an outer chunk dimension of 1 which is the current implied hardcoded implementation.

### DIFF
--- a/frameProcessor/src/Acquisition.cpp
+++ b/frameProcessor/src/Acquisition.cpp
@@ -150,7 +150,7 @@ ProcessFrameStatus Acquisition::process_frame(boost::shared_ptr<Frame> frame, HD
       // are true then increment the number of frames written.
       if (master_frame_.empty() || master_frame_ == frame_dataset_name) {
         size_t dataset_frames = current_file_->get_dataset_frames(frame_dataset_name);
-        frames_processed_++;
+        frames_processed_ += frame->get_outer_chunk_size();
         LOG4CXX_TRACE(logger_, "Master frame processed");
         size_t current_file_index = current_file_->get_file_index() / concurrent_processes_;
         size_t frames_written_to_previous_files = current_file_index * frames_per_block_ * blocks_per_file_;
--- a/frameProcessor/src/HDF5File.cpp
+++ b/frameProcessor/src/HDF5File.cpp
@@ -262,9 +262,10 @@ void HDF5File::write_frame(
   }
 #endif
 
-  // Set actual_dataset_size_ to the offset we just wrote to + 1 (Note: frame_offset is zero-indexed)
-  if (frame_offset + 1 > dset.actual_dataset_size_) {
-    dset.actual_dataset_size_ = frame_offset + 1;
+  // Check if the latest written frame has extended the dataset, and if it has then
+  // adjust the actual_dataset_size_ member of the dset structure to match the real size
+  if (offset[0] + frame.get_outer_chunk_size() > dset.actual_dataset_size_) {
+    dset.actual_dataset_size_ = offset[0] + frame.get_outer_chunk_size();
   }
 }
 
--- a/frameProcessor/include/Frame.h
+++ b/frameProcessor/include/Frame.h
@@ -69,6 +69,12 @@ class Frame {
   /** Set the image offset */
   void set_image_offset(const int &offset);
 
+  /** Set the outer chunk size */
+  void set_outer_chunk_size(const int &size);
+
+  /** Set the outer chunk size */
+  int get_outer_chunk_size() const;
+
  protected:
 
   /** Pointer to logger */
@@ -85,6 +91,9 @@ class Frame {
 
   /** Offset from frame memory start to image data */
   int image_offset_;
+
+  /** Outer chunk size of this frame (number of images in this chunk) */
+  int outer_chunk_size_;
 
 };
 
--- a/frameProcessor/src/BloscPlugin.cpp
+++ b/frameProcessor/src/BloscPlugin.cpp
@@ -149,6 +149,7 @@ boost::shared_ptr<Frame> BloscPlugin::compress_frame(boost::shared_ptr<Frame> sr
 
 
   dest_frame->set_image_size(compressed_size);
+  dest_frame->set_outer_chunk_size(src_frame->get_outer_chunk_size());
 
   return dest_frame;
 }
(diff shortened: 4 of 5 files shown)


GDYendell:
Could you link where xspress uses this to set the outer chunk to something other than 1?

ajgdls:
https://gitlab.diamond.ac.uk/controls/tools/xspress-detector/-/blob/asyncio/data/frameProcessor/src/XspressProcessPlugin.cpp#L435

It's internal to DLS I'm afraid

frameProcessor/src/Acquisition.cpp
```diff
@@ -150,7 +150,11 @@ ProcessFrameStatus Acquisition::process_frame(boost::shared_ptr<Frame> frame, HD
       // are true then increment the number of frames written.
       if (master_frame_.empty() || master_frame_ == frame_dataset_name) {
         size_t dataset_frames = current_file_->get_dataset_frames(frame_dataset_name);
-        frames_processed_++;
+        if (frame->get_outer_chunk_size() < outer_chunk_dimension){
+          frames_processed_+=frame->get_outer_chunk_size();
+        } else {
+          frames_processed_+=outer_chunk_dimension;
+        }
```
GDYendell, line 157: Why would `frame->get_outer_chunk_size()` be different to `outer_chunk_dimension`? Isn't this always just hitting the first case? Which is correct, I think.

ajgdls: Lets say you have a dataset with 1000 arrays.  The first three would be 256 (outer_chunk_dimension) and the last would be  232 so to avoid forcing the dataset to 1024 and inserting zeros you need to allow for cases where those two are different.

ajgdls: In the case where get_outer_chunk_size() was set larger you cannot write more than a single chunk.  That is actually not desired behaviour but to ensure the frames_processed_ is correct it would have to be allowed for I think.

GDYendell: Isn't `frame->get_outer_chunk_size() < outer_chunk_dimension` an error case?

ajgdls: I think currently it clips, but yes I think it should be an error.  I'll take a look at where that should be checked.

ajgdls: Actually I think my last statement was incorrect, almost certainly the write would fail if the size was greater than a chunk.  It's not explicit in our code but HDF write chunk would complain.  So I think we could make this simply `frames_processed_+=frame->get_outer_chunk_size();` for every call.

GDYendell: If it can be caught and handled, I think this code should just always do `frames_processed_ += frame->get_outer_chunk_size();`

ajgdls:
Somehow my messages were part of a review so they weren't sent, but yes I agreed with you @GDYendell , I'll push the change and I was wondering if you would have time to verify against the Eiger use case of multiple files with block sizes.

frameProcessor/src/HDF5File.cpp
```diff
@@ -262,9 +262,15 @@ void HDF5File::write_frame(
...
-    dset.actual_dataset_size_ = frame_offset + 1;
+  // Set actual_dataset_size_ to the offset we just wrote to += the outer_chunk_size (Note: frame_offset is zero-indexed)
+  if (frame.get_outer_chunk_size() < outer_chunk_dimension){
+    if (offset[0] + frame.get_outer_chunk_size() > dset.actual_dataset_size_) {
+      dset.actual_dataset_size_ = offset[0] + frame.get_outer_chunk_size();
+    }
+  } else {
+    if (offset[0] + outer_chunk_dimension > dset.actual_dataset_size_) {
+      dset.actual_dataset_size_ = offset[0] + outer_chunk_dimension;
+    }
```
GDYendell, line 273: This can have the same simplification as above re: `frame.get_outer_chunk_size() < outer_chunk_dimension` always being true.

GDYendell: It could be a little confusing what this is doing now. It is not just adding 1 because it is converting 0-index to 1-index. It is effectively shifting the size by an offset to find the highest frame written to. I think it could do with some explanation.

frameProcessor/src/Acquisition.cpp
```diff
@@ -150,7 +150,7 @@ ProcessFrameStatus Acquisition::process_frame(boost::shared_ptr<Frame> frame, HD
       // are true then increment the number of frames written.
       if (master_frame_.empty() || master_frame_ == frame_dataset_name) {
         size_t dataset_frames = current_file_->get_dataset_frames(frame_dataset_name);
-        frames_processed_++;
+        frames_processed_+=frame->get_outer_chunk_size();
```
GDYendell, line 153:
```suggestion
        frames_processed_ += frame->get_outer_chunk_size();
```

GDYendell:
I think this looks good. Could you squash these commits into one and I will try it on the Eiger sim.

timcnicholls (approved):
I'm happy with this when you're both satisfied, as it's not my core expertise 😉 

We have an application (HEXITEC) writing histogram datasets and I wonder if this will benefit that.

GDYendell:
I have tested this against the eiger sim and the datasets look the same.

GDYendell:
For some reason the PR test failed and the push test succeeded. I re-ran the PR build and it worked the second time...

ajgdls:
This PR has been particularly unhappy when I first push, with tests failing and then re-running the tests and they pass.  I can't believe the changes I made would affect the tests in this way because the default is to be exactly the same as the existing system.  If the change broke the tests then they would stay broken.  I'm not really sure what the cause of the failure is at the moment.

GDYendell (approved):

PR #298, opened 2022-05-18
Meta improvements

ajgdls:
This PR does the following:

- adds the ability to set a maxshape for a meta data dataset
- stores sockets as a dictionary indexed by endpoint instead of assuming rank 0,1,2,3 etc
- adds a compatibility change for python3

### DIFF
--- a/tools/python/odin_data/meta_writer/hdf5dataset.py
+++ b/tools/python/odin_data/meta_writer/hdf5dataset.py
@@ -20,7 +20,7 @@ class HDF5UnlimitedCache(object):
         self._fillvalue = fillvalue
         self._block_size = block_size
         self._blocks = {}
-        self._highest_index = 0
+        self._highest_index = -1
         if shape is None:
             self._shape = block_size
         else:
@@ -32,7 +32,9 @@ class HDF5UnlimitedCache(object):
             "Created unlimited HDF5UnlimitedCache for {}".format(self._name)
         )
 
-    def add_value(self, value, offset):
+    def add_value(self, value, offset=None):
+        if offset is None:
+            offset = self._highest_index + 1
         if offset > self._highest_index:
             self._highest_index = offset
         block_index, value_index = divmod(offset, self._block_size)
--- a/tools/python/odin_data/meta_writer/hdf5dataset.py
+++ b/tools/python/odin_data/meta_writer/hdf5dataset.py
@@ -14,23 +14,34 @@ class HDF5CacheBlock(object):
 class HDF5UnlimitedCache(object):
     """ An object to represent a cache used by an HDF5Dataset """
 
-    def __init__(self, name, dtype, fillvalue, block_size, stale_time, shape=None):
+    def __init__(self, name, dtype, fillvalue, block_size, data_shape, block_timeout):
+        """
+        Args:
+            name(str): Name of underlying dataset
+            dtype(str) Datatype of cache blocks
+            fillvalue(value matching dtype): Fill value of cache blocks
+            block_size(int): Number of data elements within a cache block
+            data_shape(tuple(int)): Shape of individual data elements in a cache block
+            block_timeout(int): Timeout in seconds for a block
+                If no new values are added within the timeout the block will be marked
+                for deletion.
+
+        """
         self._name = name
         self._dtype = dtype
         self._fillvalue = fillvalue
         self._block_size = block_size
         self._blocks = {}
         self._highest_index = -1
-        if shape is None:
-            self._shape = block_size
-        else:
-            self._shape = (block_size,) + shape
-        self._blocks[0] = HDF5CacheBlock(self._shape, self._fillvalue, self._dtype)
-        self._stale_time = stale_time
-        self._logger = logging.getLogger("HDF5UnlimitedCache")
-        self._logger.debug(
-            "Created unlimited HDF5UnlimitedCache for {}".format(self._name)
+        self._block_shape = (block_size,) + data_shape
+        self._block_timeout = block_timeout
+
+        self._blocks[0] = HDF5CacheBlock(
+            self._block_shape, self._fillvalue, self._dtype
         )
+
+        self._logger = logging.getLogger("HDF5UnlimitedCache")
+        self._logger.debug("Created cache for {}".format(self._name))
 
     def add_value(self, value, offset=None):
         if offset is None:
@@ -41,7 +52,7 @@ class HDF5UnlimitedCache(object):
         if block_index not in self._blocks:
             self._logger.debug("New cache block required")
             self._blocks[block_index] = HDF5CacheBlock(
-                self._shape, self._fillvalue, self._dtype
+                self._block_shape, self._fillvalue, self._dtype
             )
         self._blocks[block_index].has_new_data = True
         self._blocks[block_index].data[value_index] = value
@@ -54,7 +65,7 @@ class HDF5UnlimitedCache(object):
             if not block.has_new_data:
                 # If the block has no new data check if it has become stale and delete
                 time_since_flush = time() - block.flush_time
-                if time_since_flush > self._stale_time:
+                if time_since_flush > self._block_timeout:
                     purged_blocks.append(block_index)
                     del self._blocks[block_index]
         self._logger.debug(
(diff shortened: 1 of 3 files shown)


tools/python/odin_data/meta_writer/meta_listener.py
```diff
@@ -70,18 +71,19 @@ def run(self):
...
-            data_sockets.append(socket)
+            socket.setsockopt(zmq.SUBSCRIBE, cast_bytes(""))
+            data_sockets[endpoint] = socket
+
 
         poller = zmq.Poller()
         poller.register(ctrl_socket, zmq.POLLIN)
         for socket in data_sockets:
-            poller.register(socket, zmq.POLLIN)
+            poller.register(data_sockets[socket], zmq.POLLIN)
```
GDYendell, line 86:
```suggestion
        for socket in data_sockets.values():
            poller.register(socket, zmq.POLLIN)
```

tools/python/odin_data/meta_writer/meta_listener.py
```diff
@@ -98,8 +100,8 @@ def run(self):
                     continue
 
                 for socket in data_sockets:
-                    if sockets.get(socket) == zmq.POLLIN:
-                        self.handle_data_message(socket)
+                    if sockets.get(data_sockets[socket]) == zmq.POLLIN:
+                        self.handle_data_message(data_sockets[socket], socket)
```
GDYendell, line 104: Same here

tools/python/odin_data/meta_writer/hdf5dataset.py
```diff
@@ -20,11 +20,12 @@ def __init__(self, name, dtype, fillvalue, block_size, stale_time, shape=None):
         self._fillvalue = fillvalue
         self._block_size = block_size
         self._blocks = {}
-        self._highest_index = 0
+        self._highest_index = -1
```
GDYendell, line 23: This seems like an important change. Was this broken before?

tools/python/odin_data/meta_writer/hdf5dataset.py
```diff
@@ -20,11 +20,12 @@ def __init__(self, name, dtype, fillvalue, block_size, stale_time, shape=None):
...
         if shape is None:
             self._shape = block_size
         else:
-            self._shape = (block_size,) + shape
+            self._shape = (block_size,) + shape[1:]
```
GDYendell, line 27: Could you explain this?

tools/python/odin_data/meta_writer/meta_writer.py
```diff
@@ -201,11 +202,18 @@ def _create_datasets(self, dataset_size):
         self._logger.debug("%s | Creating datasets", self._name)
 
         for dataset in self._datasets.values():
+            chunks = None
+            try:
+                if None not in dataset.maxshape:
+                    chunks = dataset.maxshape
+            except:
+                if dataset.maxshape is not None:
+                    chunks = dataset.maxshape
```
GDYendell, line 211:
```suggestion
            chunks = dataset.maxshape
            if isinstance(chunks, int):
                chunks = (chunks,)
            if None in chunks:
                chunks = None
```

tools/python/odin_data/meta_writer/meta_writer.py
```diff
@@ -465,20 +473,20 @@ def handle_start_acquisition(self, header, _data):
         """Prepare the data file with the number of frames to write"""
         self._logger.debug("%s | Handling start acquisition message", self._name)
 
-        if self._processes_running[header[RANK]]:
+        if self._processes_running[self._endpoints.index(header["_endpoint"])]:
```
GDYendell, line 476: Could we have `ENDPOINT = "_endpoint"` and use it everywhere?

tools/python/odin_data/meta_writer/meta_listener.py
```diff
@@ -270,6 +275,8 @@ def clear_writers(self):
                 if stagnant:
                     self._logger.info("Stopping stagnant writer %s", writer_name)
                     writer.stop()
+                unfinished_writers[writer_name] = writer
+        self._writers = unfinished_writers
```
GDYendell, line 279: This is nicer :+1:

tools/python/odin_data/meta_writer/meta_listener.py
```diff
@@ -142,6 +144,9 @@ def handle_data_message(self, socket):
         header = socket.recv_json()
         self._logger.debug("Header message:\n%s", header)
 
+        # Set the endpoint of the message into the header
```
GDYendell, line 147: Could this be extended to explain that it is just to tell the writer where the data came from? And maybe an example of why it matters?

tools/python/odin_data/meta_writer/meta_listener.py
GDYendell: As it is, `socket` is just an index in a list

tools/python/odin_data/meta_writer/hdf5dataset.py
ajgdls: In certain situations this lead to an initial (zero filled) entry in the dataset.  It only showed up when the special case of Xspress datasets (n x 9) were introduced.  Previously I do not think detectors used expanding 2D datasets in this way.

tools/python/odin_data/meta_writer/meta_listener.py
ajgdls: I will update

tools/python/odin_data/meta_writer/meta_writer.py
ajgdls: @GDYendell can you explain this change?  The formatting looks out, if you only remove line 211.  Do you also want 210 removed?

tools/python/odin_data/meta_writer/meta_writer.py
ajgdls: Good idea

tools/python/odin_data/meta_writer/hdf5dataset.py
GDYendell: This currently assumes if `shape` is not given then the dataset is an array of scalars and if it is then the dataset is an array of ND data (which, I think, is what you are making this change to support), but in eiger I have individual 2D images: 

`Float32HDF5Dataset(FLATFIELD, shape=(4000, 2000), rank=2, cache=False)`

so this will now make the `block_size` (1000000, 2000).

We need to support all of these cases in a sane way. I could retain the behaviour I want by passing only the block_size as the sensor shape, but that doesn't really make sense.

tools/python/odin_data/meta_writer/meta_writer.py
GDYendell, line 211:
```suggestion
            chunks = dataset.maxshape
            if isinstance(chunks, int):
                chunks = (chunks,)
            if None in chunks:
                chunks = None
```

Sorry I should have made it a multi-line suggestion

GDYendell: Instead of catching the error when trying to iterate a scalar, we can just make it a tuple and pass that. `chunks=1` is equivalent to `chunks=(1,)`

tools/python/odin_data/meta_writer/hdf5dataset.py
ajgdls: Mm, why would you use an unlimited cache for an individual 2D image, wouldn't you simply use no cache at all?
For your case, wouldn't you actually want to pass

ajgdls: Oh actually you aren't using a cache, you have cache=False.

GDYendell: Riight OK. So for single arrays, you have to set cache=False or it could do strange things. I think that is fine for now, but the API could be made clearer (later).

tools/python/odin_data/meta_writer/hdf5dataset.py
```diff
@@ -33,6 +34,8 @@ def __init__(self, name, dtype, fillvalue, block_size, stale_time, shape=None):
         )
 
     def add_value(self, value, offset):
+        if offset is None:
```
GDYendell, line 37: I think `offset` should default to `None` if this is added. Are you explicitly calling it with `None` somewhere currently?

ajgdls: I'll update.  It is only ever called in the same file `self._cache.add_value(value, offset)` but it doesn't hurt to have the default in case code is changed in the future.

GDYendell (approved):

PR #300, opened 2022-05-27
Fix elapsed time calculation in live view rate settiing.

timcnicholls:
The elapsed time calcuation relies on truncating a posix_time duration to milliseconds as an int_32, which can fail with an incorrect value on the initial frame, or if the frame processor has been idling for long periods. To resolve this, the time since last frame variable is now stored and compared as a time_duration.

Also tidied up the live view decision logic so that the elapsed time is only calculated if the per_second rate setting is active


ajgdls (approved):
I've tested this on my system and it is working as expected.

PR #301, opened 2022-06-15
Update packaging

GDYendell:
- Move code into cpp and python directories
- Move python packaging config to setup.cfg
- Split h5py dependency into a `meta_writer` extra so that control adapters can be used without it


GDYendell:
@timcnicholls I have split into `cpp` and `python` subdirectories as discussed.

I have also rebased #303 on this to make sure that still works.

timcnicholls:
@GDYendell I think this is much cleaner 👍 . Did you try the same layout on a detector repo? Does it map to that OK?

GDYendell:
@timcnicholls I have added an equivalent commit here: https://github.com/dls-controls/excalibur-detector/pull/45

I am not sure I like `tools/python`, especially when there are not other `tools/` directories - perhaps `tools/scripts` would be better?

The top level README and python README could also both do with updating.

timcnicholls:
Thanks. Looks like it maps onto detector repos OK too.

> I am not sure I like `tools/python`, especially when there are not other `tools/` directories - perhaps `tools/scripts` would be better?
I think `tools/scripts` is fine - it's a collection of ad-hoc (and probably legacy) things that aren't a deployable artefact, so categorising them by language seems overkill.

GDYendell:
Updated excalibur-detector PR to `tools/scripts`

ajgdls:
I have updated Xspress against this version and it appears to work well.  It simplified the Xspress module a lot by allowing both control and data C++ applications to be built under a single cmake install.

ajgdls (approved):

GDYendell:
I have built all of our detectors against this branch and they all build and run (without a real detector to talk to).

I think this is ready for a final look when you get a change @timcnicholls 

Note the build is failing because of something fixed in #303. I have rebased that branch on this one to check the [build](https://github.com/odin-detector/odin-data/actions/runs/3998233987/jobs/6860603981).

timcnicholls (approved):

PR #304, opened 2023-01-24
Stats Publish

GDYendell:
These are two somewhat isolated changes to allow odin to complete a control loop between a motion controller moving some attenuators and the exposure on a detector. These are:
- Extend SumPlugin to calculate a histogram as well as total counts
- Create ParameterPublishPlugin to publish any parameters on a frame as json over ZeroMQ

I imagine these two features could be useful for other things.

- [x] Waiting on #301

### DIFF
--- a/cpp/common/include/Json.h
+++ b/cpp/common/include/Json.h
@@ -34,6 +34,8 @@ public:
 
   void add(const std::string& key, const std::string&  json_value);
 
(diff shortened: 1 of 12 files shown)


GDYendell:
This is ready for review. It is pretty noninvasive. The behavior SumPlugin is intended be the same as before unless histogram bins thresholds are configured (excluding some additional checks and iterating of empty containers).

PR #305, opened 2023-01-24
Xspress detector developments

GDYendell:
A few additions coming out of adding support for the Xspress detector


GDYendell:
@ajgdls are you happy to drop [0a0c0e6](https://github.com/odin-detector/odin-data/pull/305/commits/0a0c0e6cc7c48f4cec297fda612088ce3bd2cad5) from these changes because #300 fixes the problem?

ajgdls:
> @ajgdls are you happy to drop [01027f8](https://github.com/odin-detector/odin-data/pull/305/commits/01027f804cf910d864f7913d4a07372a868f0091) from these changes because #300 fixes the problem?

Yes agreed this can be dropped.

timcnicholls:
> Check for and update minimum pyzmq version for cast_{bytes,unicode}.
> 
> pyzmq 23.0 adds deprecation warnings for strtypes, so maybe these casts are made redundant by other changes. Investigate and see if we can remove these completely before updating to >=23.0.

@GDYendell  I'm looking into this now. While testing this branch with a fresh venv and `pyzmq` install it no longer seems possible to do a `pip install -e .`, as `setup.cfg` now expresses a dependency on `odin-control`, which doesn't resolve to a published PyPI package. It didn't use to be necessary to have `odin-control` for this package - is it now? If so, the dependency needs updating to use a Github URL rather than a simple package name.

GDYendell:
Ah sorry about that @timcnicholls. I missed that because it works at DLS with our internal PyPI mirror. 

We can add the github URL to setup.cfg - where was this defined before? It isn't in the old setup.py, setup.cfg or requirements*.txt.

GDYendell:
See #307

timcnicholls:
> We can add the github URL to setup.cfg - where was this defined before? It isn't in the old setup.py, setup.cfg or requirements*.txt.

I _think_ it relied on manual installation of `odin-control` before installing?

timcnicholls:
OK, I've investigated this a bit more, using `pyzmq` version 25 although it's applicable for >23 . The deprecation of the `strtypes` `cast_xxx` methods is not because the translation from strings to bytes is automatic, rather that these are not used internally any more. The `send_string` and `recv_string` convenience methods on the ZMQ socket class use internal serialize/deserlalize helpers. There are also no equivalent multipart send/recv methods for strings.

I think the conclusion is that we need to reimplement/retain our own `cast_xxx` helper functions in `IpcChannel`, but this is worth a discussion first. Generally we only use `IpcChannel` to send JSON objects as strings, so could switch to using the ZMQ socket `send_string` and `recv_string` (or even the JSON versions) but there are cases where the channels are used with raw data (e.g. in live view).

GDYendell:
We should probably mirror the pyzmq library and have `send` and `send_str`, but that would be a breaking change and cause exceptions in existing client code that calls `send` with `str`. As a transition we could update `send` to remove the casting and instead just type check and call `send` / `send_str` as appropriate - then at least `ipc_*_channel` won't be using the deprecated API.

I don't see an obvious way of updating `recv` without breaking client code, though, because we would have to make a choice as to whether to return `str` or `bytes`. It looks like the live view data currently gets cast to `str`, which it then passes to `np.fromstring` which seems unnecessary.

I also think adding type hints will make this much clearer.

GDYendell:
@timcnicholls I have updated this to remove usage of zmq strtypes and reimplement our own versions.

timcnicholls (approved):

PR #307, opened 2023-02-22
Add git URL for odin-control requirement

GDYendell:
- [x] Do we want to specify a tag?

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -1,0 +1,18 @@
+name: Python CI
+on: [push, pull_request]
+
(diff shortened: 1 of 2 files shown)


timcnicholls:
> [ ] Do we want to specify a tag?

Yes, otherwise it will track the head of master?

GDYendell:
@timcnicholls I have added a pin to odin-control 1.3.0 and a CI build to verify pip install works.

GDYendell:
- [ ] Check this doesn't clobber an existing editable odin-control

PR #308, opened 2023-03-10
Updated and relocate client app

timcnicholls:
This PR updates the legacy `frame_receiver_client` script, which is used for debug and development purposes, to use the correct `IpcMessage` format with message ids, relocating the source in the `odin_data` package and updating the script entry point name to `odin_data_client`.

### DIFF
--- a/python/src/odin_data/client.py
+++ b/python/src/odin_data/client.py
@@ -66,7 +66,7 @@ class OdinDataClient(object):
         parser.add_argument('--config', type=argparse.FileType('r'), dest='config_file', nargs='?',
                             default=None, const=sys.stdin,
                             help='Specify JSON configuration file to send as configure command')
-        parser.add_argument('--request_config', action='store_true',
+        parser.add_argument('--request-config', action='store_true',
                             help='Request configuration from odin-data application')
         parser.add_argument('--status', action='store_true',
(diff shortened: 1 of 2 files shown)


python/src/odin_data/client.py
```diff
@@ -1,113 +1,127 @@
...
-        parser = argparse.ArgumentParser(prog=prog_name, description='ODIN Frame Receiver Client')
+
+        parser = argparse.ArgumentParser(prog=prog_name, description='odin-data application client')
         parser.add_argument('--ctrl', type=str, default='tcp://127.0.0.1:5000', dest='ctrl_endpoint',
                             help='Specify the IPC control channel endpoint URL')
         parser.add_argument('--config', type=argparse.FileType('r'), dest='config_file', nargs='?',
-                            default=None, const=sys.stdin, 
+                            default=None, const=sys.stdin,
                             help='Specify JSON configuration file to send as configure command')
+        parser.add_argument('--request_config', action='store_true',
```
GDYendell, line 56: Could be have dashes?

```suggestion
        parser.add_argument('--request-config', action='store_true',
```

GDYendell:
This works for the meta_writer and eiger_fan.

There are some flake8 and isort errors, but we should agree on standards and add some CI for linting before worrying about it.

PR #311, opened 2023-05-05
Fix build on newer compilers

GDYendell:
Without these changes it fails with gcc 11 (used in current CI build) and 12 (on my machine), for example.


GDYendell:
Example failing build: https://github.com/odin-detector/odin-data/actions/runs/4892576823/jobs/8734495786

timcnicholls (approved):

Issue #302, opened 2022-09-22
frameProcessor startup config causing runtime_error is not caught


GDYendell:
When a config message is received from a client, any `runtime_error`s are caught and handled without shutting down the application, whereas the config messages processed before `run()` are not caught.

It is currently unclear whether and when plugins should throw `runtime_error` to propagate errors up to the controller and when they will be caught and handled correctly.

Consider whether this is the correct behavior and document the difference between the two cases.

PR #312, opened 2023-05-05
Tidy frameProcessor command line interface

### LINKED ISSUE #187: Consider rationalising configuration mechanisms for odin-data apps
Currently we have four methods of configuring the frameReceiver and frameProcessor:
1. Command line arguments
2. INI-style configuration files
3. JSON configuration files
4. JSON configuration messages sent over the control channel

None of these are capable of completely configuring the application, for instance:
- the logging configuration can only be specified at the command line or INI file
- the command line and INI file are not capable of expressing the imperative syntax needed to send successive configuration 'commands' to the FP to construct a plugin chain
- rebinding control IpcChannel endpoints is handled inconsistently across the default value, INI config file and JSON configuration, leading to binding errors at startup in the FR.

**Proposal for discussion**: drop the INI config files completely, leaving command-line arguments JSON config file and messages.

GDYendell:
- [x] #310

This could fix #187 if the frameReceiver is updated as well.


GDYendell:
Remove non-ctrl endpoints as well. Leave iothreads as this is passed in the initialiser list. Check wildcard vs 0.0.0.0 behavious on endpoints. Change to 127.0.0.1 by default and move constants to common to share in fp and fr.

ctrl, log_config, level and json config will be left.

GDYendell:
Possibly resolve #302 here as well. FR already does the good thing and shuts down cleanly.

GDYendell:
Superceded by #317

PR #314, opened 2023-05-05
Add Docs

### LINKED ISSUE #218: Add Documentation
Do some tidying and add documentation to make odin-data easier to pick up and use.

- Remove ini file configuration
- Remove command line options that were added for testing before we had JSON configuration
- Add some documentation on configuration of processes and plugins
- Add structure to documentation

Outstanding points to Discuss:

- Add equivalent documentation to detector repositories
- Remove all the example ini files in config/
- Replace them all with JSON equivalents

------

WIP Branch here: https://github.com/odin-detector/odin-data/tree/documentation

------

Moved from #217 while in progress.

GDYendell:
- [x] #318 
- [x] Top commit to be removed once tested
- [ ] ~~Update `odin_data/client.py` and [docs](https://odin-detector.github.io/odin-data/tutorials/getting_started/control.html) on using client~~ Will make separate PR

Fixes #218


GDYendell:
I think this is ready now. It is definitely not complete, but I think having this framework in place so that it is easy to add documentation as we go along will be really useful.

I have removed the use of the client app and will make a separate PR. The old `frame_receiver_client` / new `odin_data_client` is quite different from the API I had in my version and wrote the docs for, so I would like to make some changes and that requires further discussion.

GDYendell:
The docs are visible [here](https://odin-detector.github.io/odin-data/docs/index.html). Note the `docs` in the URL is not static, it is just the branch name. So, when merged there will be a https://odin-detector.github.io/odin-data/master/index.html, which https://odin-detector.github.io/odin-data will redirect to. There is a dropdown selector for each active branch so the docs can be tested before merging PRs.

timcnicholls (approved):

PR #316, opened 2023-05-16
Add version-dependent mechanism to avoid boost global placeholder deprecation

timcnicholls:
This PR addresses the issue with deprecation warnings during build that boost placeholders should not be imported into the global namespace. 

The suggested fix for this, i.e. `#include <boost/bind/bind.hpp>` and `using namespace boost::placeholders`, is not compatible with the full range of OS/Boost/Cmake versions targeted by `odin-data`. This is because the placeholders namespace did not appear in boost until 1.60.0 (EL7 is 1.53.0), and that the use of placeholders in `boost::property_tree` didn't get updated until 1.76.0 (q.v. Ubuntu 22.04, EL9 variants have boost 1.75.0). Furthermore the expansion of the Boost version in the CMake finder varies in behaviour. The following changes are implemented:

* CMake version independent checking of boost version and setting a compiler define flag `BOOST_HAS_PLACEHOLDERS` when the namespace is present (v >= 1.60.0)
* Use of that flag where necessary in the code
* Set the `BOOST_BIND_GLOBAL_PLACEHOLDERS` flag in the integration test build where `boost::property_tree` is used

Clean compilation has been checked on EL7, Ubuntu 22.04 and AlmaLinux 9.


GDYendell (approved):
I have tested building on a RHEL7 machine and it works. The CI won't work until #313 in any case.

timcnicholls:
Done.  PR CI job failed (again) though the push job passed.

PR #317, opened 2023-07-07
Tidy Application CLI interfaces

### LINKED ISSUE #187: Consider rationalising configuration mechanisms for odin-data apps

### LINKED ISSUE #302: frameProcessor startup config causing runtime_error is not caught

GDYendell:
Fixes #187 
Fixes #302


GDYendell:
I think this is ready for review @timcnicholls

ajgdls (approved):

PR #328, opened 2023-11-15
FP Adapter ParameterTree

ajgdls:
This PR is the beginning of a full ParameterTree implementation within the odin-data adapters.  The following logic has been updated:

- The IpcTornadoClient class has been updated to provide an API that allows waiting for messages to respond, and checking for rejected messages.
- The FrameProcessorAdapter class makes use of the new feeback API in the IpcTornadoClient to ensure rejected configurations are reported.  The mechanism to write files has been upgraded to support a ParameterTree implementation

This PR does **not** address the larger issue of how to support a ParameterTree for an application (FP) that can have dynamically loaded plugins (there is still no ParameterTree in the underlying OdinDataAdapter class).

### DIFF
--- a/python/src/odin_data/control/frame_processor_adapter.py
+++ b/python/src/odin_data/control/frame_processor_adapter.py
@@ -5,16 +5,21 @@ Created on 6th September 2017
 """
 import logging
 import os
+import threading
+import queue
+from tornado.escape import json_decode
 
 from odin.adapters.adapter import (
     ApiAdapterRequest,
     ApiAdapterResponse,
     request_types,
     response_types,
+    wants_metadata
 )
 from tornado import escape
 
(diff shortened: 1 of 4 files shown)


python/src/odin_data/control/frame_processor_adapter.py
```diff
@@ -5,16 +5,21 @@
...
     request_types,
     response_types,
+    wants_metadata
 )
 from tornado import escape
 
 from odin_data.control.odin_data_adapter import OdinDataAdapter
+from odin.adapters.parameter_tree import ParameterAccessor, ParameterTree
 
 FP_ADAPTER_KEY = 'fr_adapter_name'
```
wnichols1, line 24: can you delete bool_from_string()?

PR #330, opened 2023-12-07
Close leaky hdf5 references

GDYendell:
IIRC the thing that was causing (noticeable) memory leaks was the dataset handles. The other things are just bits I noticed when having a check through at the same time.


timcnicholls (approved):

ajgdls (approved):

wnichols1:
good work Gary, you have fixed several issues including the one I originally raised.

PR #331, opened 2024-01-10
fix frequency of frames in LiveViewPlugin

wnichols1:
This fix is straightforward. I don't know why it is necessary to count the number of frames going out of the live-view when the frame-number is available. This is relevant when you are pushing several datasets per frame.

### DIFF
--- a/cpp/frameProcessor/src/LiveViewPlugin.cpp
+++ b/cpp/frameProcessor/src/LiveViewPlugin.cpp
@@ -6,6 +6,7 @@
  */
 
 #include "LiveViewPlugin.h"
+#include "DebugLevelLogger.h"
 #include "version.h"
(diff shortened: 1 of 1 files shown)


GDYendell:
Discussed in the core meeting and we don't think it makes sense to go based on frame number because it doesn't work when there are multiple processes and each node will get blocks of 100 frames, or only odd frames, etc.. This is the reason it uses a simple counter, because it makes the logic local to the specific process rank without having to know how many other nodes there are.

The point about being able to publish the same frame number of multiple datasets is a good one. Currently it is possible to configure a list of datasets to publish, so it could map that list of datasets to a list of counters specific to each dataset, so that it is possible to send the Nth frame of every configured dataset. I will make an issue for this.

PR #338, opened 2024-03-07
Update to use latest odin-control

GDYendell:


timcnicholls:
> @timcnicholls do you think it would be sensible to remove the version and fetch the latest from GitHub? The reason I added the URL dependency was mainly so that the CI worked without odin-control being on PyPI. It is probably just as sensible to test the latest odin-control.

I think so. For a deployment we'd probably pin specific versions of both through explicit requirements anyway. If we introduce regressions into odin-control PRs that break something for odin-data we'd be better finding that out sooner through CI failing.

GDYendell:
OK I will just remove the specific version here instead of changing it

PR #339, opened 2024-03-07
Add TCP Receiver

jsouter:
Adds in the TCP rx thread required for merlin-detector.

### DIFF
--- a/cpp/frameReceiver/src/FrameReceiverTCPRxThread.cpp
+++ b/cpp/frameReceiver/src/FrameReceiverTCPRxThread.cpp
@@ -9,49 +9,48 @@
 
 using namespace FrameReceiver;
 
-FrameReceiverTCPRxThread::FrameReceiverTCPRxThread(FrameReceiverConfig& config,
-                                                   SharedBufferManagerPtr buffer_manager,
-                                                   FrameDecoderPtr frame_decoder,
-                                                   unsigned int tick_period_ms) :
-    FrameReceiverRxThread(config, buffer_manager, frame_decoder, tick_period_ms),
-    logger_(log4cxx::Logger::getLogger("FR.TCPRxThread")),
-    recv_socket_(-1)
-{
-  LOG4CXX_DEBUG_LEVEL(1, logger_, "FrameReceiverTCPRxThread constructor entered....");
+FrameReceiverTCPRxThread::FrameReceiverTCPRxThread(
+    FrameReceiverConfig &config, SharedBufferManagerPtr buffer_manager,
+    FrameDecoderPtr frame_decoder, unsigned int tick_period_ms)
+    : FrameReceiverRxThread(config, buffer_manager, frame_decoder,
+                            tick_period_ms),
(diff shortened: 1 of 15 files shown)


cpp/frameReceiver/src/FrameReceiverTCPRxThread.cpp
```diff
@@ -0,0 +1,110 @@
...
+      return;
+    }
+
+    // Read it back and display
+    int buffer_size;
+    socklen_t len = sizeof(buffer_size);
+    getsockopt(recv_socket, SOL_SOCKET, SO_REUSEADDR, &buffer_size, &len);
+    LOG4CXX_DEBUG_LEVEL(1, logger_, "RX thread receive buffer size for port " << rx_port << " is " << buffer_size);
+
+    // Bind the socket to the specified port
```
timcnicholls, line 64: This line (and the error message at line 82) refer to binding the socket - should both be changed to reflect fact that socket is connected (line 79)?

cpp/frameReceiver/src/FrameReceiverTCPRxThread.cpp
```diff
@@ -0,0 +1,110 @@
...
+      return;
+    }
+
+    // Register this socket
+    this->register_socket(recv_socket, boost::bind(&FrameReceiverTCPRxThread::handle_receive_socket, this, recv_socket, (int)rx_port));
+  }
+}
+
+void FrameReceiverTCPRxThread::cleanup_specific_service(void)
+{
```
timcnicholls, line 93: I think the socket be removed from the reactor and closed in the cleanup method to avoid dangling TCP connections in TIMED_WAIT states? See the ZMQ  RX thread implementation for an example.

timcnicholls (changes requested):
Looks good! A few minor comments. It would be great if there was a dummy decoder plugin and integration test added for TCP receivers too 😉

jsouter:
Cheers Tim, I'll get to work on those changes

jsouter:
Having some trouble REGISTERing a second Dummy decoder, think it's an issue with the CMake in the FR test directory which has to link the APP_SOURCES cpp files to the executable, and the REGISTER macro doesn't work when it is called twice within a single translation unit(?)

jsouter:
Looking at this again, have made the minor changes to error messages mentioned above and added in a remove_socket call in the TCPRxThread's cleanup_specific_service. Had some issues with getting the build to work when registering a new Dummy decoder class but now I have a DummyTCPFrameDecoder somewhat resembling the Merlin frame decoder. Will set about working on the actual unit tests soon.

jsouter:
Okay, I removed the *Lib.cpp files from the glob in the CMakeLists.txt as discussed and removed the namespacing of the registrations. I will note that because of recent changes to master using Lambda functions, we may need to add in a set(CMAKE_CXX_STANDARD 11) call somewhere in the CMake, which is absent from this PR.

timcnicholls (approved):

ajgdls (approved):

GDYendell:
Thanks @jsouter!

PR #351, opened 2024-07-23
Add the ability to specify attributes on HDF5Datasets

DominicOram:
See https://github.com/DiamondLightSource/ADOdin/issues/22

To test:
* Run unit tests and confirm they pass


GDYendell (approved):
This looks great. Only nitpicks.

DominicOram:
Thanks for the review, comments addressed

python/src/odin_data/meta_writer/hdf5dataset.py
```diff
@@ -1,10 +1,15 @@
...
+from typing import Optional
 
 import h5py as h5
 import numpy as np
 
 
+def units(unit: str):
+    return {"units": unit}
+
+
```
ajgdls, line 12: Where is this used?

DominicOram: In the detector specific code at https://github.com/DiamondLightSource/eiger-detector/pull/24/files. The use of `units` as an attribute name seemed useful enough that it should live in the core writer. Happy to move it into the `eiger` if you'd rather though

ajgdls (approved):

PR #353, opened 2024-08-02
Docs update

GDYendell:


GDYendell:
Confirmed working in docs: https://odin-detector.github.io/odin-data/docs-update/index.html

Issue #354, opened 2024-08-08
Crash in IpcMessage::set_param when requestConfiguration called on plugin without name set


GDYendell:
The function assumes a certain length when checking for array notation [here](https://github.com/odin-detector/odin-data/blob/master/cpp/common/include/IpcMessage.h#L231).

Ohisemega:
Hi @GDYendell, I noticed this while working on making the unit test for issue #204
I wonder if a default name should be appended. The issue is that the default name parameter is an empty std::string.

PR #355, opened 2024-08-08
Allow setting hdf plugin frames to zero

### LINKED ISSUE #306: odin-data FileWriterPlugin does not accept 0 frames as a config setting.
When submitting a configuration to the FileWriter plugin for odin-data FP applications, there is a check that the number of frames (FileWriterPlugin::CONFIG_FRAMES) requested is greater than 0.

However, event based detectors require that the number of frames is set to 0.  If the number of frames is mistakenly set to something greater than 0 in this case, it can then no longer be set back to zero.  This is also the case for frame detectors which are run in continuous mode.

GDYendell:
This ensures that the hdf frame count can be changed from 0 and then changed back to 0 again for free-run mode.

Fixes #306 

Note: Added #354 during this work. This is why the call to `set_name` is required.


ajgdls (approved):
This all looks good.  I cannot believe we had this bug for so long without it being noticed before!

PR #359, opened 2024-09-13
Implemented command and requestCommands methods in the FrameProcessor.

ajgdls:
This PR adds the core code for supporting commands withinFrameProcessor plugins.
No commands are implemented within this PR.


timcnicholls (changes requested):
Thanks @ajgdls for doing this. I have a couple of comments/musings:

- the new `MsgVal` enum entries should also be added to [`IpcMessage::msg_val_map_init()`](https://github.com/odin-detector/odin-data/blob/991eb75a9b625e05b2d5bd25a81a78f35662ab25/cpp/common/src/IpcMessage.cpp#L599) in `IpcMessage.cpp` to allow JSON encoding
- in retrospect, now having a `MsgType` of `MsgTypeCmd` and a `MsgVal` of `MsgValCmdCommand` now looks a bit odd. We can't change `MsgTypeCmd` easily without breaking existing code, but I wonder if the naming of the new value could be tweaked to separate them a bit? Maybe something like `MsgTypeCmdExecute`? I'm not sure this is a huge issue, other than avoiding the need for a historical explanation 😉
- at some point we should add enumeration of the types and values to `IpcMessage` in the python client...

GDYendell:
> Maybe something like MsgTypeCmdExecute?

:+1:

GDYendell (changes requested):
Could you add a command the Dummy process plugin and then add a test the calls command() and request_commands()?

cpp/frameProcessor/src/DummyUDPProcessPlugin.cpp
```diff
@@ -135,6 +138,73 @@ namespace FrameProcessor
...
+   * This method responds to command executions requests from client, populating the supplied IpcMessage
+   * reply with the commands and command parameters supported by this plugin.
+   * 
+   * \param[in,out] reply - IpcMessage response to client, to be populated with plugin commands and parameters
+   */
+  void DummyUDPProcessPlugin::requestCommands(OdinData::IpcMessage& reply)
+  {
+    const std::string command_str = get_name() + "/" + DummyUDPProcessPlugin::EXECUTE_PRINT + "/" + DummyUDPProcessPlugin::EXECUTE_PRINT_NAME;
+    const std::string val_str = "";
+    reply.set_param(command_str, val_str);
```
GDYendell, line 205: When there are parameters the value will be a dictionary, so I think it should return an empty dictionary for no parameters.

cpp/frameProcessor/test/DummyUDPProcessPluginTest.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    // Request the command set of the DummyUDPProcessPlugin
+    BOOST_REQUIRE_NO_THROW(dummy_plugin.requestCommands(command_reply));
+
+    OdinData::IpcMessage sub_reply(command_reply.get_param<const rapidjson::Value&>(std::string("dummy")),
+                                   command_reply.get_msg_type(),
+                                   command_reply.get_msg_val());
+
+    // Verify that the command set includes the command to execute a print statement
+    BOOST_CHECK_EQUAL(sub_reply.has_param(std::string("print")), true);
```
GDYendell, line 43:
```suggestion
    BOOST_CHECK(sub_reply.has_param(std::string("print")));
```
?

cpp/frameProcessor/test/DummyUDPProcessPluginTest.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    OdinData::IpcMessage param_reply(sub_reply.get_param<const rapidjson::Value&>(std::string("print")),
+                                     sub_reply.get_msg_type(),
+                                     sub_reply.get_msg_val());
+
+    // Verify that the command has the parameter name, which names the configuration item to print
+    BOOST_CHECK_EQUAL(param_reply.has_param(std::string("name")), true);
```
GDYendell, line 50: and here

cpp/frameProcessor/src/DummyUDPProcessPlugin.cpp
GDYendell: Wait, this command does take parameters, so I am confused how this works

ajgdls: The alternative is to return the type as some sort of enum (string set of types etc).  I'm not sure which I like better.  More imprtantly, how do we make this clear to a developer what is expected?  The code doesn't place any restriction on what is returned by definition and so we are simply doing this by convention.

GDYendell: Yeah it would just have to be documented as a convention.

I think inventing enum of types across python and C++ code bases is potentially not worth the overhead. Maybe there is some standard API we can copy?

ajgdls: I'm not sure of any standard API.  I have added comments directly into this method for this specific case.  I will add to the user guide documentation as part of this PR - if we agree on this convention (or if there is a better way).  @timcnicholls any thoughts on this ?

ajgdls:
Hi @GDYendell @timcnicholls I have pushed the changes resulting from our conversation yesterday.

cpp/frameProcessor/test/DummyUDPProcessPluginTest.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+    FrameProcessor::DummyUDPProcessPlugin dummy_plugin;
+};
+
+BOOST_FIXTURE_TEST_SUITE(DummyUDPProcessPluginUnitTest, DummyUDPProcessPluginTestFixture);
+
+BOOST_AUTO_TEST_CASE( DummyUDPProcessPlugin_commands )
+{
+    std::vector<std::string> commands_reply;
+    std::vector<std::string> expected_reply = {"print"};
```
GDYendell, line 32: This isn't used?

cpp/frameProcessor/src/FrameProcessorPlugin.cpp
```diff
@@ -168,6 +168,33 @@ void FrameProcessorPlugin::requestConfiguration(OdinData::IpcMessage& reply)
...
+ *
+ * In this abstract class the command method does perform any
+ * actions, this should be overridden by subclasses.
+ *
+ * \param[in] command - String containing the command to execute.
+ * \param[out] reply - Response IpcMessage.
+ */
+void FrameProcessorPlugin::execute(const std::string& command, OdinData::IpcMessage& reply)
+{
+  // Default method simply does nothing
```
GDYendell, line 181: I think if we get here, no command will be run but it won't reply with nack. Maybe the "Submitted command not supported" should be returned here and plugins should call into the parent if they don't find a command? It is conceivable that we will add a command here that applies to all plugins.

ajgdls: I agree this is probably an error case.  I think if we get here then the resposne should be "This plugin does not support any commands" or similar.  If in the future we add a command supported by all plugins we could remove this reply, but I think we should avoid letting plugins call into the parent (because a developer might forget to do it).  So I would recommend plugin wide commands are handled just outside this mechanism if we need to.

GDYendell: That makes sense to me.

cpp/frameProcessor/src/DummyUDPProcessPlugin.cpp
GDYendell: We have decided that commands will not take parameters, so this is moot.

cpp/frameProcessor/test/DummyUDPProcessPluginTest.cpp
ajgdls: Sorry I completely missed this.  Pushed change.

ajgdls:
OK I think this is ready for review again.  I have also tested sending commands with a python test client to a running FrameProcessor application.

cpp/frameProcessor/include/DummyUDPProcessPlugin.h
```diff
@@ -54,6 +56,11 @@ class DummyUDPProcessPlugin : public FrameProcessorPlugin
   /** Configuraiton constant for copy frame mode **/
   static const std::string CONFIG_COPY_FRAME;
 
+  /** Command execution constant for print command **/
+  static const std::string EXECUTE_PRINT;
+  /** Command execution parameter constant for print command **/
+  static const std::string EXECUTE_PRINT_NAME;
```
timcnicholls, line 62: Is `EXECUTE_PRINT_NAME ` now redundant with the removal of command parameters?

cpp/frameProcessor/src/DummyUDPProcessPlugin.cpp
```diff
@@ -15,6 +15,9 @@ namespace FrameProcessor
   const std::string DummyUDPProcessPlugin::CONFIG_IMAGE_WIDTH = "width";
   const std::string DummyUDPProcessPlugin::CONFIG_IMAGE_HEIGHT = "height";
   const std::string DummyUDPProcessPlugin::CONFIG_COPY_FRAME = "copy_frame";
+  // Command and parameters
+  const std::string DummyUDPProcessPlugin::EXECUTE_PRINT = "print";
+  const std::string DummyUDPProcessPlugin::EXECUTE_PRINT_NAME = "name";
```
timcnicholls, line 20: As above

timcnicholls (approved):
Looks good! Is `EXECUTE_PRINT_NAME` now redundant?

cpp/frameProcessor/src/DummyUDPProcessPlugin.cpp
ajgdls: Yes! Removed.

GDYendell (approved):

Issue #346, opened 2024-06-10
Implement ParameterTree for FR adapter


GDYendell:

GDYendell:
This should just work with the new generic OdinDataAdapter. Check this.

GDYendell:
@jsouter did you do this while working on https://github.com/DiamondLightSource/fastcs-odin/pull/33?

GDYendell:
That is pretty much all I did for the meta writer so that may be all that is required.

jsouter:
Here is the diff of a branch I made that makes OdinDataController generic, and makes subclasses for FrameProcessorController and FrameReceiverController (which is actually an alias for the base OdinDataController), and FrameReceiver(/Processor)Adapter classes that use these controllers. 

This would mean we could still use the existing odin_server.cfg's module paths without having to specify a frame_processor flag. 
If we'd rather keep fastcs-dev as it is, I can make a PR to just delete frame_receiver_adapter.py as OdinDataAdapter makes it redundant.

https://github.com/odin-detector/odin-data/compare/fastcs-dev...jsouter:odin-data:controllers

GDYendell:
I think I would like to see that diff as PR to fastcs-dev for discussion.

Do we understand why the fr check is commented out, currently? @ajgdls?

jsouter:
Okay, PR here https://github.com/odin-detector/odin-data/pull/362/files

ajgdls:
The check was not implemented during development of the new version of the class.  What is commented out there is a stub which needs an implementation to check on availability of buffers from the FR.

ajgdls:
Oh sorry, for some context... In the previous version of the frame processor adapter, the FP adapter would connect to and check the number of free buffers according to the FR adapter.  It would test this against a defined threshold and refuse to send the command to start writing if the check failed.
The idea was to give prior warning if the system was being starved of buffers rather than allowing it to start and then potentially fail quickly.

GDYendell:
Completed in #362. Added #364 for further work.

PR #362, opened 2024-10-07
Create FR and FP controllers inheriting from OdinController

jsouter:
Following discussion from https://github.com/odin-detector/odin-data/issues/346

Removes all logic relating to the frame processor flag from OdinDataController, and implements FR and FP specific logic in subclasses. If we do want to use the same controller for both the FP and FR adapter I have tested that it works okay with FastCS and it would be sufficient to delete the redundant (and now broken) frame_receiver_adapter.py

Also a small fix that renames a parameter in the Dummy decoders to class instead of name (as signals called "name" and "description" get excluded when the full ParameterTree is requested and can not be added to FastCS).


python/src/odin_data/control/odin_data_adapter.py
```diff
@@ -85,7 +83,7 @@ def get(self, path, request):
         #    logging.error("{}".format(response))
         except ParameterTreeError as param_error:
             response = {
-                "response": "OdinDatatAdapter GET error: {}".format(param_error)
+                "response": "OdinDataAdapter GET error: {}".format(param_error)
```
jsouter, line 86: Realising this should probably get the class name instead if we're subclassing

GDYendell: Agree. Let's see what we decide first.

GDYendell:
Alan's odin-control branch has been merged and deleted. I have update fastcs-dev to use odin-control main instead. If you rebase (and make sure to remove the commit to pin odin-control to a branch) it should fix the build here.

jsouter:
> Alan's odin-control branch has been merged and deleted. I have update fastcs-dev to use odin-control main instead. If you rebase (and make sure to remove the commit to pin odin-control to a branch) it should fix the build here.

Okay pushed that, seems to work as before after rebasing.

ajgdls (changes requested):
Would it be possible to break out the FrameProcessorController into its own file (so basically one file for each class, named the same as the class but with underscore rather than camel case)?  I think it will make it clearer to see what is available if you are just looking around the directories.

jsouter:
> Would it be possible to break out the FrameProcessorController into its own file (so basically one file for each class, named the same as the class but with underscore rather than camel case)? I think it will make it clearer to see what is available if you are just looking around the directories.

Okay, just pushed that change

python/src/odin_data/control/odin_data_adapter.py
GDYendell: OK so I think we change this to `self.__class__` now?

GDYendell:
Note this will break fastcs-odin when merged, it will need to have its config files updated.

GDYendell:
@jsouter could you squash the last 3 commits into one? I think we are good to merge.

jsouter:
> @jsouter could you squash the last 3 commits into one? I think we are good to merge.

Done!

GDYendell (approved):

GDYendell:
Uh, does this logic need to be moved somewhere else? https://github.com/odin-detector/odin-data/pull/362/files#diff-af879f9e3b965b6fe0832f413026c82c57dd5f10c0103d6c3a39723f68ff0cdfL41

jsouter:
Ah good point, it's possible, I had assumed this was covered by OdinDataController.process_config_changes. I will try and test if the decoder gets reloaded when it should.

PR #366, opened 2024-10-24
Add start/stop writing commands to FileWriterPlugin

jsouter:
Can't automatically merge as I've built this on top of fastcs-dev rebased on master.

Adds `start_writing` and `stop_writing` commands to the FileWriterPlugin and calls `execute_command` in the `FrameProcessorController`, removing the old logic to send a write value to `config/hdf/write` from the plugin and the adapter.

Happy to rework this to be built on top of master instead of fastcs-dev if preferred.
Commands gets added to the ParameterTree for the FP controller per client

like {"0": {"status": ..., "config": ..., "commands": {"hdf": ["start_writing", "stop_writing"], ...}}

Depends on https://github.com/odin-detector/odin-data/pull/367, see my fork for a more readable diff https://github.com/jsouter/odin-data/pull/1


GDYendell:
I think it should be possible to merge this directly into master and then rebase fastcs-dev once it is merged.

Issue #357, opened 2024-08-16
Add start and stop commands to FileWriterPlugin


GDYendell:
- [x] Depends on #359 

- Implement `"start_writing"` and `"stop_writing"` in `execute` to replace `hdf/config/write`
- Return `["start_writing", "stop_writing"]` in `requestCommands`

---

To be tested alongside
- [ ] https://github.com/odin-detector/odin-data/issues/363
- [ ] https://github.com/DiamondLightSource/fastcs-odin/issues/41

jsouter:
Commenting so I can be assigned!

PR #368, opened 2024-11-20
Add start/stop writing commands to FileWriterPlugin

jsouter:
Fixes https://github.com/odin-detector/odin-data/issues/357

Looks like force pushing auto-closed the other PR so recreating.

Adds support for start_writing and stop_writing commands for the FrameProcessor on top of master. To support this, I had to include Alan's changes for ipc_tornado_client.py by breaking up https://github.com/odin-detector/odin-data/commit/218754111769b122131acd8ee1cb0874aed535e2 into multiple commits and then cherry-picking it - so I will need to rebase fastcs-dev on top of these changes to fix the history.

From my testing, the Start and Stop buttons on the OD screen open and close files in each of the 4 FrameProcessor clients when using Diamond's 1-0-0 release of ADOdin's eiger example IOC against a simulator.

edit: i see tests are failing, will investigate

### DIFF
--- a/cpp/frameProcessor/src/FileWriterPlugin.cpp
+++ b/cpp/frameProcessor/src/FileWriterPlugin.cpp
@@ -64,6 +64,8 @@ const std::string FileWriterPlugin::WRITE_ERROR_DURATION = "write_error_duration
 const std::string FileWriterPlugin::FLUSH_ERROR_DURATION = "flush_error_duration";
 const std::string FileWriterPlugin::CLOSE_ERROR_DURATION = "close_error_duration";
 
+const std::string FileWriterPlugin::START_WRITING = "start_writing";
+const std::string FileWriterPlugin::STOP_WRITING = "stop_writing";
 
 /**
  * Create a FileWriterPlugin with default values.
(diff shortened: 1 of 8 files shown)


GDYendell:
@ajgdls could you take a look at this and see how it fits into your latest changes?

jsouter:
Okay, seems like cpp/test/integrationTest/config/dummyUDP-fp.json explains why the integration test is failing - the config file has a `"write": true` in it, I guess the choices here are to amend the test so the start_writing command is called manually, or I rollback the change to remove the `config/hdf/write` support so we can either send config messages or use the new commands.

cpp/frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -990,6 +972,35 @@ size_t FileWriterPlugin::calc_num_frames(size_t totalFrames)
...
+          this->current_acquisition_ = boost::shared_ptr<Acquisition>(new Acquisition(hdf5_error_definition_));
+        }
+        LOG4CXX_INFO(logger_,
+                      "FrameProcessor will not receive any frames from this acquisition and so no output file will be created");
+      } else {
+        this->start_writing();
+      }
+  } else if (command == FileWriterPlugin::STOP_WRITING) {
+      this->stop_writing();
+  } else std::cout << "Command " << command << " not implemented for FileWriterPlugin\n";
```
jsouter, line 992: Can make this so that `reply` gets updated

jsouter:
Okay integration tests fixed!

ajgdls:
I've looked at this.  My change to IpcTornadoClient has a very similar execution method, so I can simply swap to use this instead.  It looks like the storing of supported commands is identical which is great.

The only thing I notice is that this odin_data_adapter looks like it is based on quite an old version of odin-data, before we switched to fully supported ParameterTree instrospection of applications and using controller class files.  So this is consistent for this version of odin-data, but we will need to revisit the odin_data_adapter file when we apply further odin-data changes.
All of the other C++ changes etc and IpcTornadoClient look fine I think.

jsouter:
> I've looked at this. My change to IpcTornadoClient has a very similar execution method, so I can simply swap to use this instead. It looks like the storing of supported commands is identical which is great.
> 
> The only thing I notice is that this odin_data_adapter looks like it is based on quite an old version of odin-data, before we switched to fully supported ParameterTree instrospection of applications and using controller class files. So this is consistent for this version of odin-data, but we will need to revisit the odin_data_adapter file when we apply further odin-data changes. All of the other C++ changes etc and IpcTornadoClient look fine I think.

Yep, my other branch adds the support in for the new style of adapters, if this gets merged in I can rebase that on top of master.

python/src/odin_data/control/ipc_tornado_client.py
```diff
@@ -26,57 +47,163 @@ def __init__(self, ip_address, port):
...
+        """Handes incoming messages from the main ZeroMQ socket.
+
+        The handled message types are:
+        - Reply to 'request_version'
+        - Reply to 'request_configuration'
+        - Reply to 'request_commands'
+        - Reply to 'status'
+        - Reply to 'configure'
+
+        Replies to request_version, request_configuration and status
```
GDYendell, line 114: and command

GDYendell (commented):
This looks fine, but it seems like some of the things added to `IpcTornadoClient` aren't actually used, e.g. `read_rejected_configs`, `check_for_rejection`, `wait_for_response`. Is this because the change is cherry picked and it is used in fastcs-dev?

GDYendell:
@ajgdls what are these new methods for?

GDYendell (approved):

timcnicholls (approved):

PR #379, opened 2024-12-17
Implement command API in FrameReceiver

jsouter:
Prevent error response from FR
@GDYendell 

- [x] Depends on #388

### DIFF
--- a/cpp/frameReceiver/src/FrameReceiverController.cpp
+++ b/cpp/frameReceiver/src/FrameReceiverController.cpp
@@ -820,6 +820,12 @@ void FrameReceiverController::handle_ctrl_channel(void)
             this->request_configuration(ctrl_reply);
             break;
 
+          case IpcMessage::MsgValCmdRequestCommands:
+            LOG4CXX_DEBUG_LEVEL(3, logger_,
+                "Got control channel read commands request from client " << client_identity);
+            this->request_commands(ctrl_reply);
+            break;
+
           case IpcMessage::MsgValCmdStatus:
             LOG4CXX_DEBUG_LEVEL(3, logger_,
                 "Got control channel status request from client " << client_identity);
@@ -1203,6 +1209,21 @@ void FrameReceiverController::request_configuration(OdinData::IpcMessage& config
   // Add frame count to reply parameters
   config_reply.set_param(CONFIG_FRAME_COUNT, config_.frame_count_);
 
+}
+
+/**
+ * Request the command set supported by this FrameReceiver
+ *
+ * \param[out] reply - Response IpcMessage with the current supported command set.
+ */
+void FrameReceiverController::request_commands(OdinData::IpcMessage& reply)
+{
+  LOG4CXX_DEBUG_LEVEL(3, logger_, "Request for supported commands made");
+  std::vector<std::string>::iterator cmd;
+  std::vector<std::string> decoder_commands = frame_decoder_->request_commands();
+  for (cmd = decoder_commands.begin(); cmd != decoder_commands.end(); ++cmd) {
+    reply.set_param("decoder/supported[]", *cmd);
+  }
 }
 
 //! Reset the frame receiver statistics.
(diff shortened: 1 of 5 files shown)


GDYendell:
Thanks @jsouter ! We should probably also add `FrameReceiverController::execute` that calls the decoder execute and `FrameDecoder::execute` that just returns an error like [here](https://github.com/odin-detector/odin-data/pull/359/files#diff-3aeb57affb4eb79f843f3c8ee926cb64796ea427e03204c4fd2ec5c96fc3bb23R179), so that when a specific decoder adds a command it works.

jsouter:
Okay, done that. Note that the program panics when trying to send a command to the decoder, same is true for FP unless we're sending the command to the HDF plugin in which case the message just gets a NACK response for an invalid command - do we want to always have the process exit on a bad command, always just return a NACK or is it fine as is?

GDYendell:
Ah, so this error is not caught, it just crashes? It should return NACK and keep running.

cpp/frameReceiver/src/FrameReceiverController.cpp
```diff
@@ -1205,6 +1218,54 @@ void FrameReceiverController::request_configuration(OdinData::IpcMessage& config
...
+ * name and a structure of any parameters required by the command.
+ *
+ * \param[in] config - IpcMessage containing command and any parameter data.
+ * \param[out] reply - Response IpcMessage.
+ */
+void FrameReceiverController::execute(OdinData::IpcMessage& config, OdinData::IpcMessage& reply)
+{
+  LOG4CXX_DEBUG_LEVEL(1, logger_, "Command submitted: " << config.encode());
+
+  // Loop over plugins, checking for command messages
```
GDYendell, line 1235:
```suggestion
  // Check for decoder commands
```

cpp/frameReceiver/src/FrameReceiverController.cpp
```diff
@@ -1205,6 +1218,54 @@ void FrameReceiverController::request_configuration(OdinData::IpcMessage& config
...
+                                    config.get_msg_val());
+    if (subConfig.has_param("command")){
+      commandPresent = true;
+      // Extract the command and execute on the decoder
+      std::string commandName = subConfig.get_param<std::string>("command");
+      frame_decoder_->execute(commandName, reply);
+    }
+  }
+  if (!commandPresent){
+    // If no valid commands have been found after checking through all plugins then NACK the reply
```
GDYendell, line 1249:
```suggestion
    // If no valid commands have been found then NACK the reply
```

GDYendell (commented):
Squash / force push these changes in

GDYendell (approved):

cpp/frameReceiver/src/FrameReceiverController.cpp
```diff
@@ -1205,6 +1218,54 @@ void FrameReceiverController::request_configuration(OdinData::IpcMessage& config
...
+ *
+ * \param[in] config - IpcMessage containing command and any parameter data.
+ * \param[out] reply - Response IpcMessage.
+ */
+void FrameReceiverController::execute(OdinData::IpcMessage& config, OdinData::IpcMessage& reply)
+{
+  LOG4CXX_DEBUG_LEVEL(1, logger_, "Command submitted: " << config.encode());
+
+  // Check for decoder commands
+  bool commandPresent = false;
```
timcnicholls, line 1236: Some local variables (commandPresent, subConfig, commandName) should be snake rather than camel case.

timcnicholls (approved):
I've made a very nitpicky comment about variable naming style, otherwise this is fine to merge.

GDYendell:
CI is broken for unrelated reasons :disappointed: 

It looks like we 3.7 is not available in the test runners any more and versioneer is broken.

I think we need to drop 3.7 and either update versioneer and hope that `configparser` error goes away or swap to setuptools_scm, which is now stable and we are using for all python projects at DLS now. Any thoughts @timcnicholls?

PR #381, opened 2025-01-14
Update Codebase to C++17 Standard

Ohisemega:
Update rapidjson to latest version
Remove boost::shared_ptr, boost::optional, boost::bind, boost::mutex, boost::to_string, boost::lexical_cast<std::string>, boost::function, etc Replace the above boos libraries with std equivalent


GDYendell:
Could you split the rapid json update into its own commit? The diff is massive and it is probably 95% rapidjson.

Ohisemega:
> Could you split the rapid json update into its own commit? The diff is massive and it is probably 95% rapidjson.

Done!

PR #385, opened 2025-01-29
Update to c++17

Ohisemega:
Update Codebase to C++17 Standard

Remove boost::shared_ptr, boost::optional, boost::bind, boost::mutex, boost::to_string, boost::lexical_cast<std::string>, boost::function, etc
Replace the above boost libraries with std:: equivalent.
Update the Rapidjson library to version 1.1.0

### DIFF
--- a/cpp/frameReceiver/src/FrameReceiverController.cpp
+++ b/cpp/frameReceiver/src/FrameReceiverController.cpp
@@ -137,7 +137,7 @@ void FrameReceiverController::run(void)
 
 #ifdef FR_CONTROLLER_TICK_TIMER
   int tick_timer_id = reactor_.register_timer(
-      deferred_action_delay_ms, 0, boost::bind(&FrameReceiverController::tick_timer, this));
+      deferred_action_delay_ms, 0, std::bind(&FrameReceiverController::tick_timer, this));
 #endif
 
   // Run the reactor event loop
@@ -172,7 +172,7 @@ void FrameReceiverController::stop(const bool deferred)
   if (deferred)
   {
     reactor_.register_timer(deferred_action_delay_ms, 1,
-      boost::bind(&FrameReceiverController::stop, this, false)
+      std::bind(&FrameReceiverController::stop, this, false)
     );
   }
   else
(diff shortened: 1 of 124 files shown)


cpp/frameReceiver/src/FrameReceiverController.cpp
```diff
@@ -470,7 +470,7 @@ void FrameReceiverController::configure_frame_decoder(OdinData::IpcMessage& conf
   // decoder configuration; update and force a reconfig if so
   if (config_msg.has_param(CONFIG_DECODER_CONFIG))
   {
-    boost::scoped_ptr<IpcMessage> new_decoder_config(
+    std::unique_ptr                                                                                                            <IpcMessage> new_decoder_config(
```
GDYendell, line 473: Is the diff being weird or is there a load of whitespace here? Could that be removed?

GDYendell:
Could you reword the first commit so that "Update the Rapidjson libary to version 1.1.0" is in the subject instead of the body? Let me know if you need help with this.

cpp/frameReceiver/src/FrameReceiverController.cpp
Ohisemega: I'll fix it :) its a missed key-stroke.

Ohisemega:
Hi Gary


> Could you reword the first commit so that "Update the Rapidjson library to version 1.1.0" is in the subject instead of the body? Let me know if you need help with this.

Yes, I've done this!

ajgdls:
Hi @Ohisemega , to make this easier to progress it would be better if we can split it into multiple smaller PRs which can then be reviewed and potentially merged.
To start with, it would be nice to have a PR that lets the current odin-data source build under c++17 without any compiler warnings.  Would you consider creating a new PR that just resolves that issue and then we could take a look at it?
Thanks

Ohisemega:
> Hi @Ohisemega , to make this easier to progress it would be better if we can split it into multiple smaller PRs which can then be reviewed and potentially merged. To start with, it would be nice to have a PR that lets the current odin-data source build under c++17 without any compiler warnings. Would you consider creating a new PR that just resolves that issue and then we could take a look at it? Thanks

Hi Alan!

Pardon my late response to this message @ajgdls . I would create a new issue in that case and open a new PR.

PR #389, opened 2025-01-31
Replace versioneer with setuptools_scm

### LINKED ISSUE #388: Replace versioneer with setuptools_scm

GDYendell:
Drop python 3.7 as 3.8 is the minimum supported version
Move all config to pyproject.toml now that we don't need to call versioneer in setup.py
Fix coverage paths for src package restructure

Fixes #388


GDYendell:
@timcnicholls the CI still doesn't pass because odin-control has the same problem. I expect CI is broken on main there too.

GDYendell:
OK the docs build does run on 3.11. So actually maybe the problem is that the install was failing on whatever the default python environment is (the docs build was not using setup-python). I still think this is a good thing to get in, though.

timcnicholls:
@GDYendell I've realised after approving this that dropping 3.7 support is problematic as our embedded systems are based on 3.7.6 We can look to update this but it needs validation. We can live in the short term with using the current odin-data version but we'll to plan this transition.

PR #392, opened 2025-02-14
Log error when TCP socket reads negative bytes

jsouter:
Also removes warning about pointer arithmetic during compilation (maybe better to change the buffer type to char* at init? It would be a breaking change so fine to leave it)

I'll set as ready once I've managed to test the failure case with merlin-detector, which I'm having to make some minor tweaks to get it working with the new odin-data/FastCS

### DIFF
--- a/cpp/frameReceiver/src/FrameReceiverTCPRxThread.cpp
+++ b/cpp/frameReceiver/src/FrameReceiverTCPRxThread.cpp
@@ -107,7 +107,7 @@ void FrameReceiverTCPRxThread::handle_receive_socket(int recv_socket_,
                                                      int recv_port) {
   // Receive a message from the main thread channel and place it directly into
   // the provided memory buffer
-  void *frame_buffer = frame_decoder_->get_next_message_buffer();
+  char *frame_buffer = static_cast<char*>(frame_decoder_->get_next_message_buffer());
   size_t message_size = frame_decoder_->get_next_message_size();
   size_t bytes_received = 0;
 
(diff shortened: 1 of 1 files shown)


GDYendell:
Is this ready for review @jsouter ?

jsouter:
> Is this ready for review @jsouter ?

Yes, I was hoping to test it against a full deployment which I've not been able to get together yet (my merlin-detector only works with an older odin-data and ADOdin and I'm about halfway through updating it). I haven't gotten an image out but I have seen a notification in the FP that it acknowledged a frame was ready for processing after telling the detector to acquire 1 image. It's a simple enough change that that's probably sufficient.

BenBradnick:
@GDYendell I can't test this branch specifically because there are some breaking API changes in December which upset Xspress detector (I guess I should look into this at some point).

I'm happy with the new behavior in this change though.

cpp/frameReceiver/src/FrameReceiverTCPRxThread.cpp
```diff
@@ -107,15 +107,21 @@ void FrameReceiverTCPRxThread::handle_receive_socket(int recv_socket_,
...
-        read(recv_socket_, frame_buffer, message_size - bytes_received);
-    bytes_received += msg_len;
-    frame_buffer += msg_len;
+  while (total_bytes_read < message_size) {
+    bytes_read = read(recv_socket_, frame_buffer, message_size - total_bytes_read);
+
+    if (bytes_read < 0) {
+      LOG4CXX_ERROR(logger_, "Error reading from socket in TCP RX thread");
+      return;
+    }
```
jsouter, line 121: TODO: test what happens when there is a socket connection failure, will this get called repeatedly?

ajgdls: Hello @jsouter , is this something that you have been able to test yet, or have a plan to be able to test?

jsouter: I added a little extra logging, looks like the `read(recv_socket_, frame_buffer, message_size - total_bytes_read)` continues to get called on loop, but returns 0 every time.

PR #398, opened 2025-03-31
Mention the correct binary (frameSimulator) in user tutorials.

gustavosr8:


GDYendell:
Thanks for spotting that @gustavosr8 !

Can I ask, what are you currently investigating Odin for? Are you looking at support for specific detectors? I could provide some better guidance than is currently in the docs if I know what your looking to achieve.

GDYendell (approved):

gustavosr8:
Hi Gary, thanks for the reply!
So, here at LNLS we are developing a new detector, TUPI, and part of the POCs for this project is evaluating some DAC frameworks, and one of them is Odin (I think you may have talked to @ericonr about LImA 2 for the same purpose as well).

At the moment I'm trying to use the example application and after that I'll use it to evaluate the performance of our processing infrastructure.

GDYendell:
That sounds like a great idea. There has been a lot of work put into various frameworks, but not often people investigating to see if they could adopt them.

Let me know if you need any help getting things set up. The docs we have are pretty minimal and possibly incorrect, as you have found, because, as far as I know, no one outside DLS / STFC has read them.

PR #400, opened 2025-04-11
DO NOT MERGE! BUG: Test Crash During ClassModule CPP17 Update

Ohisemega:
Line 57 in the ClassLoader.h template class header file leads to a crach in the frameProcessorTest The Update simply utilizes the new CPP17 feature `static inline <variable_name>` Which allows Class headers to be defined in CPP with static variables not having to be initialized with a static method.

### DIFF
--- a/cpp/common/include/ClassLoader.h
+++ b/cpp/common/include/ClassLoader.h
@@ -54,7 +54,7 @@ public:
    */
   ClassLoader(std::string name, maker_t *value)
   {
-    factory_map()[name] = value;
+    factory[name] = value;
   }
 
   /**
@@ -86,8 +86,8 @@ public:
     }
     boost::shared_ptr<BaseClass> obj;
     try {
-      if (factory_map().count(name)) {
-        obj = factory_map()[name]();
+      if (factory.count(name)) {
+        obj = factory[name]();
       }
     } catch(const std::exception& ex) {
       // Most likely the requested class and loaded class do not match.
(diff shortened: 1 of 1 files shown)


Ohisemega:
@GDYendell 
While working on updating the codebase to CPP17. I encountered this interesting bug, which causes the `frameProcessorTest` to crash. I modified the `ClassLoader.h` file to utilize a CPP17 feature called `static inline` for template class data members. 
Synopsis: _The apparent issue with defining static data members in a class header is including them in multiple translation files (.cpp files). This would lead to more than one instantiation, and the linker, at link time, is unhappy. But since our `ClassLoader` is a template class, it isn't instantiated until specialized. So, ideally, everything should work nicely, but the program crashes at line 57 when the program attempts to initialize the static data member_

What I think I have encountered is a  **static initialization order fiasco**: The static data member has not been initialized when the `ClassLoader` constructor is called. To be fair, the previous Meyer's Singleton, which I attempted to replace, is a good solution; what I missed out is that the factory object is also a templated and **Dynamic Initialization** rule would mean that the variable's definition is unordered.

Ohisemega:
Hello Alan! @ajgdls I wanted to know what the function `cl(Name, OdinData::maker<Base,Class>)`  in the ClassLoader is defined. I tried searching the codebase and the included header files, but I haven't found a definition for it.

`#define REGISTER(Base,Class,Name) OdinData::ClassLoader<Base> cl(Name, OdinData::maker<Base,Class>);`

Ohisemega:
> Line 57 in the ClassLoader.h template class header file leads to a crash in the frameProcessorTest. The Update simply utilizes the new CPP17 feature `static inline <variable_name>`, which allows Class headers to be defined in C++ with static variables not having to be initialized with a static method.

I realized the issue, I had a misconception, which led me to believe that the static member object `static std::map<std::string, maker_t *> factory;` was well defined before the instantiation of the class. This issue is known as the **static initialization order fiasco**, and it is related to the **Dynamic Initialization** of static non-local templated objects in C++.
Synopsis: _The apparent issue with defining static data members in a class header is including them in multiple translation files (.cpp files). This would lead to more than one instantiation, and the linker, at link time, is unhappy. But since our ClassLoader is a template class, it isn't instantiated until specialized. So, ideally, everything should work nicely, but the program crashes at line 57 when the program attempts to initialize the static data member_

What I have encountered is a _**static initialization order fiasco**_: The static data member has not been initialized when the ClassLoader constructor is called. To be fair, the previous Meyer's Singleton, which I attempted to replace, is a good solution; what I missed out is that the factory object is also a templated and Dynamic Initialization rule would mean that the variable's definition is unordered.

GDYendell:
We have understood the problem with the new code and why it had to be the way it was.

PR #411, opened 2025-09-01
Implement requestMetadata Within C++ FP Apps

### LINKED ISSUE #406: Implement requestMetadata within C++ FP applications
Implement base method for requestMetadata.  Pass in the same IpcMessage that was passed into the requestConfiguration method (consider also a version for the status request messages).
Call the method for each plugin within the control class.
Ensure the existing API remains unchanged so that all plugins (odin-data and external detector) compile and run without any required changes other than a rebuild.

Ohisemega:
Add a requestStatusMetadata method  to the base FrameProcessorPlugin class. Add a requestMetadata method  to the base FrameProcessorPlugin class. Call requestStatusMetadata in the provideStatus method of FrameProcessorController for all plugins. Call requestMetadata in requestConfiguration method of FrameProcessorController for all plugins.

Fixes #406

### DIFF
--- a/cpp/frameProcessor/include/FrameProcessorController.h
+++ b/cpp/frameProcessor/include/FrameProcessorController.h
@@ -41,10 +41,10 @@ public:
   virtual ~FrameProcessorController();
   void handleCtrlChannel();
   void handleMetaRxChannel();
-  void provideStatus(OdinData::IpcMessage& config, OdinData::IpcMessage& reply);
+  void provideStatus(OdinData::IpcMessage& reply, bool metadata);
   void provideVersion(OdinData::IpcMessage& reply);
   void configure(OdinData::IpcMessage& config, OdinData::IpcMessage& reply);
-  void requestConfiguration(OdinData::IpcMessage& config, OdinData::IpcMessage& reply);
+  void requestConfiguration(OdinData::IpcMessage& reply, bool metadata);
   void execute(OdinData::IpcMessage& config, OdinData::IpcMessage& reply);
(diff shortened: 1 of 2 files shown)


cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -45,6 +45,8 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
   std::vector<std::string> get_warnings();
   virtual void configure(OdinData::IpcMessage& config, OdinData::IpcMessage& reply);
   virtual void requestConfiguration(OdinData::IpcMessage& reply);
+  virtual void requestMetadata(OdinData::IpcMessage& reply);
```
GDYendell, line 48:
```suggestion
  virtual void requestConfigurationMetadata(OdinData::IpcMessage& reply);
```

ajgdls:
Sorry @GDYendell I don't think the conversation was entirely matched to what you just stated?

Ohisemega:
Hi @GDYendell ,
I wanted to clarify a detail on the new approach. The **_FrameProcessorController_** currently has the methods: `void provideStatus(IpcMessage&) `and `void requestConfiguration(IpcMessage&)`. With the new approach, my understanding is that there will be two new **_FrameProcessorController_** methods, possibly named:  `void requestConfigurationMetadata(IpcMessage&)` and  `void requestStatusMetadata(IpcMessage&)` to call the respective _**FrameProcessorPlugin**_ (parent class and non-virtual) method for requesting configuration and status metadata, respectively.

PR #414, opened 2025-09-12
Add "Request Metadata" Methods FrameProcessorPlugin Class

Ohisemega:


GDYendell:
I think this is just about there now. Could we add a test that calls the two methods with metadata=true and asserts that the messages returned has an empty "metadata" dictionary? Then when we implement the methods in some of the plugins that test will need to be updated to pass.

cpp/common/include/ParamContainer.h
```diff
@@ -308,4 +308,4 @@ class ParamContainer
 
 } // namespace OdinData
 
-#endif // PARAMCONTAINER_H_
```
ajgdls, line 311: Can you just confirm what changed here?  Is this new line removal, I'm just not seeing it.

ajgdls (approved):
The only minor comment I have is that most of these changes were code cleanup rather than to do with the PR.
I think @GDYendell already mentioned adding a unit test.

cpp/common/include/ParamContainer.h
GDYendell: It is GitHub saying there was no new line at the end of the file and Famous has added it, which is correct.

ajgdls:
In case I'm holding this up, I have no issues with the refactoring :+1:

Issue #407, opened 2025-08-28
Design and implement helper methods for adding meta data


ajgdls:
Adding meta data will require specific addition of JSON parameters to the IpcMessage reply.  This must not alter the existing API for replying to configuration (and possibly status) requests.
The helper methods should make it easy for a developer to add meta data, removing any complexity of updating the existing JSON structure if at all possible.

GDYendell:
Are we planning to add a `metadata` key at the top-level that then mirrors the existing tree? e.g. for

```
{
        "ctrl_endpoint": "tcp://0.0.0.0:10004",
        "hdf": {
            "file": {
                "first_number": 1,
            },
            "frames": 0,
        },
}
```

we add 
```
{
    "metadata": {
        "ctrl_endpoint": {"type": "str", "access_mode": "rw"},
        "hdf": {
            "file": {
                "first_number": {"type": "bool", "access_mode": "rw"}
            },
            "frames": {"type": "int", "access_mode": "rw", "min": 0}
        },
}
```

other keys could be `allowed_values`, `max`, `units`.

GDYendell:
From conversation with @Ohisemega. We should make a struct to store metadata. Each plugin will define vector of these structs, one for each parameter. Then there can be a helper function that takes an IPC message and calls set_param for each field using slash separation to define the nesting.

```cpp
struct ParamMetadata {
    std::string path;
    std::string type;
    std::string access_mode;
    std::vector<T> allowed_values;
    int min;
    int max;
    ...

    bool has_min;  // So that we know whether the value is a min or just unset
    bool has_max;
};

void add_metadata(IPCMessage message, ParamMetadata metadata) {
    message.set_param("metadata" / metadata.path / "type", metadata.type)
    ...
}
```

And in fact, if the path is in the struct, I think the base class could implement requestMetadata...

@ajgdls is this what you had in mind? Or if you didn't have something in mind, what do you think?

ajgdls:
No, eventually we might make use of the metadata within the application if we ever implemented a fully structured ParameterTree.  But that feels like a long way off, so a vector is fine for this.

PR #415, opened 2025-09-13
Metadata helper methods issue 407 fix

Ohisemega:


Ohisemega:
This PR replaces the #413 PR and addresses issue #407.

Ohisemega:
Hi @GDYendell,

I have a few concerns I wanted to raise:

- While implementing the helper functions for the `FrameProcessorPlugin` class, I realised that we might need each plugin to have two vectors to contain the metadata for the configuration and the status separately. Is this in line with your thoughts?
- Secondly, the current implementation, which runs over a vector of parameters for **each** plugin, which I speculate to have a significant performance overhead. Is this a concern that we might share, as it would matter how often we request metadata and its overall impact on the application?

GDYendell:
@Ohisemega I have changed the target branch to add-request-metadata. Notice the diff is now correct and will remain the same when #414 is merged and this is updated to target main.

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -66,6 +66,29 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
...
+  */
+  void add_metadata(OdinData::IpcMessage& message, const OdinData::ParamMetadata& metadata) const
+  {
+    // TO-DO add "allowed_values" metadata
+    message.set_param("metadata/" +  metadata.path +  "/type", metadata.type);
+    message.set_param("metadata/" +  metadata.path +  "/access_mode", metadata.access_mode);
+    message.set_param("metadata/" +  metadata.path +  "/min", metadata.min);
+    message.set_param("metadata/" +  metadata.path +  "/max", metadata.max);
+    message.set_param("metadata/" +  metadata.path +  "/has_min", metadata.has_min);
+    message.set_param("metadata/" +  metadata.path +  "/has_max", metadata.has_max);
```
GDYendell, line 81: This should do a match case on `allowed_values.which` and call set_param with the correct `allowed_values.get<>()`. You can use the "array notation" like in the [FileWriterPlugin](https://github.com/odin-detector/odin-data/blob/master/cpp/frameProcessor/src/FileWriterPlugin.cpp#L396).

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -66,6 +66,29 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
...
+  }
+
+  /** This is a private pure virtual method which
+   *  MUST be customized by every derived FrameProcessor 
+   *  plugin class that can append metadata.
+   *  It is essentially something in guise of a factory method.
+   * \returns a vector of ParamMetadata
+   */
+  virtual std::vector<OdinData::ParamMetadata>& get_plugin_metadata() const;
```
GDYendell, line 90:
```suggestion
  virtual std::vector<OdinData::ParamMetadata>& get_metadata() const;
```
plugin is redundant in this context. And as you suggested in your question, there should be two arrays and two getters, for status and config.

cpp/common/include/IpcMessage.h
```diff
@@ -18,9 +18,27 @@
...
+struct JsonNullType{};
+
+/** This struct is a representation of the metadata
+*/
+struct ParamMetadata
+{
+  const std::string path;
+  const std::string type;
+  const std::string access_mode;
+  const std::vector<boost::variant<std::string, int>> allowed_vals;
```
GDYendell:
```suggestion
  const std::vector<boost::variant<std::string, int>> allowed_values;
```

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -66,6 +66,64 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
...
+
+
+  /** Metadata helper function (implicitly inlined)
+   *  \param [in]  metadata - metadata struct to be read from
+   *  \param [out] message  - IpcMessage to be appended with metadata
+   */
+  void add_metadata(OdinData::IpcMessage& message, const OdinData::ParamMetadata& metadata) const
+  {
+    std::string str = "metadata/";
+    static std::string& str_static = str;
```
GDYendell, line 94: These make it harder to read. Could we just inline the string construction? They aren't going to be that long because it is mostly contained in `metadata.path`.

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -66,6 +66,64 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
...
+    static OdinData::IpcMessage&& message_static = std::move(message);
+
+    str += metadata.path;
+
+    message.set_param(str +  "/type", metadata.type);
```
GDYendell, line 99: This needs to include the plugin name, like [here](https://github.com/odin-detector/odin-data/blob/master/cpp/frameProcessor/src/FileWriterPlugin.cpp#L378).

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -66,6 +66,64 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
...
+    message.set_param(str +  "/access_mode", metadata.access_mode);
+    if(metadata.has_min)
+      message.set_param(str +  "/min", metadata.min);
+    if(metadata.has_max)
+      message.set_param(str +  "/max", metadata.max);
+
+    str +=  "/allowed_vals[]";
```
GDYendell, line 106:
```suggestion
    str +=  "/allowed_values[]";
```

cpp/frameProcessor/src/FrameProcessorPlugin.cpp
```diff
@@ -168,22 +168,44 @@ void FrameProcessorPlugin::requestConfiguration(OdinData::IpcMessage& reply)
...
+}
+
+/** Request the plugin's Status Metadata.
+ *
+ * \returns reference to vector of ParamMetadata structs specific to each plugin.
+ */
+std::vector<OdinData::ParamMetadata>& FrameProcessorPlugin::get_pluginstatus_metadata() const noexcept
+{
+  
+}
```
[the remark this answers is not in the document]
Ohisemega: Ah, I was conflicted, but I remember you mentioned it, I will push an upate

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -29,10 +29,11 @@ struct JsonNullType{};
 */
 struct ParamMetadata
 {
+  using param_metadata_vec = std::vector<boost::variant<std::string, int>>;
   const std::string path;
   const std::string type;
   const std::string access_mode;
-  const std::vector<boost::variant<std::string, int>> allowed_vals;
+  const param_metadata_vec allowed_values;
```
GDYendell, line 36: This name is very confusing. It is not a vector of parameter metadata, that would better describe what the plugins will return from `get_*_metadata`. What about this?

```suggestion
  using allowed_value_t = boost::variant<std::string, int>;
  const std::vector<allowed_value_t> allowed_values;
```

Is this type going to be needed by the specific plugins?

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -90,45 +91,32 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
    */
   void add_metadata(OdinData::IpcMessage& message, const ParamMetadata& metadata) const
   {
-    std::string str = "metadata/";
-    static std::string& str_static = str;
-    static OdinData::IpcMessage&& message_static = std::move(message);
-
-    str += metadata.path;
-
+    std::string str = get_name() + metadata.path + '/';
```
GDYendell, line 94: This needs to be nested under "metadata" otherwise it will clash with the actual values.

```suggestion
    std::string str = "metadata" + get_name() + metadata.path + '/';
```

GDYendell, line 94:
```suggestion
    std::string param_prefix = get_name() + metadata.path + '/';
```

The parameter in the set_param method that we are passing this full string to is called `param_name`, so this maps nicely.

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -90,45 +91,32 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
...
-    auto end = metadata.allowed_vals.end();
-    for (; first != end; ++first)
-	    boost::apply_visitor(VariantVisitor(), *first);
-    // std::for_each(metadata.allowed_vals.begin(), metadata.allowed_vals.end(), boost::apply_visitor(VariantVisitor()));
-
+    }
+
+    str +=  "/allowed_values[]";
+
+    auto first = metadata.allowed_values.begin();
```
GDYendell, line 106:
```suggestion
    auto itr = metadata.allowed_values.begin();
```

It isn't first for the whole for loop.

cpp/common/include/IpcMessage.h
```diff
@@ -18,9 +18,11 @@
 
 #include "boost/date_time/posix_time/posix_time.hpp"
 #include "boost/bimap.hpp"
+#include "boost/variant.hpp"
 
 namespace OdinData
 {
+
 //! IpcMessageException - custom exception class implementing "what" for error string
```
GDYendell, line 26: This can be removed now because ParamMetadata has been moved

cpp/frameProcessor/include/FrameProcessorPlugin.h
Ohisemega: Hi Gary,

I looked at one of the sample json structure you shared on issue #407 :
```
{
    "metadata": {
        "ctrl_endpoint": {"type": "str", "access_mode": "rw"},
        "hdf": {
            "file": {
                "first_number": {"type": "bool", "access_mode": "rw"}
            },
            "frames": {"type": "int", "access_mode": "rw", "min": 0}
        },
}
```
I assumed the `"metadata"` part of the string was going to be within the `metadata.path` string.

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -66,6 +66,64 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
...
+      void operator() (int n) const{
+        message_static.set_param(str_static, n);
+      }
+    };
+
+    auto first = metadata.allowed_vals.begin();
+    auto end = metadata.allowed_vals.end();
+    for (; first != end; ++first)
+	    boost::apply_visitor(VariantVisitor(), *first);
+    // std::for_each(metadata.allowed_vals.begin(), metadata.allowed_vals.end(), boost::apply_visitor(VariantVisitor()));
```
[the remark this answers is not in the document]
Ohisemega: While comparing this implementation with the switch-case method. One thing realized is that the switch-case method relies on knowing the position of the template arguments the variant (`boost::variant<std::string, int>`) is specialized with.

Essentially, if for some reason in the future, we change the variant's template argument position, then the **_switch case_** must change. The `struct VariantVisitor` relies on template argument deduction however and compilation should fail if the `operator()()` provided doesn't match the variant type passed to it.

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -66,6 +81,51 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
...
+
+
+  /** Metadata helper function (implicitly inlined)
+   *  \param [in]  metadata - metadata struct to be read from
+   *  \param [out] message  - IpcMessage to be appended with metadata
+   */
+  void add_metadata(OdinData::IpcMessage& message, const ParamMetadata& metadata) const
+  {
+    std::string param_prefix = "metadata" + get_name() + metadata.path + '/';
+    message.set_param(param_prefix +  "type", metadata.type);
```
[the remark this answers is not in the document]
Ohisemega: Fixed!

Ohisemega:
Hi @GDYendell, I have responded to all the issues raised related to this PR, and Alan has given his approval as well.

Ohisemega:
@GDYendell @ajgdls, kindly review the latest changes.

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -62,10 +78,148 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
   void push(boost::shared_ptr<Frame> frame);
   void push(const std::string& plugin_name, boost::shared_ptr<Frame> frame);
 
+  /** This struct is a representation of the metadata
+  */
+  struct ParamMetadata
```
timcnicholls: Can this class and associated types be factored out into a separate header in `common/include` (e.g. like ParamContainer) ?

cpp/frameProcessor/src/FrameProcessorPlugin.cpp
```diff
@@ -21,6 +21,8 @@ FrameProcessorPlugin::FrameProcessorPlugin() :
 {
   OdinData::configure_logging_mdc(OdinData::app_path.c_str());
   logger_ = log4cxx::Logger::getLogger("FP.FrameProcessorPlugin");
+  config_metadata_bucket_.reserve(256);
```
timcnicholls: What is the motivation for reserving for 256 elements? How did you arrive at that number? 

Is it necessary from a performance perspective to pre-reserve this number to avoid rehashing? Presumably the metadata maps are only populated once at plugin initialisation, in which case the rehashing overhead has negligible performance penalty, whereas the reservation increases memory utilisation?

I would suggest removing the `reserve(256)` calls if possible. If not, the size should not be a magic number but should be a constant that is defined in the appropriate location.

timcnicholls (changes requested):
Thanks for this PR @Ohisemega ! I like the implementation in general - I've left some specific comments and suggestions for change in the inline comments.

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -62,10 +78,148 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
...
+  {
+    using allowed_values_t = boost::variant<boost::blank, std::string, int>;
+    /**
+     * Note on this Parametized Constructor:
```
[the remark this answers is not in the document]
GDYendell: I would just remove this comment now. It is more of an explanation for the reviewer to consider and not something that should remain in the code forever more.

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -62,10 +78,148 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
...
+
+
 private:
   /** Pointer to logger */
   LoggerPtr logger_;
 
+  /** These private member data are "parameter buckets,"
+   *  they are populated by calling add_status_param_metadata()
+   *  and add_config_param_metadata() methods respectively
+   */
```
GDYendell: This can probably be removed too with the renamed class it is pretty self-explanatory what they do and the methods to add to it are immediately below

cpp/frameProcessor/src/FrameProcessorPlugin.cpp
Ohisemega: Removed the reserve() calls on the maps!

PR #418, opened 2025-10-15
Implement core plugins metadata fix issue 408

### LINKED ISSUE #408: Implement meta data for odin-data core plugins
For each odin-data plugin construct an implementation of the meta data request that fills in appropriate meta data for the configuration (also status requests).

Consider multiple PRs to avoid large codebase changes in a single merge.

Ohisemega:
Fixes #408

### DIFF
--- a/cpp/frameProcessor/include/FrameProcessorPlugin.h
+++ b/cpp/frameProcessor/include/FrameProcessorPlugin.h
@@ -96,8 +96,8 @@ private:
   void add_metadata(OdinData::IpcMessage& message, const std::pair<std::string, ParamMetadata>& metadata) const
   {
     std::string param_prefix;
-    param_prefix.reserve(100);
-    "metadata/" + get_name() + '/' + metadata.first + '/';
+    param_prefix.reserve(128);
+    param_prefix = "metadata/" + get_name() + '/' + metadata.first + '/';
     message.set_param(param_prefix +  "type", metadata.second.type);
     message.set_param(param_prefix +  "access_mode", metadata.second.access_mode);
     if(metadata.second.min == INT32_MIN){


Ohisemega:
@ajgdls  @GDYendell

cpp/frameProcessor/src/BloscPlugin.cpp
```diff
@@ -55,6 +55,11 @@ void create_cd_values(const BloscCompressionSettings& settings, std::vector<unsi
 BloscPlugin::BloscPlugin() :
 current_acquisition_(""), data_buffer_ptr_(NULL), data_buffer_size_(0)
 {
+  FrameProcessorPlugin::config_metadata_bucket_ = { {CONFIG_BLOSC_COMPRESSOR, {"int", "rw", {BLOSC_BLOSCLZ, BLOSC_LZ4, BLOSC_LZ4HC, BLOSC_SNAPPY, BLOSC_ZLIB, BLOSC_ZSTD}, INT32_MIN, INT32_MAX} },
+                                                    {CONFIG_BLOSC_THREADS,    {"int", "rw", {},                                                                            1,         INT32_MAX} },
+                                                    {CONFIG_BLOSC_LEVEL,      {"int", "rw", {},                                                                            1,         9        } },
+                                                    {CONFIG_BLOSC_SHUFFLE,    {"int", "rw", {BLOSC_NOSHUFFLE, BLOSC_SHUFFLE, BLOSC_BITSHUFFLE},                            INT32_MIN, INT32_MAX} }
+};
```
GDYendell: This makes me think a helper method for adding parameters would be better. We can use default values and different overrides to make a nicer API and to add restrictions, for example so that you can only pass allowed values or min/max.

```cpp
create_parameter(CONFIG_BLOSC_COMPRESSOR, "int", "rw", {BLOSC_BLOSCLZ, BLOSC_LZ4, BLOSC_LZ4HC, BLOSC_SNAPPY, BLOSC_ZLIB, BLOSC_ZSTD});
create_parameter(CONFIG_BLOSC_LEVEL, "int", "rw", 1, 9)
```

Thoughts @Ohisemega @ajgdls

PR #424, opened 2025-11-05
Implement  FileWriterPlugin Metadata Fix Issue 408

Ohisemega:
FileWriterPlugin Metadata Implementation

### DIFF
--- a/cpp/frameProcessor/src/FileWriterPlugin.cpp
+++ b/cpp/frameProcessor/src/FileWriterPlugin.cpp
@@ -117,7 +117,7 @@ FileWriterPlugin::FileWriterPlugin() :
   add_config_param_metadata(FileWriterPlugin::CONFIG_MASTER_DATASET, PMDD::STRING_T, PMDA::READ_WRITE);
   add_config_param_metadata(FileWriterPlugin::ACQUISITION_ID, PMDD::STRING_T, PMDA::READ_WRITE);
   add_config_param_metadata(FileWriterPlugin::CLOSE_TIMEOUT_PERIOD, PMDD::UINT_T, PMDA::READ_WRITE, 0, PMD::MAX_UNSET);
-  add_config_param_metadata(FileWriterPlugin::START_CLOSE_TIMEOUT, PMDD::UINT_T, PMDA::READ_WRITE, 0, PMD::MAX_UNSET);
+  add_config_param_metadata(FileWriterPlugin::START_CLOSE_TIMEOUT, PMDD::BOOL_T, PMDA::READ_WRITE);
 
   add_status_param_metadata(STATUS_WRITING, PMDD::BOOL_T, PMDA::READ_ONLY);
   add_status_param_metadata(STATUS_FRAMES_MAX, PMDD::UINT_T, PMDA::READ_ONLY, 0, PMD::MAX_UNSET);


cpp/frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -92,6 +92,44 @@ FileWriterPlugin::FileWriterPlugin() :
...
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_PROCESS_NUMBER, "uint", "rw", 1, ParamMetadata::MAX_UNSET);
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_PROCESS_RANK, "uint", "rw", 0, ParamMetadata::MAX_UNSET);
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_PROCESS_BLOCKSIZE, "uint", "rw", 1, ParamMetadata::MAX_UNSET);
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_PROCESS_BLOCKS_PER_FILE, "uint", "rw", 0, ParamMetadata::MAX_UNSET);
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_PROCESS_EARLIEST_VERSION, "bool", "rw", {0, 1});
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_PROCESS_ALIGNMENT_THRESHOLD, "uint", "rw", 1, ParamMetadata::MAX_UNSET);
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_PROCESS_ALIGNMENT_VALUE, "uint", "rw", 1, ParamMetadata::MAX_UNSET);
+
+  prefix = FileWriterPlugin::CONFIG_FILE + '/';
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_FILE_PREFIX, "string", "rw", {});
```
GDYendell: Ah. I hadn't registered that we didn't add an overload that doesn't require min/mac or allowed_values. That would be better.

cpp/frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -92,6 +92,44 @@ FileWriterPlugin::FileWriterPlugin() :
...
+  add_config_param_metadata(FileWriterPlugin::CLOSE_TIMEOUT_PERIOD, "uint", "rw", 0, 5);
+  add_config_param_metadata(FileWriterPlugin::START_CLOSE_TIMEOUT, "uint", "rw", 0, ParamMetadata::MAX_UNSET);
+
+  add_status_param_metadata("writing", "bool", "rw", {0, 1});
+  add_status_param_metadata("frames_max", "uint", "rw", 0, ParamMetadata::MAX_UNSET);
+  add_status_param_metadata("frames_written", "uint", "r", 0, ParamMetadata::MAX_UNSET);
+  add_status_param_metadata("frames_processed", "uint", "r", 0, ParamMetadata::MAX_UNSET);
+  add_status_param_metadata("file_path", "string", "r", {});
+  add_status_param_metadata("file_name", "string", "r", {});
+  add_status_param_metadata(FileWriterPlugin::ACQUISITION_ID, "string", "r", {});
```
GDYendell: It is odd for some of these to use literals and some constants

GDYendell (changes requested):
There are some unrelated changes here that I don't think we want to include.

I am wondering if we should have constants for "bool"/"int" and "r"/"rw" etc, or maybe even an enum so we could restrict what is passed to the functions.

I would be OK with adding `remove_*_metadata` methods to handle datasets being deleted, but I don't think it is critical.

cpp/frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -92,6 +92,44 @@ FileWriterPlugin::FileWriterPlugin() :
...
+  add_status_param_metadata("processes",  "uint", "rw", 1, ParamMetadata::MAX_UNSET);
+  add_status_param_metadata(FileWriterPlugin::CONFIG_PROCESS_RANK,"uint", "rw", 0, ParamMetadata::MAX_UNSET);
+  add_status_param_metadata("/timeout_active", "bool", "r", {0, 1});
```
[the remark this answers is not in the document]
Ohisemega: Fixed

cpp/frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -92,6 +92,58 @@ FileWriterPlugin::FileWriterPlugin() :
         timeout_thread_running_(true),
         timeout_thread_(boost::bind(&FileWriterPlugin::run_close_file_timeout, this))
 {
+  std::string prefix = FileWriterPlugin::CONFIG_PROCESS + '/';
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_PROCESS_NUMBER, PMDD::UINT_T, PMDA::READ_WRITE, 1, PMD::MAX_UNSET);
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_PROCESS_RANK, PMDD::UINT_T, PMDA::READ_WRITE, 0, PMD::MAX_UNSET);
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_PROCESS_BLOCKSIZE, PMDD::UINT_T, PMDA::READ_WRITE, 1, PMD::MAX_UNSET);
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_PROCESS_BLOCKS_PER_FILE, PMDD::UINT_T, PMDA::READ_WRITE, 0, PMD::MAX_UNSET);
+  add_config_param_metadata(prefix + FileWriterPlugin::CONFIG_PROCESS_EARLIEST_VERSION, PMDD::BOOL_T, PMDA::READ_WRITE, {0, 1});
```
GDYendell: We shouldn't provide allowed values for bool. Could you update any others too.

PR #441, opened 2026-02-10
Implement RawFileWriter Plugin

### LINKED ISSUE #434: Convert TmpfsFrame into a plugin
During stream2 developments for for eiger, a new type of Frame was created that copys data to files in /dev/shm and uses that as the memory backing the Frame. This works OK, but would be more useful if it was just a plugin available for any detector.

Convert the TmpfsFrame from [here](https://github.com/odin-detector/odin-data/blob/c8e78833dbee1b4ef3df372295d93874fb232044/cpp/frameProcessor/src/TmpfsFrame.cpp) and the config + directory creation from eiger [here](https://github.com/DiamondLightSource/eiger-detector/blob/stream2/cpp/data/frameProcessor/src/EigerProcessPlugin.cpp#L166) into the new plugin.

The plugin should take config for
- enable/disable
- the path to the root directory to create files in - e.g. /dev/shm/eiger
- a field from the Frame to use as an additional nested directory - e.g. "series_id", default is empty to use root directory

Ohisemega:
Initial Commit
Add RawFileWriterPlugin *.cpp, *Lib.cpp and *.h files.

Fixes #434

### DIFF
--- a/cpp/frameProcessor/include/FrameProcessorPlugin.h
+++ b/cpp/frameProcessor/include/FrameProcessorPlugin.h
@@ -79,28 +79,63 @@ protected:
   void push(const std::string& plugin_name, boost::shared_ptr<Frame> frame);
 
   using ParameterMetadataMap_t = std::unordered_map<std::string, ParamMetadata>;
+  using All_val_vec_t = std::vector<ParamMetadata::allowed_values_t>;
+
+  // template type traits alias to enforce the DataType passed into function 
+  template<typename TYPE>
+  using is_all_vals_vec = std::is_same<typename std::decay<TYPE>::type, All_val_vec_t>;
+  template<typename TYPE>
+  using is_datatype     = std::is_same<typename std::decay<TYPE>::type, ParamMetadata::Datatype>;
+  template<typename TYPE>
+  using is_str_conv     = std::is_convertible<TYPE, std::string>; // can convert TYPE o std::string
+  template <typename DataType_, typename Allowed_>
+  using enableif_t                  = typename std::enable_if<(is_str_conv<DataType_>::value || is_datatype<DataType_>::value) && is_all_vals_vec<Allowed_>::value>::type;
+  template <typename DataType_>
+  using enableif_str_or_datatype_t  = typename std::enable_if<is_str_conv<DataType_>::value || is_datatype<DataType_>::value>::type;
+
+
   /**
    * Helper functions construct the hash_map's elements in-place
-   * \param[in] param: key for the ParamMetadata value-object. it is moved to trigger the move-semantics of the key-element in the pair
-   * \param[in] type, access_mode, allowed_values, min, max: arguments for ParamMetadata's constructor.They are not moved since the constructor
-   *                                                         of ParamMetadata accepts lvalues as references.
+   * \param[in] param: key for the ParamMetadata value-object. it is moved to trigger the move-semantics 
+   *                                    of the key-element in the pair
+   * \param[in] access_mode, min, max: arguments for ParamMetadata's constructor. They are not moved 
+   *                                    since the constructor of ParamMetadata accepts lvalues as references.
+   * \param[in] type:           For this argument, the template function ONLY allows std::string 
+   *                                    convertible types OR ParamMetadata::Datatype enum values to be passed im!
+   * \param[in] allowed_values: For this argument, the template function ONLY allows 
+   *                                    std::vector<ParamMetadata::allowed_values_t> types to be passed in!
+   * 
+   * Illustrative Usecase in YourPlugin.
+   * Where YourPlugin::CONFIG_PARAM[N] is your parameter's string:
+   * 
+   *    using PMD  = struct ParamMetadata;
+   *    using PMDA = PMD::AccessMode; // Alias for the AccessMode enums scope
+   *    using PMDD = PMD::Datatype;   // Alias for the Datatype enums scope
+   * 
+   ***  // 2nd overload is instantiated 'min', 'max' and 'allowed_values' ALL UNSET state
+   *    add_config_param_metadata(YourPlugin::CONFIG_PARAM1, PMDD::STRING_T, PMDA::READ_WRITE);
+   * 
+   ***  // 1st overload is instantiated NOTE that a custom string - ("DOUBLE_ARRAY"), was passed to DataType argument
+   *    // allowed_values argument is: {"align1", "align2"}.
+   *    add_config_param_metadata("MY_PLUGIN_PARAM_STRING", "DOUBLE_ARRAY", PMDA::READ_ONLY, {"align1", "align2"});
+   * 
+   ***  // 2nd overload is instantiated only 'min' is set to 1, 'max' and 'allowed_values' UNSET state
+   *    add_config_param_metadata(YourPlugin::CONFIG_PARAM3, PMDD::UINT_T, PMDA::READ_WRITE, 1); 
    */
-  
-  auto add_config_param_metadata(std::string param, std::string type, std::string access_mode, std::vector<ParamMetadata::allowed_values_t>& allowed_values)->void {
+  template <typename DataType, typename Allowed_ = All_val_vec_t, typename = enableif_t<DataType, Allowed_>>
+  auto add_config_param_metadata(std::string param, DataType type, ParamMetadata::AccessMode access_mode, Allowed_&& allowed_values)->void {
     config_metadata_.emplace(std::piecewise_construct,
                              std::forward_as_tuple(std::move(param)),
                              std::forward_as_tuple(type, access_mode, allowed_values, ParamMetadata::MIN_UNSET, ParamMetadata::MAX_UNSET)
     );
   }
 
-  auto add_config_param_metadata(std::string param, std::string type, std::string access_mode, std::vector<ParamMetadata::allowed_values_t>&& allowed_values)->void {
-    config_metadata_.emplace(std::piecewise_construct,
-                             std::forward_as_tuple(std::move(param)),
-                             std::forward_as_tuple(type, access_mode, allowed_values, ParamMetadata::MIN_UNSET, ParamMetadata::MAX_UNSET)
-    );
-  }
-
-  auto add_config_param_metadata(std::string param, std::string type, std::string access_mode, int32_t min, int32_t max)->void {
+  template <typename DataType, typename = enableif_str_or_datatype_t<DataType>>
+  auto add_config_param_metadata(std::string param, 
+                                  DataType type, 
+                                  ParamMetadata::AccessMode access_mode, 
+                                  int32_t min = ParamMetadata::MIN_UNSET, 
+                                  int32_t max = ParamMetadata::MAX_UNSET)->void {
     std::vector<ParamMetadata::allowed_values_t>allowed_vals{}; // This allows us to forward the vector as an lvalue reference to the constructor
     config_metadata_.emplace(std::piecewise_construct,
                              std::forward_as_tuple(std::move(param)),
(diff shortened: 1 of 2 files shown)


cpp/frameProcessor/include/TmpFSPlugin.h
```diff
@@ -0,0 +1,53 @@
...
+  const static std::string CONFIG_REMOVE_FILE;
+
+
+private:
+  /** Mutex used to make this class thread safe */
+  boost::recursive_mutex mutex_;
+  /** Pointer to logger */
+  LoggerPtr logger_;
+  /** Pointer to memory mapped file **/
+  // void* data_ptr_;
```
GDYendell: Remove

cpp/frameProcessor/include/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ SET(HEADERS DataBlock.h
             SumPlugin.h
             GapFillPlugin.h
             KafkaProducerPlugin.h
-            ParamMetadata.h)
+            ParamMetadata.h
+            TmpFSPlugin.h)
```
GDYendell:
```suggestion
            TmpfsPlugin.h)
```

The name comes from

```
❯ df
Filesystem                                                          1K-blocks        Used   Available Use% Mounted on
devtmpfs                                                             16129620           0    16129620   0% /dev
tmpfs                                                                16160520      719180    15441340   5% /dev/shm
```

(with a capital T because PascalCase)

cpp/frameProcessor/src/TmpFSPlugin.cpp
```diff
@@ -0,0 +1,108 @@
...
+TmpFSPlugin::TmpFSPlugin() : full_file_path_{""} {
+  // Setup logging for the class
+  logger_ = Logger::getLogger("FP.BloscPlugin");
+  LOG4CXX_TRACE(logger_, "TmpFSPlugin constructor.");
+}
+
+void TmpFSPlugin::process_frame(boost::shared_ptr<Frame> frame) {
+  // Protect this method
+  boost::lock_guard<boost::recursive_mutex> lock(mutex_);
+  LOG4CXX_DEBUG_LEVEL(3, logger_, "TmpFS received a new frame...");
```
GDYendell: This log message isn't useful - it will always be the same. Potentially it could be if it had a frame number, for example, but I don't think it will add much.

GDYendell (changes requested):
First pass

Ohisemega:
Personal Note:

Write **UNIT Test** for **_TempfsPlugin_**

cpp/frameProcessor/src/TmpfsPlugin.cpp
```diff
@@ -0,0 +1,108 @@
...
+  logger_ = Logger::getLogger("FP.BloscPlugin");
+  LOG4CXX_TRACE(logger_, "TmpFSPlugin constructor.");
+}
+
+void TmpFSPlugin::process_frame(boost::shared_ptr<Frame> frame) {
+  // Protect this method
+  boost::lock_guard<boost::recursive_mutex> lock(mutex_);
+  LOG4CXX_DEBUG_LEVEL(3, logger_, "TmpFS received a new frame...");
+
+  int fd = open(this->full_file_path_.c_str(), O_CREAT | O_RDWR, 0666);
```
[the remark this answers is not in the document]
Ohisemega: Thanks, I just understood your previous comment too.

cpp/frameProcessor/src/RawFileWriterPlugin.cpp
```diff
@@ -0,0 +1,104 @@
...
+    std::stringstream error;
+    error << "Failed to create directory: " << e.what();
+    throw std::runtime_error(error.str());
+  }
+  const std::string& acq_id = frame->get_meta_data().get_acquisition_ID();
+  long long fr_num = frame->get_frame_number();
+  std::string f_path = this->full_file_path_.string() + '/' + acq_id + '/' + std::to_string(fr_num);
+  int fd = open(f_path.c_str(), O_CREAT | O_RDWR, 0666);
+  if (fd == -1) {
+    throw std::runtime_error("Failed to open " + f_path);
```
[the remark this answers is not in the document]
ajgdls: @Ohisemega sorry to bring this up once again, but this process_frame logic sits outside of handleCtrlChannel that you mention above.  It is handled by the queues and callbacks, would you mind just checking that this exception will be caught somewhere in that process chain?

cpp/frameProcessor/src/RawFileWriterPlugin.cpp
```diff
@@ -0,0 +1,130 @@
...
+    } catch (boost::filesystem::filesystem_error& e) {
+      this->enabled_ = false;
+      std::stringstream error;
+      error << "Failed to create directory: " << file_path_.c_str() << ", Error:" << e.what();
+      this->set_error(error.str());
+      file_path_.clear();
+      throw std::runtime_error(error.str());
+    }
+  }
+}
```
ajgdls: This method should also provide a mechanism for enabling/disabling the plugin, as the enabled parameter is returned as a config item.

GDYendell: > Am I reflecting our discussion accurately?

Mostly yes, but that is not what the code is doing... Currently the user can't set enabled/disabled as you say they should be able to. And you have added enabled to `requestConfiguration`, just not `configure`. This is what Alan is pointing out.

We need something like:

```cpp
if (config.has_param(RawFileWriterPlugin::CONFIG_ENABLED)) {
  this->enabled_ = config.get_param<bool>(RawFileWriterPlugin::CONFIG_ENABLED);
}
...
```

cpp/frameProcessor/src/RawFileWriterPlugin.cpp
```diff
@@ -31,23 +35,27 @@ void RawFileWriterPlugin::process_frame(boost::shared_ptr<Frame> frame) {
...
-    std::array<char, 128> msg{};
+  if(write(fd, frame->get_data_ptr(), dsize) == -1){
+    ++this->dropped_frames_;
     strerror_r(errno, msg.data(), msg.max_size());
-    error << "Failed to Write to file - " << this->file_path_.string() <<": " << msg.data();
-    this->set_error(error.str());
-    throw std::runtime_error(error.str());
+    LOG4CXX_ERROR(logger_, "Failed to Write to file - " << this->file_path_.string() << ": " << msg.data());
+    this->push(frame);
+    return;
```
GDYendell: There is a lot of duplication here.
- Add the helper that logs an error with `errno`
- Call `push` once at the start of the function - we always push regardless of errors it doesn't need to be in every if

cpp/frameProcessor/src/RawFileWriterPlugin.cpp
ajgdls: Ah yes sorry @Ohisemega I did not explain enough in my comment.  The two methods `configure` and `requestConfiguration` should generally mirror each others configuration options - any configuration options of a plugin can be set in the `configure` method and read out in the `requestConfiguration` method.  There is no automatic mechanism for configuration items to be set, it must be explicitly coded as @GDYendell presents in the comment above.

cpp/frameProcessor/src/RawFileWriterPlugin.cpp
```diff
@@ -0,0 +1,122 @@
...
+  if(!this->enabled_){
+    ++this->dropped_frames_;
+    this->push(frame);
+    return;
+  }
+
+  const std::string& acq_id = frame->get_meta_data().get_acquisition_ID();
+  long long fr_num = frame->get_frame_number();
+  std::string f_path = this->file_path_.string() + '/' + acq_id + '/' + std::to_string(fr_num);
+  size_t dsize = frame->get_data_size();
```
GDYendell: You could just inline this. `get_data_size` just returns a member, so there is no gain by storing a local copy.

cpp/frameProcessor/src/RawFileWriterPlugin.cpp
```diff
@@ -0,0 +1,122 @@
...
+  std::array<char, 128> msg{};
+  if(!this->enabled_){
+    ++this->dropped_frames_;
+    this->push(frame);
+    return;
+  }
+
+  const std::string& acq_id = frame->get_meta_data().get_acquisition_ID();
+  long long fr_num = frame->get_frame_number();
+  std::string f_path = this->file_path_.string() + '/' + acq_id + '/' + std::to_string(fr_num);
```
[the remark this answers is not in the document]
Ohisemega: Thanks Gary,

I believe I have fixed all the issues with the string path as of now, but do kindly review and let me know.

cpp/frameProcessor/src/RawFileWriterPlugin.cpp
GDYendell: This is not done

cpp/frameProcessor/src/RawFileWriterPlugin.cpp
```diff
@@ -0,0 +1,131 @@
...
+  if(mkdir(f_path.c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IWOTH) == -1 && (errno != EEXIST)) {
+    this->enabled_ = false;
+    ++this->dropped_frames_;
+    strerror_r(errno, msg.data(), msg.max_size());
+    LOG4CXX_ERROR(logger_, "Failed to create directory: " << f_path << " errno: " << msg.data());
+    this->push(frame);
+    return;
+  }
+
+  long long fr_num = frame->get_frame_number();
```
GDYendell:
```suggestion
  long long frame_number = frame->get_frame_number();
```
fr_num could mean something very different in the context of odin

cpp/common/include/logging.h
```diff
@@ -7,7 +7,8 @@ namespace OdinData
 {
  #define LOG_WITH_ERRNO(logger, message) { \
   char msg[128]; \
-  LOG4CXX_ERROR(logger, message << " [errno: " << errno << " - " << strerror_r(errno, msg, sizeof(msg)) << ']'); \
+  const std::string& logging_msg = message; \
+  LOG4CXX_ERROR(logger, logging_msg << " [errno: " << errno << " - " << strerror_r(errno, msg, sizeof(msg)) << ']'); \
```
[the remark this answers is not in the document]
Ohisemega: Done

cpp/frameProcessor/src/RawFileWriterPlugin.cpp
```diff
@@ -0,0 +1,124 @@
...
+namespace FrameProcessor
+{
+
+const std::string RawFileWriterPlugin::CONFIG_FILE_PATH = "file_path";
+const std::string RawFileWriterPlugin::CONFIG_ENABLED = "writing_enabled";
+
+
+RawFileWriterPlugin::RawFileWriterPlugin() : file_path_{""}, dropped_frames_{0}, enabled_{false} {
+  // Setup logging for the class
+  logger_ = Logger::getLogger("FP.TempfsPlugin");
```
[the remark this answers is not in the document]
GDYendell: Yes please!

cpp/frameProcessor/src/RawFileWriterPlugin.cpp
```diff
@@ -0,0 +1,124 @@
...
+  // Protect this method
+  boost::lock_guard<boost::recursive_mutex> lock(mutex_);
+  this->push(frame);
+  if(!this->enabled_) {
+    return;
+  }
+  
+  const std::string& acq_id = frame->get_meta_data().get_acquisition_ID();
+  std::string&& full_file_path = this->file_path_.string() + '/' + acq_id + '/';
+  if(mkdir(full_file_path.c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IWOTH) == -1 && (errno != EEXIST)) {
```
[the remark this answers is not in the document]
Ohisemega: Hi Alan,
boost's `boost::filesystem::create_directory()` throws. We don't want the `process_frame()` method to handle `throw`-ing, as that is an operation with relatively high overhead (for a computational function like `process_frame()`), and we have to catch it somewhere on the call stack. So we are logging the errors and using `mkdir` to return `errno` faults.

PR #443, opened 2026-02-11
Fix TCP Socket Potential CPU Lock

### LINKED ISSUE #387: An error from read in FrameReceiverTCPRxThread can cause integer underflow in number of bytes received
An error return value of `read` is not currently handled by `FrameReceiverTCPRxThread::handle_receive_socket`. A return value of -1 to indicate an error can currently result in the number of bytes received to be 2**64 -1 due to underflow:

```
  size_t bytes_received = 0;

  while (bytes_received < message_size) {
    int msg_len =
        read(recv_socket_, frame_buffer, message_size - bytes_received);
    bytes_received += msg_len;
    frame_buffer += msg_len;
  }
```

This then tells the decoder that there are 18446744073709551615 bytes to process.

### LINKED ISSUE #433: TCP socket potential CPU lock
Whilst reviewing #392 I notice that there is a potential for a CPU core to get locked into a tight loop.  If the TCP socket fails (or is restarted) whilst reading bytes then the read call would return 0 bytes read.

This could result in a tight loop and locking up of a CPU core as the number of bytes read would never reach the total expected, and there is no timeout in the loop.

Ohisemega:
Add check for the return value of the network read() instruction in every while-loop iteration and break if less than 1

Fixes #433
Fixes #387


cpp/frameReceiver/src/FrameReceiverTCPRxThread.cpp
```diff
@@ -109,13 +109,20 @@ void FrameReceiverTCPRxThread::handle_receive_socket(int recv_socket_,
...
-    bytes_received += msg_len;
-    frame_buffer += msg_len;
+  while(((bytes_read = read(recv_socket_, frame_buffer, message_size - total_bytes_read)) > 0) && total_bytes_read < message_size) {
+    total_bytes_read += bytes_read;
+    frame_buffer += bytes_read;
   }
-  frame_decoder_->process_message(bytes_received);
+
+  if (bytes_read < 0) {
+        LOG4CXX_ERROR(logger_, "Error reading from socket in TCP RX thread");
```
timcnicholls: Inconsistent indentation in this block.

You could also decode `errno` to give the error more useful context?

timcnicholls (changes requested):

cpp/frameReceiver/src/FrameReceiverTCPRxThread.cpp
GDYendell: In eiger-detector we added this nice helper:
```cpp
#define LOG_WITH_ERRNO(log, message)                                             \
  std::stringstream error; \
  {                                                                              \
  error << message << " [errno: " << errno << " - " << strerror(errno) << "]"; \
    std::stringstream error;                                                     \
  LOG4CXX_ERROR(log, error.str()); \
    error << message << " [errno: " << errno << " - " << strerror(errno) << "]"; \
}
    LOG4CXX_ERROR(log, error.str());                                             \
  }
```

Ohisemega: Ah interesting, this is very helpful. As I have seen myself repeat the same errno error-reporting code lines in two files.  Would you prefer I add this to Odin-data @GDYendell ?

Ohisemega:
@GDYendell 

```
#define LOG_WITH_ERRNO(log, message) {  \
   std::stringstream error;  \
   error << message << " [errno: " << errno << " - " << strerror(errno) << "]";     \
   LOG4CXX_ERROR(log, error.str()); \
}
```
  
  Would you like me to add this code you referenced from Eiger-detector to odin-data?

timcnicholls:
> @timcnicholls Do you have any suggestions on handling the error state when `total_bytes_read < message_size`?Currently, I am simply early-returning from the function.

I was wondering that myself 😉 I had to think carefully if that clause in the code could ever be reached. I think it can, but only if the read is made non-blocking - TCP is a streaming protocol, so it is possible that the socket handler gets invoked when not all the data for a message is available on the socket receive queue (e.g. due to a slow source, network congestion etc) and read() returns zero (but not <0). 

I think a more robust implementation would a) make the `read()` call non-blocking (don't forget it's running in an `IpcReactor` which behaves like a select/poll call) and b) track the state of the size received across multiple calls of the handler, so that the message can be built in the buffer and then passed to `process_message` once the full message has been received. As it stands your implementation assumes that the entire message is available to it as soon as the handler is invoked I think.

Ohisemega:
> I think a more robust implementation would a) make the `read()` call non-blocking (don't forget it's running in an `IpcReactor` which behaves like a select/poll call)

Hi Time @timcnicholls, regarding this PR, would you like the non-blocking implementation to be implemented now, or do we merge this PR as-is and open another issue targeted at converting the blocking `read() `syscall to non-blocking? 

Also, I was thinking of the size of change required to implement a non-blocking version of this function.

Ohisemega:
> While the endless loop is avoided, I'm still not sure this implementation is complete... what happens if read() returns 0? The "we handle this case" clause is invoked and the call returns, but what happens then?

Ok, I will work on the non-blocking conversion. Thanks!

Ohisemega:
Progress Note: Need to convert `read()` call to `NON_BLOCK` version.

Ohisemega:
Could we merge this PR with your approvals? @JamesOHeaDLS and @ajgdls?
cc: @timcnicholls, we need your approval for this PR as well, please. As per Alan's comments below, I have also raised an issue regarding the non-blocking read on #507.

cpp/frameReceiver/src/FrameReceiverTCPRxThread.cpp
```diff
@@ -112,12 +112,25 @@ void FrameReceiverTCPRxThread::handle_receive_socket(int recv_socket_, int recv_
...
     void* frame_buffer = frame_decoder_->get_next_message_buffer();
     size_t message_size = frame_decoder_->get_next_message_size();
-    size_t bytes_received = 0;
+    ssize_t bytes_read = 0;
+    size_t total_bytes_read = 0;
+
+    unsigned char* fb_bytePtr = reinterpret_cast<unsigned char*>(frame_buffer
+    ); // reinterpret the void* address as an unsigned char pointer so we can do arithmetic
+    while (((bytes_read = read(recv_socket_, reinterpret_cast<void*>(fb_bytePtr), message_size - total_bytes_read)) > 0)
+           && total_bytes_read < message_size) {
```
JamesOHeaDLS: I think the order of the condition should be swapped so that 
`total_bytes_read < message_size`
is evaluated before read is called

This means read is only called when there is space

cpp/frameReceiver/src/FrameReceiverTCPRxThread.cpp
```diff
@@ -112,12 +112,25 @@ void FrameReceiverTCPRxThread::handle_receive_socket(int recv_socket_, int recv_
...
-        int msg_len = read(recv_socket_, frame_buffer, message_size - bytes_received);
-        bytes_received += msg_len;
-        frame_buffer += msg_len;
+    if (bytes_read < 0) {
+        std::array<char, 128> err_msg {};
+        strerror_r(errno, err_msg.data(), err_msg.max_size());
+        LOG4CXX_ERROR(logger_, err_msg.data());
+        return;
+    } else if (total_bytes_read < message_size) {
+        // We handle this case
```
JamesOHeaDLS: What is meant by "We handle this case"?

ajgdls: @Ohisemega can you add a log message here so that we know what has gone wrong if the code has entered this statement (however unlikely this is).

ajgdls (changes requested):

cpp/frameReceiver/src/FrameReceiverTCPRxThread.cpp
Ohisemega: I will add a log message for that condition.

ajgdls (approved):

JamesOHeaDLS (approved):

ajgdls:
Hi @timcnicholls we decided to limit the scope of this PR so that it can be merged to fix the reported issue and then raise another issue to tackle the non blocking and other raised points, so that this PR doesn't become stale.  There is no-one currently using this TCP reader plugin at DLS and so it is not so easily tested with hardware, and so lower priority than other tickets.  The merge is currently blocked due to your requested change, would you mind looking at this and either giving it the OK or comment to say we need to change this now before merging?  Thanks.

Ohisemega:
This PR requires Gary's review and approval. Given that he's moved on to another organisation, it might be better to create a new PR to address the issue on this one. #518 does that!

PR #457, opened 2026-02-27
Fix GCC's 'Deprecated' Warnings for std::iterator Tags in the RapidJson Header Lib.

### LINKED ISSUE #452: Fix build warnings on ubuntu 24 / C++17 standard
- Make minimal changes to odin-data code to print no warnings when compiling
- Add ignores / compiler flags for vendored code - e.g. RapidJson

Ohisemega:
Add GCC #pragma guard to ignore deprecated iterator tags for the C++17 build system.
This PR fixes >90% of the compile-time warnings Odin-data has. 
The only other warning is a pointer-arithmetic being done in the FrameReceiver source code. That should preferably be fixed by doing a `reinterpret_cast<unsigned char>(variable_name) `.

Fixes #452


Ohisemega:
cc: @ajgdls & @GDYendell

PR #462, opened 2026-03-09
Update BLOSC Configuration Data from Integers to Strings

### LINKED ISSUE #204: Update blosc settings to take strings instead of ints
See discussion on #203

Ohisemega:
Fixes #204


ajgdls (changes requested):

Ohisemega:
Hi @JamesOHeaDLS,
This PR isn't critical to the current project, but it's something we would like to get merged in. Kindly review at your convenience.

cpp/frameProcessor/src/BloscPlugin.cpp
```diff
@@ -82,21 +96,27 @@ BloscPlugin::BloscPlugin() :
...
     add_config_param_metadata(CONFIG_BLOSC_THREADS, PMDD::UINT_T, PMDA::READ_WRITE, 1);
     add_config_param_metadata(CONFIG_BLOSC_LEVEL, PMDD::UINT_T, PMDA::READ_WRITE, 1, 9);
     add_config_param_metadata(
-        CONFIG_BLOSC_SHUFFLE, PMDD::UINT_T, PMDA::READ_WRITE, { BLOSC_NOSHUFFLE, BLOSC_SHUFFLE, BLOSC_BITSHUFFLE }
+        CONFIG_BLOSC_SHUFFLE, PMDD::STRING_T, PMDA::READ_WRITE,
+        { BLOSC_NOSHUFFLE_STR, BLOSC_SHUFFLE_STR, BLOSC_BITSHUFFLE_STR }
+    );
+    add_config_param_metadata(
+        CONFIG_BLOSC_MODE, PMDD::STRING_T, PMDA::READ_WRITE,
+        { Mode_map::COMPRESS_MODE, Mode_map::DECOMPRESS_MODE, Mode_map::OFF_MODE }
```
JamesOHeaDLS: This is repeated below?

cpp/frameProcessor/src/BloscPlugin.cpp
```diff
@@ -372,46 +395,34 @@ void BloscPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMessage&
...
             this->update_compression_settings();
         }
     }
 
     if (config.has_param(BloscPlugin::CONFIG_BLOSC_COMPRESSOR)) {
-        unsigned int blosc_compressor = config.get_param<unsigned int>(BloscPlugin::CONFIG_BLOSC_COMPRESSOR);
-        if (blosc_compressor > BLOSC_ZSTD) {
+        std::string&& blosc_compressor = config.get_param<std::string>(BloscPlugin::CONFIG_BLOSC_COMPRESSOR);
+        // Range checking: 0, 1, 2, 3, 4, 5 are valid values. Anything else result in setting value  (no shuffle)
+        if (!compressor_str2i.count(blosc_compressor)) {
```
JamesOHeaDLS, line 384: Same here

cpp/frameProcessor/src/BloscPlugin.cpp
```diff
@@ -372,46 +395,34 @@ void BloscPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMessage&
     }
 
     if (config.has_param(BloscPlugin::CONFIG_BLOSC_SHUFFLE)) {
-        unsigned int blosc_shuffle = config.get_param<unsigned int>(BloscPlugin::CONFIG_BLOSC_SHUFFLE);
+        std::string&& blosc_shuffle = config.get_param<std::string>(BloscPlugin::CONFIG_BLOSC_SHUFFLE);
         // Range checking: 0, 1, 2 are valid values. Anything else result in setting value 0 (no shuffle)
-        if (blosc_shuffle > BLOSC_BITSHUFFLE) {
+        if (!shuffle_str2i.count(blosc_shuffle)) {
```
JamesOHeaDLS, line 360: count() here returns 0/1 based on key existence, so this is validating the compressor name rather than a numeric range. Comment could be tweaked to reflect this

eg Validate compressor name against lookup table. unordered_map::count() returns 1 if present, 0 otherwise.

cpp/frameProcessor/src/BloscPlugin.cpp
```diff
@@ -372,46 +395,34 @@ void BloscPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMessage&
...
 
     if (config.has_param(BloscPlugin::CONFIG_BLOSC_COMPRESSOR)) {
-        unsigned int blosc_compressor = config.get_param<unsigned int>(BloscPlugin::CONFIG_BLOSC_COMPRESSOR);
-        if (blosc_compressor > BLOSC_ZSTD) {
+        std::string&& blosc_compressor = config.get_param<std::string>(BloscPlugin::CONFIG_BLOSC_COMPRESSOR);
+        // Range checking: 0, 1, 2, 3, 4, 5 are valid values. Anything else result in setting value  (no shuffle)
+        if (!compressor_str2i.count(blosc_compressor)) {
             LOG4CXX_WARN(
                 logger_,
                 "Commanded blosc compressor: " << blosc_compressor << " is invalid. Setting compressor: " << BLOSC_LZ4
```
JamesOHeaDLS, line 387: I'm a bit confused by BLOSC_LZ4 here - is it a string?  How does it differ from blosc_compressor?

Should it still be used with reply.set_param below?

cpp/frameProcessor/test/BloscPluginTest.cpp
```diff
@@ -30,9 +32,11 @@ class BloscPluginTestFixture {
         // set compression level to the highest - 9
         BOOST_CHECK_NO_THROW(cfg_.set_param(FrameProcessor::BloscPlugin::CONFIG_BLOSC_LEVEL, 9));
         // set shuffle state to "BLOSC_BITSHUFFLE" - 2
-        BOOST_CHECK_NO_THROW(cfg_.set_param(FrameProcessor::BloscPlugin::CONFIG_BLOSC_SHUFFLE, 2));
+        BOOST_CHECK_NO_THROW(cfg_.set_param(FrameProcessor::BloscPlugin::CONFIG_BLOSC_SHUFFLE, std::string("shuffle")));
         // set the compressor to be - "zlib"
-        BOOST_CHECK_NO_THROW(cfg_.set_param(FrameProcessor::BloscPlugin::CONFIG_BLOSC_COMPRESSOR, 4));
+        BOOST_CHECK_NO_THROW(
+            cfg_.set_param(FrameProcessor::BloscPlugin::CONFIG_BLOSC_COMPRESSOR, std::string(BLOSC_ZLIB_COMPNAME))
```
JamesOHeaDLS: This feels a bit inconsistent - 

CONFIG_BLOSC_SHUFFLE is set with std::string("shuffle")

but

CONFIG_BLOSC_COMPRESSOR is set with std::string(BLOSC_ZLIB_COMPNAME)

Can BLOSC_SHUFFLE_STR be used instead of "shuffle"?

cpp/frameProcessor/src/BloscPlugin.cpp
Ohisemega: - `blosc_compressor` is the temporary string parameter provided by the user. It can be erratic.
- `BLOSC_LZ4` is the compressor **number or code** for the default compressor if the user provides a wrong name in `blosc_compressor`. It's for logging.
- `BLOSC_LZ4_COMPNAME` is the **name** of the default compressor if the user provides a wrong name in `blosc_compressor`. It's to set the default compressor and for logging too.

cpp/frameProcessor/include/BloscPlugin.h
```diff
@@ -46,6 +46,14 @@ class BloscPlugin : public FrameProcessorPlugin {
     static const std::string CONFIG_BLOSC_SHUFFLE;
     static const std::string CONFIG_BLOSC_MODE;
 
+    constexpr static char BLOSC_NOSHUFFLE_STR[] = "noshuffle";
+    constexpr static char BLOSC_SHUFFLE_STR[] = "shuffle";
+    constexpr static char BLOSC_BITSHUFFLE_STR[] = "bitshuffle";
+
+    constexpr static char COMPRESS_MODE[] = "compress";
+    constexpr static char DECOMPRESS_MODE[] = "decompress";
+    constexpr static char OFF_MODE[] = "off";
```
JamesOHeaDLS: Minor thing for consistency - make the modes end in _STR as well

eg COMPRESS_MODE_STR[]

Arguably they should start with BLOSC too

cpp/frameProcessor/src/BloscPlugin.cpp
```diff
@@ -279,15 +288,7 @@ void BloscPlugin::update_compression_settings()
...
-    ret = blosc_set_compressor(p_compressor_name);
-    if (ret < 0) {
-        LOG4CXX_ERROR(
-            logger_,
-            "Blosc failed to set compressor: " << " " << this->compression_settings_.blosc_compressor << " "
-                                               << *p_compressor_name
-        );
-        throw std::runtime_error("Blosc failed to set compressor");
-    }
+    blosc_set_compressor(p_compressor_name);
```
JamesOHeaDLS, line 279: Is error handling handled inside blosc_set_compressor?

cpp/frameProcessor/src/BloscPlugin.cpp
```diff
@@ -366,25 +364,20 @@ void BloscPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMessage&
             reply.set_param<std::string>("warning: level", "Capped at upper range: 9");
         }
         this->commanded_compression_settings_.compression_level = blosc_level;
-        if (this->current_acquisition_ == "") {
```
JamesOHeaDLS, line 369: I see that this code block has been removed in a few places - could you explain why please

cpp/frameProcessor/src/BloscPlugin.cpp
JamesOHeaDLS: Sorry I still think the comment is confusing - it reads as if it is validating against numerical value (ie 0, 1, 2), but if I understand the code correctly it is actually validating against the string.  .count is then returning 1 if the string is present within shuffle_str2i, or 0 if not

cpp/frameProcessor/src/BloscPlugin.cpp
JamesOHeaDLS: Thanks - that makes sense

cpp/frameProcessor/src/BloscPlugin.cpp
Ohisemega: The method `blosc_update_settings()` is called from the `configure()` function, where error checking is done. The `configure()` function is also the only function that modifies the `commanded_compression_settings_`. Hence, whatever goes into the `commanded_compression_settings_` **is already validated**.

cpp/frameProcessor/src/BloscPlugin.cpp
Ohisemega: Yes, it is redundant to keep updating the actual `compression_settings_` after obtaining each configuration. So I removed the duplicate checks and calls and simply put it at the end of `configure()`'s checks.

cpp/frameProcessor/src/BloscPlugin.cpp
Ohisemega: I agree with your review. I've modified it now. Hopefully, it's more explicit what's going on.

cpp/frameProcessor/src/BloscPlugin.cpp
```diff
@@ -136,40 +146,31 @@ BloscPlugin::~BloscPlugin()
...
-    size_t dest_data_size = c_settings.uncompressed_size + BLOSC_MAX_OVERHEAD;
+    size_t type_size = get_size_from_enum(src_frame->get_meta_data().get_data_type());
+    int uncompressed_size = src_frame->get_image_size();
+    size_t dest_data_size = uncompressed_size + BLOSC_MAX_OVERHEAD;
 
     boost::shared_ptr<Frame> dest_frame;
     try {
-        dest_frame = boost::make_shared<DataBlockFrame>(dest_meta_data, dest_data_size);
-
+        dest_frame = boost::make_shared<DataBlockFrame>(src_frame->get_meta_data(), dest_data_size);
```
JamesOHeaDLS, line 154: Previously this data was fetched via get_meta_data_copy(), whereas now it uses get_meta_data().  Is this intentional?

Ohisemega: Yes, this was to reduce the number of copies we create in the function. I returned a const reference, which could then be copied directly rather than creating a local copy and then copying the local copy again, while creating the destination frame buffer.

JamesOHeaDLS (approved):

Ohisemega:
Hi Alan, @ajgdls, this PR is ready for you to review at your convenience. Thank you!

Ohisemega:
Hi Alan @ajgdls ,
Just a kind reminder about this PR's review/approval.

ajgdls (approved):

PR #480, opened 2026-03-13
Added python tests to improve coverage.

ajgdls:


ajgdls:
Coverage for Python tests pushed above 80%

python/tests/test_ipc_tornado_channel.py
```diff
@@ -0,0 +1,45 @@
+from unittest import TestCase
```
JamesOHeaDLS: There is a typo in the filename for this test (tornao instead of tornado)

python/tests/test_ipc_client.py
```diff
@@ -0,0 +1,158 @@
...
+            msg = client._read_message(1.0)
+            assert msg.get_msg_val() == "test"
+            assert msg.get_msg_id() == 0
+        except IpcMessageException:
+            self.fail("send_request() raised IpcMessageException unexpectedly")
+
+        ipcChannel.poll.assert_called()
+        ipcChannel.recv.assert_called()
+
+    def test_check_for_valid_message(self, ipcChannel):
```
JamesOHeaDLS: This test triggers a warning for the source file which looks like an easy one to address:

> =============================== warnings summary ===============================
> python/tests/test_ipc_client.py::TestIpcClient::test_check_for_valid_message
>   /workspaces/odin-data/python/src/odin_data/control/ipc_client.py:47: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
>     self.logger.warn("Dropping reply message with id [" + str(id) + "] as was expecting [" + str(expected_id) + "]")

JamesOHeaDLS (changes requested):
Thanks Alan, this all looks good and runs nicely.

I have made comments for 2 very minor points.

A discussion point for another time - at the moment we seem to have a fairly even mix of unittest.TestCase style tests and pytest style tests.  This is most apparent with the assertEqual(a, b) vs assert a == b lines.  I think it would make sense to use a consistent approach across all tests.  pytest would be my preference.  I can raise this at the next core developers meeting.

ajgdls:
> Thanks Alan, this all looks good and runs nicely.
> 
> I have made comments for 2 very minor points.
> 
> A discussion point for another time - at the moment we seem to have a fairly even mix of unittest.TestCase style tests and pytest style tests. This is most apparent with the assertEqual(a, b) vs assert a == b lines. I think it would make sense to use a consistent approach across all tests. pytest would be my preference. I can raise this at the next core developers meeting.

Thanks @JamesOHeaDLS yeah I did debate a bit around this and changed a couple already.  I could convert to pytest pretty easily for some of them, so I'll see if it's easy to just switch them all over.

python/tests/test_ctrl_client.py
```diff
@@ -0,0 +1,159 @@
...
+@pytest.fixture
+def tmp_json_file():
+    """Creates a temporary JSON file and cleans it up after the test."""
+    tmp_file = tempfile.NamedTemporaryFile(mode="w", delete=False)
+    yield tmp_file
+    tmp_file_name = tmp_file.name
+    if not tmp_file.closed:
+        tmp_file.close()
+    if os.path.exists(tmp_file_name):
+        os.remove(tmp_file_name)
```
JamesOHeaDLS: This fixture could be removed by using the built-in pytest fixture tmp_path.  It will handle clean-up too

eg

```
def test_config_cmd(self, client, tmp_path):
    file = tmp_path / "config.json"
    file.write_text('{"test": "val"}\n')
    ...
```

python/tests/test_ctrl_client.py
```diff
@@ -0,0 +1,159 @@
...
+        new_id = client._next_msg_id()
+        assert new_id == 1
+        assert client._msg_id == 1
+
+    def test_shutdown(self, client):
+        client.args.config_file = None
+        client.args.request_config = False
+        client.args.request_commands = False
+        client.args.status = False
+        client.args.shutdown = True
```
JamesOHeaDLS: A lot of this client config gets repeated across many tests - maybe could be made into fixture too?

JamesOHeaDLS (commented):
Thanks @ajgdls - looks good and all tests run succesfully

I've just two suggestions for fixture tweaks

JamesOHeaDLS (approved):
Thanks Alan - looks good and all tests run and pass

Ohisemega (approved):

Issue #497, opened 2026-03-27
frameProcessor IpcMessages have invalid IpcMessage type


andyduller:
Status messages and shutdown messages (and possibly others), report "illegal" as the msg_type.

Issue #482, opened 2026-03-17
Recreate System Test Framework Based on Pytest


GDYendell:
Replace the current system test framework with a python test harness that spawns processes and sends individual commands over the zmq control channels to configure and run acquisitions, testing state throughout.

- Create a new `tests/` python test directory with a conftest.py
- Create fixtures for running up a frameReceiver, frameProcessor, frameSimulator
- Replicate current single shot acquisition with explicit configuration messages sent via the [client](https://github.com/odin-detector/odin-data/blob/master/python/src/odin_data/client.py) (these tests can install the odin_data python package to setup the environment, and then import the client)

Potentially then the logic for supporting single shot mode with automatic shutdown can be removed

From here, variations of tests should be easy to add
- Check order-dependence of setting file_path / frames - do values get copied over when acquisition id is changed
- Different data shapes / sizes
- Different acquisition rates
- A series of acquisitions with minimal overhead - rely on status to check when writing, done writing, ready for next acquisition, etc.
- Compress data with blosc and then write to hdf5
- Get a monitor image from the live view
- Test sending bad configs returns with a nice error message - bad file path, bad permissions, invalid values
- Give limited buffers and check dropped frames are reported

PR #502, opened 2026-03-27
#497: Set the frameProcessor reply msg_type to be ack.

andyduller:
Currently the msg_type is not set i.e. it defaults to "illegal". The #482 branch is probably the only thing that cares about this :-)

### DIFF
--- a/cpp/frameProcessor/src/FrameProcessorController.cpp
+++ b/cpp/frameProcessor/src/FrameProcessorController.cpp
@@ -118,6 +118,7 @@ void FrameProcessorController::handleCtrlChannel()
     try {
         OdinData::IpcMessage ctrlMsg(ctrlMsgEncoded.c_str());
         OdinData::IpcMessage replyMsg; // Instantiate default IpmMessage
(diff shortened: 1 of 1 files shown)


ajgdls (approved):
Patch coverage at 0% but I think writing a test for this needs to happen under a different PR.

andyduller:
Even with the line excluded as far as gcov is concerned, codecov is not happy. I've checked that gcov understands the GCOV_EXCLUDE_LINE, but codecov is ignoring it. I've not been able to find a way around this I'm afraid.

andyduller:
t may be possible to pip install codecov-cli (https://pypi.org/project/codecov-cli/) and see what codecov will do without doing a commit. It may be better to use another coverage tool e.g. lcov, which various posts seem to suggest. I've not used it before.

andyduller:
Don't think codecov-cli helps at all for this :-(

andyduller:
A test repository just to play with this might be the easiest way of finding out what it is really doing.

JamesOHeaDLS (approved):

PR #511, opened 2026-04-17
Refactor FrameMetaData Class

### LINKED ISSUE #510: Refactor FrameMetadata Class and Fix it's Issues
The `FrameMetadata` class has some issues that have been under the radar for a while:
- The member data `parameters_` has the type - `std::map<std::string, boost::any>` which allows any type to be the value of an element in the RB-tree map. This needs to be constrained by, say, a `boost::variant<...>`, which will allow us to specify a predefined, compile-time set of types.
- The `set_parameter(...)` is defined as: 

```
template <class T> void set_parameter(const std::string& parameter_name, T value)
    {
        parameters_[parameter_name] = value;
    }

```
is a template method with no constraint on what the template type - `T` should be.

- The `get_parameter(...)` method is defined as:

```
template <class T> T get_parameter(const std::string& parameter_name) const
    {
        std::map<std::string, boost::any>::const_iterator iter = parameters_.find(parameter_name);
        if (iter == parameters_.end()) {
            LOG4CXX_ERROR(logger, "Unable to find parameter: " + parameter_name);
            throw std::runtime_error("Unable to find parameter");
        }
        try {
            return boost::any_cast<T>(iter->second);
        } catch (boost::bad_any_cast& e) {
            LOG4CXX_ERROR(logger, "Parameter has wrong type: " + parameter_name);
            throw std::runtime_error("Parameter has wrong type");
        } catch (std::exception& e) {
            throw std::runtime_error("Unknown exception fetching parameter");
        }
    }
```
also has no constraints AND throws exceptions. This method is called in the `void HDF5File::write_parameter(...)`  method, which in turn is called in the `ProcessFrameStatus Acquisition::process_frame(...)` of the Acquisition class, and that in turn is called in `FileWriterPlugin::process_frame(...)`. We want to avoid `throw`-ing exceptions up the stack of `process_frame() `methods of plugins; therefore, the `throw` statements must be removed!

- This class also has no `move` constructor.

Ohisemega:
Refactor FrameMetaData:
        Make all methods inline in the header file.
        Delete `FrameMetaData.cpp`
        Modify the `parameter_` member data to an `unordered_map` of `std::string` & `boost::variant` pairs.
        Remove explicit exception throwing in the `get_parameter() `method.
        Add Rvalue reference overloads to `set_*()` methods.
        Add `type_traits` constraints to `get_parameter()` and `set_parameter()` overloaded methods.

Update tests and methods that utilised the `FrameMetaData` APIs that were modified.


ajgdls (commented):
1. You now use references and std::move where we were previously copying and declaring input parameters as const.  This means that those variables could now be potentially changed, have you confirmed that there is no possibility of introducing an error into the existing code by making these changes?
2. I would like to see this tested thoroughly on a detector deployment (test detector in the lab?) before we merge this to ensure we have confidence there aren't any unexpected behaviour changes.
3. Personally, I don't have any issue with the functions moving into the header although there are arguments for and against all header files.  I would like to ask @timcnicholls  to review this as well so that we can get his opinion on this.
4. Finally, are there some methods missing the noexcept keyword?

Ohisemega:
> 1. You now use references and `std::move` where we were previously copying and declaring input parameters as const.  This means that those variables could now be potentially changed, have you confirmed that there is no possibility of introducing an error into the existing code by making these changes?
> 
>     2. I would like to see this tested thoroughly on a detector deployment (test detector in the lab?) before we merge this to ensure we have confidence there aren't any unexpected behaviour changes.
> 
>     3. Personally, I don't have any issue with the functions moving into the header although there are arguments for and against all header files.  I would like to ask @timcnicholls  to review this as well so that we can get his opinion on this.
> 
>     4. Finally, are there some methods missing the noexcept keyword?

Hi Alan, @ajgdls, thank you for your review. I wanted to respond to and clarify a couple of the issues you raised.
1. For the functions which use `std::move(...)` they are actually overloads of the original functions, which I left unmodified (aside from adding the `noexcept` qualifiers where relevant). The 'move' overloads only match with calls that pass `RVALUE` arguments to the function as arguments. For instance:
```
set_acquisition_ID("FAMOUS"); // will call set_acquisition_ID(std::string&& str) i.e. std::move version
std::string my_str = "famous";
set_acquisition_ID(my_str); // will call set_acquisition_ID(const std::string& str);
```
4. The only functions not marked `noexcept` are those which can potentially `throw` exceptions. Typically, those not marked are functions that return a copy of a data structure that uses heap-memory, e.g. those that return `std::string` type objects by value (`std::string`) and NOT by reference (`std::string&`).

I'm also happy for the PR to be tested on a detector, but I'm pretty sure there is no functional change to the program's behaviour.

cpp/frameProcessor/include/FrameMetaData.h
```diff
@@ -64,73 +99,188 @@ class FrameMetaData {
...
+    }
 
     /** Return acquisition ID */
-    const std::string& get_acquisition_ID() const;
+    const std::string& get_acquisition_ID() const noexcept
+    {
+        return this->acquisition_ID_;
+    }
+
+    std::string get_acquisition_ID_c() const
```
ajgdls, line 212: noexcept can be applied to this one?

Ohisemega: It's returning an `std::string` object by value, therefore it can potentially `throw` an exception if memory cannot be allocated for the string. Its `this`-object pointer is qualified `const` because it does not modify any data member of the class.
If we mark a function that can actually `throw` exceptions as `noexcept`, the program will abort at runtime when an exception is thrown, even if there is a `catch` statement up the call stack.

ajgdls: I was comparing line 212 and 178.  One is marked noexcept and the other is not, and I cannot see the difference.

Ohisemega: Oh, I see ... that was an oversight on my part, I will fix it. The method on line 178 is NOT a `noexcept` method.

ajgdls: I can see the potential for problems with memory allocation resulting in a throw, and so I think actually I should have stated that the other method should not have noexcept.

More generally, if we run out of memory whilst returning copies of the acquisition ID or similar parameters, then we almost certainly have an underlying memory leak which is going to cause application termination at some point.  I think we need to be careful about how we approach updating the codebase in this way, to single files.  I will discuss this further with the collaboration.

PR #512, opened 2026-06-04
Update Odin-data's Python Adapter Metadata Request & Parse Mechanism

### LINKED ISSUE #342: Handling of allowed_values for enums
Putting this here as this is where the most necessary dev changes for interfacing with FastCS is being done, though could arguable be an issue for odin-control/ADOdin.

ADOdin IOCs can use MBBO records with a dropdown to select an enumerated value, and these can be given meaningful labels on EPICS, (e.g. for the detector trigger mode, or bit depth). The actual value sent between the AD driver and the Odin adapters is the integer value of the enum. In practice this means that if we provide the allowed_values metadata field it would have to look like ["0", "1", "2", "3"] for a record with 4 options, which provides no meaning for the values.

Since we want FastCS to generate records/screens directly from a ParameterTree with metadata, it could be helpful if allowed_values could either be reworked to provide a mapping between integer and string values, or if another metadata field could be provided with this mapping. Of course these parameters might be reworked to be strings with specific allowed values, but that would break compatibility with the ADOdin records.

### LINKED ISSUE #409: Update odin-data adapters to understand meta data responses
Implement the Python adapter changes so that they understand the meta data structure replies contained within FP message replies and can embed this correctly within the `ParameterTree` structure.

Ensure we can do a GET with `-H 'Accept: application/json;metadata=True'` and see the same response that we get from statically defined parameter trees ([https://github.com/odin-detector/odin-data/issues/348#issuecomment-2903774846]).

### LINKED ISSUE #527: Implementation of ParameterTree that Matches the Underlying C++ Structure.
1. The Python adapter controller performs a regular request for status and configuration from underlying applications.
2. For each reply, the structural update timestamp (formerly hash) will be checked.
3. If this is different to the previous timestamp (should always be the same or newer), then trigger a `metadata` request and completely rebuild the `ParameterTree`.
4. The controller can contain a cache member variable that the incoming `status`/`config` is assigned to.
5. When the `ParameterTree` is rebuilt, the `RW` attribute will need to be checked.  If `R`, then pass in a getter.  If `RW`, then pass in a `setter` and `getter`.
6. The `getter` should be wrapped using a `partial`, supplying the path to the leaf node.  This can be used within the `getter` to index into the cached member.
7. The `setter` (if required) can also be wrapped in a `partial`, and the path can be used to build the correct form of `IpcMessage` to send to the underlying C++ applications.
8. When a structural change occurs, throw away the `ParameterTree` and completely rebuild a new one with the new structure.

Ohisemega:
Update the adapter's controller to track the hash version of  the metadata
Fixes #409
Fixes #527
Fixes #342


python/src/odin_data/control/ipc_tornado_client.py
```diff
@@ -161,7 +169,7 @@ def _update_versions(self, version_msg):
 
         :param version_msg: Incoming version message response
         """
-        params = version_msg['params']
+        params = version_msg[IpcTornadoChannel.PARAMS_KEY]
```
JamesOHeaDLS: I think these IpcTornadoChannel references should be 'self' as that is where you have defined the new keys, etc.

python/src/odin_data/control/ipc_tornado_client.py
```diff
@@ -170,17 +178,27 @@ def _update_configuration(self, config_msg):
...
-        params = status_msg['params']
+        params = status_msg[IpcTornadoChannel.PARAMS_KEY]
         params['timestamp'] = status_msg['timestamp']
         self._parameters[self.IPC_VAL_STATUS] = params
+        self._parameters[IpcTornadoChannel.IPC_VAL_STATUS_METADATA_HASH] = status_msg[IpcTornadoChannel.METADATA_HASH_KEY]
+        if(self.METADATA_KEY in status_msg):
+            self._parameters[self.IPC_VAL_STATUS_METADATA] = status_msg[self.METADATA_KEY]
+        else:
+            self._parameters.pop(self.IPC_VAL_STATUS_METADATA, None)
+
```
JamesOHeaDLS: This is repeated code and could be moved into its own function?

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -8,12 +8,32 @@
...
-from odin.adapters.parameter_tree import ParameterTree
 
+from odin.adapters.parameter_tree import ParameterAccessor, ParameterTree
 from odin_data.control.ipc_tornado_client import IpcTornadoClient
 
 
+def recursive_splice(path=[], params_node={}, metadata={}):
+    if isinstance(params_node, dict):
+        for k, v in params_node.items():
+            recursive_splice(path.append(str(k)), v, metadata)
```
JamesOHeaDLS: append doesn't return the appended list so can't be used as an argument into recursive_splice

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -8,12 +8,32 @@
...
+        return {
+            k: recursive_splice(path + [k], v, metadata) for k, v in params_node.items()
+        }
```
JamesOHeaDLS: This looks like the above implemented again with the path append correction - was this meant to be in place of the above?

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -8,12 +8,32 @@
...
 
+from functools import reduce
 from deepdiff import DeepDiff
-from odin.adapters.parameter_tree import ParameterTree
 
+from odin.adapters.parameter_tree import ParameterAccessor, ParameterTree
 from odin_data.control.ipc_tornado_client import IpcTornadoClient
 
 
+def recursive_splice(path=[], params_node={}, metadata={}):
```
JamesOHeaDLS: Default values are created when the function is defined, not each time it’s called, so these variables can be shared between calls to the function

python/src/odin_data/control/odin_data_controller.py
JamesOHeaDLS: This looks incomplete

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -129,36 +153,59 @@ def update_loop(self):
...
                         try:
-                            msg = client.send_request(parameter_tree)
+                            if(param_req == IpcTornadoClient.IPC_VAL_REQ_CFG and self.config_metadata_hash != self.config_metadata_hash_prev):
+                                msg = client.send_request(param_req, True)
+                                self.config_metadata_hash_prev = self.config_metadata_hash
+                            elif(param_req == IpcTornadoClient.IPC_VAL_STATUS and self.status_metadata_hash != self.status_metadata_hash_prev):
+                                msg = client.send_request(param_req, True)
+                                self.status_metadata_hash_prev = self.status_metadata_hash
+                            else:
+                                msg = client.send_request(param_req)
```
JamesOHeaDLS: I think this if statement could be simplified as the responses are so similar

JamesOHeaDLS (changes requested):
Hi @Ohisemega 

I've added my comments.  I think we also really needs an associated set of tests to check that these tree interactions are doing what we expect

python/src/odin_data/control/ipc_tornado_client.py
Ohisemega: Thank you, that was an oversight on my part.

python/src/odin_data/control/ipc_tornado_client.py
Ohisemega: Hi @JamesOHeaDLS,
Regarding this point you raised, I believe you meant that it duplicated lines 182-186, but they use different keys to index the message dictionary:
Compare the above to these lines (181-186) in `_update_configuration(self, config_msg)` - 
```
        self._parameters[self.IPC_VAL_CONFIG] = config_msg[self.PARAMS_KEY]
        self._parameters[self.IPC_VAL_CONFIG_METADATA_HASH] = config_msg[self.METADATA_HASH_KEY]
        if(self.METADATA_KEY in config_msg):
            self._parameters[self.IPC_VAL_CONFIG_METADATA] = config_msg[self.METADATA_KEY]
        else:
            self._parameters.pop(self.IPC_VAL_CONFIG_METADATA, None)
```

python/src/odin_data/control/odin_data_controller.py
Ohisemega: Typographical error, will fix!

python/src/odin_data/control/odin_data_controller.py
Ohisemega: Ok removed!

python/src/odin_data/control/odin_data_controller.py
Ohisemega: That's Alan's work; I can't make much comment on it at the moment.
cc: @ajgdls, Hi Alan, could you kindly comment on what you had in mind for implementing error logs? Is it a vector of error string as in the Odin-data's FrameProcessorPlugin class `set_error()` function?

python/src/odin_data/control/odin_data_controller.py
Ohisemega: Removed the for-loop.

Ohisemega:
> Hi @Ohisemega
> 
> I've added my comments. I think we also really need an associated set of tests to check that these tree interactions are doing what we expect

I've appended a pytest case for that, thank you for the prompt!

LuisFSegalla (commented):
It's looking good to me.
My only suggestion would be to maybe add some comments explaining the necessity of the double hash tracking for the metadata calls.

python/src/odin_data/control/odin_data_controller.py
ajgdls: Hi @JamesOHeaDLS @Ohisemega , indeed as part of the Python coverage PR #480 the unit tests uncovered this bug and it is resolved in that PR.

Ohisemega: Thanks, Alan. We'll leave it to be resolved in #480 then.

LuisFSegalla (approved):

python/src/odin_data/control/ipc_tornado_client.py
```diff
@@ -170,17 +178,23 @@ def _update_configuration(self, config_msg):
...
+        if(self.METADATA_KEY in config_msg):
+            self._parameters[self.IPC_VAL_CONFIG_METADATA] = config_msg[self.METADATA_KEY]
 
     def _update_status(self, status_msg):
         """Store the response to a status message in the _parameters dict.
 
         :param status_msg: Incoming status message response
         """
-        params = status_msg['params']
+        params = status_msg[self.PARAMS_KEY]
```
JamesOHeaDLS, line 204: As there is so much similarity between update_config and update_status I wonder whether the majority of their functionality could be combined into one function.

There should be an associated test, which can also help with combining into one function

python/tests/test_control.py
```diff
@@ -0,0 +1,52 @@
+from odin_data.control.odin_data_controller import splice_params_metadata
+
+class TestControl:
```
JamesOHeaDLS: I really like how you have structured this test - it makes it very readable and understandable

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -129,36 +154,60 @@ def update_loop(self):
...
+                        self._params.replace(f"{index}/{IpcTornadoClient.IPC_VAL_STATUS}", status_resp)
+                    if IpcTornadoClient.IPC_VAL_CONFIG in client.parameters:
+                        if(IpcTornadoClient.CONFIG_PARAMS_KEY in client.parameters[IpcTornadoClient.IPC_VAL_CONFIG]):
+                            config_resp = client.parameters[IpcTornadoClient.IPC_VAL_CONFIG][IpcTornadoClient.CONFIG_PARAMS_KEY]
+                        if(IpcTornadoClient.IPC_VAL_CONFIG_METADATA_HASH in client.parameters):
+                            self.config_metadata_hash = client.parameters[IpcTornadoClient.IPC_VAL_CONFIG_METADATA_HASH]
+                        if (IpcTornadoClient.IPC_VAL_CONFIG_METADATA in client.parameters):
+                            metadata = client.parameters[IpcTornadoClient.IPC_VAL_CONFIG_METADATA]
+                            config_resp = splice_params_metadata(config_resp, metadata)
+                        self._params.replace(f"{index}/{IpcTornadoClient.IPC_VAL_CONFIG}", config_resp)
```
JamesOHeaDLS: Although the keys differ the structure repeats so I think this should be turned into a function or a for loop where the parameters are fed in.  This will make it much easier to parse by eye

Ohisemega: Would you prefer that I lift the common if-statements into a method which can also switch between config and status?

JamesOHeaDLS: Yes, I was thinking of a function along the lines of

```
def_update_params_with_metadata(
    self,
    client,
    index,
    value_key,
    metadata_key,
    metadata_hash_key,
    hash_attr)
    ...
```

where you pass in the differing params and then the body is just the common logic

python/src/odin_data/control/odin_data_controller.py
ajgdls: We shouldn't need to perform any DeepDiff checks now.  When the ParameterTree gets rebuilt  on a structure change update it should also rebuild the command execution callbacks at that time.

Basically, the commands would never change unless there was a structural change and so you only need to rebuild the command logic when you are rebuilding the entire ParameterTree, and I don't think it needs to be checked on every update loop.

ajgdls (changes requested):

python/src/odin_data/control/odin_data_controller.py
ajgdls: @Ohisemega the adapter is working well now, but I think this bit of logic can be cleaned up.

Ohisemega: Hi @ajgdls,
Are you saying the `parse_available_commands()` function is redundant?

Ohisemega: A change to the command structure will only happen if a new plugin is loaded.
So I added a boolean flag that will be set to `True` if the `status`/`config` structure changes in the Odin-Control application.
When the next commands are received, it will rebuild, so essentially commands update when `config`/`status` updates. I have pushed the change.
In light of that change, I also removed the `Deepdiff` function call.

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -93,13 +104,70 @@ def get(self, path, meta):
...
+                # Safe fallback: keeps the data, flags missing metadata
+                return (params_node, None)
+
+    def splice_params_metadata(self, index, resp_type:str, params:dict, metadata:dict):
+        path = []
+        params = self.recursive_splice(index, resp_type, path, params, metadata)
+        return params
+
+    def _update_params_with_metadata(self, value_dict:dict, index:int, value_key:str, param_key:str, metadata_key:str, metadata_ts_key:str, client: IpcTornadoClient):
+        # NB: 'STATUS' can be replace with 'CONFIG' in the following comments:
```
timcnicholls: This comment block is not a docstring and its intent is unclear.

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -123,81 +191,86 @@ def update_loop(self):
...
-                        "request_commands",
+                    for param_req in [
+                        IpcTornadoClient.IPC_VAL_STATUS,
+                        IpcTornadoClient.IPC_VAL_REQ_CFG,
+                        IpcTornadoClient.IPC_VAL_REQ_CMDS,
                     ]:
                         try:
-                            msg = client.send_request(parameter_tree)
+                            with_metadata = False
+                            # Check if the previous values of the config and status metadata hash matches the latest value.
```
timcnicholls: Multiple references to `hash` in comments that should refer to the timestamp now instead?

timcnicholls (changes requested):
It's quite hard to follow the logic in this change, even understanding the timestamp / metadata refresh mechanism. Some more comments might help to clarify this.

I've picked up a few minor specifics, otherwise it seems to be as intended.

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -7,13 +7,21 @@
 import logging
 import threading
 import time
+from functools import partial, reduce
 
 from deepdiff import DeepDiff
```
ajgdls: Remove deepdiff dependency in any project file and also remove this import statement.

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -93,13 +100,69 @@ def get(self, path, meta):
...
+        params = self.recursive_splice(index, resp_type, path, params, metadata)
+        return params
+
+    #   NB: '<PT>' represents the "Parameter Type" and can be either
+    #       'CONFIG' or 'STATUS'.
+    #   - IpcTornadoClient.<PT>_PARAMS_KEY == param_key
+    #   - client.parameters[IpcTornadoClient.IPC_VAL_<PT>] == value_dict
+    #   - IpcTornadoClient.IPC_VAL_<PT> == value_key
+    #   - IpcTornadoClient.IPC_VAL_<PT>_TS == metadata_ts_key
+    #   - IpcTornadoClient.IPC_VAL_<PT>_METADATA == metadata_key
```
ajgdls: Docstring should be immediately after the function definition, and use 3 quotes ?

ajgdls: Maybe this isn't supposed to be a docstring, but then I'm not sure this comment helps.

ajgdls (changes requested):
I think @timcnicholls raised a valid point.  Even though the PR was not specifically raised to tackle documentation, it would be a worthwhile exercise to ensure we have good docstring comments for each function in this file as well as in the class, and maybe some additional comments for any complex logic.  This is a relatively complex adapter controller and good documentation now whilst it is fresh is a good idea for the future.
@Ohisemega if you would like me to contribute to the comments (as I haven't written all of the functions it might be useful for another perspective) let me know and I can do that.

Ohisemega:
> I think @timcnicholls raised a valid point. Even though the PR was not specifically raised to tackle documentation, it would be a worthwhile exercise to ensure we have good docstring comments for each function in this file as well as in the class, and maybe some additional comments for any complex logic. This is a relatively complex adapter controller and good documentation now whilst it is fresh is a good idea for the future. @Ohisemega if you would like me to contribute to the comments (as I haven't written all of the functions it might be useful for another perspective) let me know and I can do that.

Hi Alan, @ajgdls, yes, I'm OK with you writing some documentation strings for the functions. I will review the information too.

python/src/odin_data/control/odin_data_controller.py
Ohisemega: I see what you mean. I will delete the comment and put a proper docstring explaining the function of each argument.

python/src/odin_data/control/odin_data_controller.py
Ohisemega: I have appended a docstring.

Ohisemega:
Thanks for the docstrings @ajgdls. They describe the functions very well and as intended!

ajgdls (changes requested):
Hi @Ohisemega this is now close.  I noticed that command execution is not quite working as I expected though.  I think the current mechanism does this:
put is called on a command execution.
execute callback puts the name of the command and the plugin on the callback queue.

I see no mechanism that results in the queued commands being sent down to the C++ applications.  We can discuss this today at our meeting, it may be that now we do not need to queue commands any longer and can instead simply send them straight through to the C++ applications.

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -7,23 +7,64 @@
...
         self._error = ""
+        self.config_ts: list[int] = [-1] * len(endpoints)
+        self.status_ts: list[int] = [-1] * len(endpoints)
+        self.config_ts_prev: list[int] = [0] * len(endpoints)
+        self.status_ts_prev: list[int] = [0] * len(endpoints)
         self._endpoints = []
-        self._config_cache = None
+        self._command_needs_update: bool = False
+        self._config_resposes: list[dict] = [None] * len(endpoints)
+        self._status_resposes: list[dict] = [None] * len(endpoints)
```
JamesOHeaDLS: Variable name typos - resposes -> responses

Also the type hinting should be list[dict | None] as it is intialised with None

Ohisemega:
> Hi @Ohisemega this is now close. I noticed that command execution is not quite working as I expected though. I think the current mechanism does this: put is called on a command execution. execute callback puts the name of the command and the plugin on the callback queue.
> 
> I see no mechanism that results in the queued commands being sent down to the C++ applications. We can discuss this today at our meeting, it may be that now we do not need to queue commands any longer and can instead simply send them straight through to the C++ applications.

I've rectified this, thank you!

JamesOHeaDLS:
Now that DeepDiff is no longer used in the code it can be removed as a dependency in pyproject.toml

ajgdls:
Hi @Ohisemega I have been trying to get this latest version to work, but I think there is an issue.  You are calling `self.can_execute` before executing but I cannot find this method and I am seeing the following error....
```
     File "/workspaces/odin-data/python/src/odin_data/control/odin_data_controller.py", line 387, in send_command
        if self.can_execute(index, plugin, value):
           ^^^^^^^^^^^^^^^^
    AttributeError: 'FrameProcessorController' object has no attribute 'can_execute'
```

ajgdls:
It now works!

ajgdls (approved):

Ohisemega:
@timcnicholls Could you please do another review of this PR and approve if it's OK?

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -7,23 +7,64 @@
...
-from deepdiff import DeepDiff
-from odin.adapters.parameter_tree import ParameterTree
+from odin.adapters.parameter_tree import ParameterAccessor, ParameterTree
 
 from odin_data.control.ipc_tornado_client import IpcTornadoClient
 
 
+def setter_func(tornado_client: IpcTornadoClient, path: list, value):
+    """
+    Send a configuration message to a C++ odin-data application with the parameter name
```
timcnicholls: The 'C++' in the comment is redundant

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -93,13 +139,117 @@ def get(self, path, meta):
...
+            except (KeyError, TypeError):
+                # Safe fallback: keeps the data, flags missing metadata
+                return (params_node, None)
+
+    def splice_params_metadata(
+        self, index, resp_type: str, params: dict, metadata: dict
+    ):
+        """
+        Recursive function to append metadata of each parameter to it's value together in a tuple.
+        This is the format the ParameterTree expects it.
```
timcnicholls, line 205: Missing blank line before param list

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -93,13 +139,117 @@ def get(self, path, meta):
...
+    def _update_params_with_metadata(
+        self,
+        value_dict: dict,
+        index: int,
+        param_key: str,
+        metadata_key: str,
+        metadata_ts_key: str,
+    ):
+        """
+        dict - The dictionary containing the status or config_request response.
```
timcnicholls, line 228: No docstring comment

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -123,109 +273,118 @@ def update_loop(self):
...
 
-    def queue_command(self, index, plugin, value):
-        """Called for each command PUT that is received by the adapter
+    def send_command(self, index, plugin, value):
+        """Called for each command from parse_available_commands
         PUT URI is of the form index/command/plugin/execute and the
         value is the name of the command to execute.
-        This method simply queues commands for execution after any configuration
-        changes have been applied.
+        This method sends each commands as demanded.
```
timcnicholls: Missing param list in docstring. Docstring formatting has short comment lines

timcnicholls (commented):
Looks good! I haven't tested its behaviour with running odin-data applications, but I note that @ajgdls has done this.

Some minor comments about docstring formatting and conventions. The [PEP257](https://peps.python.org/pep-0257/) conventions that can be checked by linters (ruff etc.) are useful for this. `odin-control` has the appropriate settings for ruff in its [pyproject.toml](https://github.com/odin-detector/odin-control/blob/main/pyproject.toml)

timcnicholls (approved):

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -203,6 +203,7 @@ def splice_params_metadata(
         """
         Recursive function to append metadata of each parameter to it's value together in a tuple.
```
JamesOHeaDLS: Typo - it's --> its

python/src/odin_data/control/odin_data_controller.py
```diff
@@ -374,10 +379,12 @@ def parse_available_commands(self, index, client):
         self._supported_commands[index] = client.parameters["commands"]
 
     def send_command(self, index, plugin, value):
-        """Called for each command from parse_available_commands
-        PUT URI is of the form index/command/plugin/execute and the
-        value is the name of the command to execute.
-        This method sends each commands as demanded.
+        """
+        This method sends each commands as demanded as an IpcMessage.
```
JamesOHeaDLS: Typo - commands ---> command

JamesOHeaDLS (commented):
2 tiny typos

JamesOHeaDLS (approved):
Thanks @Ohisemega

PR #514, opened 2026-06-05
Add Metadata Hash of Parameter Structure to JSON Response from Odin-data

### LINKED ISSUE #513: Add Hash-id Version Stamp to Configuration and Status Request Replies
odin-data C++ server needs to append a timestamp field to its JSON response, which will only be updated if there is a change to its configuration/status metadata structure. The timestamp will then be validated by odin-control.

Ohisemega:
- Implement an `IpcMessage::get_keys()` method to return a delimited string of all keys in the "params" node of the response. 
- Add last_hash_*_metadata_ for status and configuration requests. Implement logic to hash the parameter keys and update the last_hash_ value if changed.
- Add the last_hash_*_metadata_ value to the JSON response.

Fixes #513

### DIFF
--- a/cpp/common/include/IpcMessage.h
+++ b/cpp/common/include/IpcMessage.h
@@ -477,7 +477,7 @@ private:
     // helper function to append
     static void get_keys_helper(const rapidjson::Value& value, const char delim, std::string& result)
     {
-        if (value.IsObject()) {
+        if (__builtin_expect(value.IsObject(), 1)) {
             for (const auto& member : value.GetObject()) {
                 // Store the found key
                 result.append(member.name.GetString()) += delim;
--- a/cpp/common/include/IpcMessage.h
+++ b/cpp/common/include/IpcMessage.h
@@ -272,10 +272,11 @@ public:
     }
 
     // returns all the key-strings of param_str as a single 'delim' delimited string
-    std::string get_keys(const std::string& param_str = "", const char delim = '/')
+    std::string get_keys(const std::string& param_str, const char delim = '/')
     {
         std::string keys_str;
-        keys_str.reserve(256);
+        constexpr size_t string_size = 256;
+        keys_str.reserve(string_size);
         if (this->doc_.FindMember(param_str.c_str()) == doc_.MemberEnd())
             return "";
         rapidjson::Value& params = this->doc_[param_str.c_str()];
@@ -474,7 +475,7 @@ private:
     //! Maps an internal message timestamp representation to an ISO8601 extended format string
     std::string valid_msg_timestamp(boost::posix_time::ptime msg_timestamp);
 
-    // helper function to append
+    // helper function to recursively append key-strings in 'value' to the 'result' variable
     static void get_keys_helper(const rapidjson::Value& value, const char delim, std::string& result)
     {
         if (__builtin_expect(value.IsObject(), 1)) {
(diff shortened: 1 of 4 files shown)


cpp/common/include/IpcMessage.h
```diff
@@ -271,6 +271,18 @@ class IpcMessage {
         }
     }
 
+    // returns all the key-strings of param_str as a single 'delim' delimited string
+    std::string get_keys(const std::string& param_str = "", const char delim = '/')
+    {
+        std::string keys_str;
+        keys_str.reserve(256);
```
ajgdls: Is this magic number a good representation of what gets returned?  I expect you want to reserve and thus save continued reallocation as the string grows which is a good idea.  I would like to see the number defined with a meaningful name rather than using the magic constant.

cpp/common/include/IpcMessage.h
```diff
@@ -462,6 +474,25 @@ class IpcMessage {
     //! Maps an internal message timestamp representation to an ISO8601 extended format string
     std::string valid_msg_timestamp(boost::posix_time::ptime msg_timestamp);
 
+    // helper function to append
```
ajgdls: Update this comment to contain more information.

ajgdls (changes requested):
Would you mind adding a unit test for the value walking in the IpcMessage class?

cpp/common/include/IpcMessage.h
Ohisemega: You're right, though, also thinking of it, there's not much point in using the reserve functionality. But I'll fix the 'magic number' issue.

Ohisemega:
> Would you mind adding a unit test for the value walking in the IpcMessage class?

OK, I'll create a "test" directory for the common folder, and `IpcMessage` would be the first.

Ohisemega:
> > Would you mind adding a unit test for the value walking in the IpcMessage class?
> 
> OK, I'll create a "test" directory for the common folder, and `IpcMessage` would be the first.

On review, it seems the common library modules have a unit test suite, but this is present in the `FrameReceiver` folder, and it needs to be moved to a 'test folder' in the 'common' directory.
Edit: I have created an issue ticket (#516) to move the common-directory modules unit test suites to the common folder.

ajgdls (approved):

cpp/frameReceiver/test/IpcMessageUnitTest.cpp
```diff
@@ -384,4 +384,31 @@ BOOST_AUTO_TEST_CASE(TestIpcMessageCreationSpeed)
...
+                           "                   \"last\": \"data\""
+                           "               },"
+                           "               \"dob\": 2011"
+                           "          }"
+                           "      }"
+                           "}";
+    OdinData::IpcMessage test_msg(json_str, true);
+    BOOST_CHECK(test_msg.is_valid() == true);
+    std::string keys_list = test_msg.get_keys("params");
+    std::cout << keys_list << '\n';
```
ajgdls: Remove this print statement

ajgdls (changes requested):

Ohisemega:
Hi @JamesOHeaDLS, could you kindly review at your convenience?

ajgdls (approved):

JamesOHeaDLS (approved):
Thanks @Ohisemega - looks good and test passes

PR #529, opened 2026-07-03
Replace Metadata Hash With a 'Latest Version Timestamp'

### LINKED ISSUE #526: Replace metadata hash with a timestamp of structural change.
It has been decided that the following logic would be preferred to a hash of keys when deciding if a structural change has taken place.

1. Plugin parent class contains a timestamp private member variable.
2. The timestamp is updated upon plugin construction (at the end of the constructor).
3. Plugin parent class supports a protected method(?) that can be called to update the timestamp to "now".
4. Plugin parent class inserts this timestamp into any config or status request response.
5. When the controller classes receive these timestamps it passes back the latest timestamp as the "hash".

The python adapter controller will be able to check the timestamp against the previous and trigger a rebuild in the same way as it did previously with the hash.

Many plugins will simply return the timestamp created upon the plugin construction.  Some (FileWriter) will update the timestamp by calling the protected method when certain configurations have been updated, eg new datasets created.

Ohisemega:
Implement a time-stamping feature in plugins.

- Update the timestamp at locations of structural change in the plugin. Report the latest timestamp in `requestConfiguration()` and `provideStatus()`.
- Remove 'key-hashing' related implementation and tests. 

Fixes #526

### DIFF
--- a/cpp/frameProcessor/include/FrameProcessorPlugin.h
+++ b/cpp/frameProcessor/include/FrameProcessorPlugin.h
@@ -73,30 +73,30 @@ public:
     void remove_callback(const std::string& name);
     void remove_all_callbacks();
     void notify_end_of_acquisition();
-    static constexpr char METADATA_VERSION[] = "metadata_version";
+    /** return the latest ts of config/status metadata for this plugin instance */
+    inline int64_t get_config_metadata_ts() const
+    {
+        return this->config_metadata_ts_;
+    }
+    inline int64_t get_status_metadata_ts() const
+    {
+        return this->status_metadata_ts_;
+    }
 
 protected:
     /** function to update the config time-stamp! */
     inline void update_config_metadata_version(void)
     {
-        auto time_stamp_ms = std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::system_clock::now());
+        auto time_stamp_ms = std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now());
         auto val = std::chrono::duration_cast<std::chrono::milliseconds>(time_stamp_ms.time_since_epoch());
         this->config_metadata_ts_ = val.count();
     }
     /** function to update the status time-stamp! */
     inline void update_status_metadata_version(void)
     {
-        auto time_stamp_ms = std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::system_clock::now());
+        auto time_stamp_ms = std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now());
         auto val = std::chrono::duration_cast<std::chrono::milliseconds>(time_stamp_ms.time_since_epoch());
         this->status_metadata_ts_ = val.count();
-    }
-    inline int64_t get_config_metadata_version()
-    {
-        return this->config_metadata_ts_;
-    }
-    inline int64_t get_status_metadata_version()
-    {
-        return this->status_metadata_ts_;
     }
     void push(boost::shared_ptr<Frame> frame);
     void push(const std::string& plugin_name, boost::shared_ptr<Frame> frame);
@@ -267,7 +267,7 @@ private:
     CallDuration process_duration_;
     /** time-stamp to represent the version of config metadata */
     int64_t config_metadata_ts_;
-    /** time-stamp to represent the version of config metadata */
+    /** time-stamp to represent the version of status metadata */
     int64_t status_metadata_ts_;
 };
 
--- a/cpp/frameProcessor/include/FrameProcessorPlugin.h
+++ b/cpp/frameProcessor/include/FrameProcessorPlugin.h
@@ -73,8 +73,20 @@ public:
     void remove_callback(const std::string& name);
     void remove_all_callbacks();
     void notify_end_of_acquisition();
+    static constexpr char METADATA_VERSION[] = "metadata_version";
 
 protected:
+    /** function to update the config time-stamp! */
+    inline void update_config_metadata_version(void)
+    {
+        auto time_stamp_ms = std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::system_clock::now());
+        auto val = std::chrono::duration_cast<std::chrono::milliseconds>(time_stamp_ms.time_since_epoch());
+        this->config_metadata_ts_ = val.count();
+    }
+    inline int64_t get_metadata_version()
+    {
+        return this->config_metadata_ts_;
+    }
     void push(boost::shared_ptr<Frame> frame);
     void push(const std::string& plugin_name, boost::shared_ptr<Frame> frame);
 
@@ -242,6 +254,8 @@ private:
     boost::mutex mutex_;
     /** process_frame performance stats */
     CallDuration process_duration_;
+    /** time-stamp to represent the version of config metadata */
+    int64_t config_metadata_ts_;
 };
 
 } /* namespace FrameProcessor */
--- a/cpp/frameProcessor/include/FrameProcessorPlugin.h
+++ b/cpp/frameProcessor/include/FrameProcessorPlugin.h
@@ -74,29 +74,29 @@ public:
     void remove_all_callbacks();
     void notify_end_of_acquisition();
     /** return the latest ts of config/status metadata for this plugin instance */
-    __attribute__((always_inline)) int64_t get_config_metadata_ts() const
+    __attribute__((always_inline)) int64_t get_config_ts() const
     {
-        return this->config_metadata_ts_;
+        return this->config_ts_;
     }
-    __attribute__((always_inline)) int64_t get_status_metadata_ts() const
+    __attribute__((always_inline)) int64_t get_status_ts() const
     {
-        return this->status_metadata_ts_;
+        return this->status_ts_;
     }
 
 protected:
     /** function to update the config time-stamp! */
-    __attribute__((always_inline)) void update_config_metadata_version(void)
+    __attribute__((always_inline)) void update_config_ts(void)
     {
         auto time_stamp_ms = std::chrono::time_point_cast<std::chrono::microseconds>(std::chrono::steady_clock::now());
         auto val = std::chrono::duration_cast<std::chrono::microseconds>(time_stamp_ms.time_since_epoch());
-        this->config_metadata_ts_ = val.count();
+        this->config_ts_ = val.count();
     }
     /** function to update the status time-stamp! */
-    __attribute__((always_inline)) void update_status_metadata_version(void)
+    __attribute__((always_inline)) void update_status_ts(void)
     {
         auto time_stamp_ms = std::chrono::time_point_cast<std::chrono::microseconds>(std::chrono::steady_clock::now());
         auto val = std::chrono::duration_cast<std::chrono::microseconds>(time_stamp_ms.time_since_epoch());
-        this->status_metadata_ts_ = val.count();
+        this->status_ts_ = val.count();
     }
     void push(boost::shared_ptr<Frame> frame);
     void push(const std::string& plugin_name, boost::shared_ptr<Frame> frame);
@@ -265,10 +265,10 @@ private:
     boost::mutex mutex_;
     /** process_frame performance stats */
     CallDuration process_duration_;
-    /** time-stamp to represent the version of config metadata */
-    int64_t config_metadata_ts_;
-    /** time-stamp to represent the version of status metadata */
-    int64_t status_metadata_ts_;
+    /** time-stamp to represent the version of config structure */
+    int64_t config_ts_;
+    /** time-stamp to represent the version of status structure */
+    int64_t status_ts_;
 };
 
 } /* namespace FrameProcessor */
--- a/cpp/frameProcessor/include/FrameProcessorPlugin.h
+++ b/cpp/frameProcessor/include/FrameProcessorPlugin.h
@@ -83,9 +83,20 @@ protected:
         auto val = std::chrono::duration_cast<std::chrono::milliseconds>(time_stamp_ms.time_since_epoch());
         this->config_metadata_ts_ = val.count();
     }
-    inline int64_t get_metadata_version()
+    /** function to update the status time-stamp! */
+    inline void update_status_metadata_version(void)
+    {
+        auto time_stamp_ms = std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::system_clock::now());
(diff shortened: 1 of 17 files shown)


cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -83,10 +83,21 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
         auto val = std::chrono::duration_cast<std::chrono::milliseconds>(time_stamp_ms.time_since_epoch());
         this->config_metadata_ts_ = val.count();
     }
-    inline int64_t get_metadata_version()
+    /** function to update the status time-stamp! */
+    inline void update_status_metadata_version(void)
+    {
+        auto time_stamp_ms = std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::system_clock::now());
```
JamesOHeaDLS: system_clock can change but steady_clock is monotonic

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -256,6 +267,8 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
     CallDuration process_duration_;
     /** time-stamp to represent the version of config metadata */
     int64_t config_metadata_ts_;
+    /** time-stamp to represent the version of config metadata */
```
JamesOHeaDLS: Copy-paste typo in comment - should be status metadata

cpp/frameProcessor/include/FrameProcessorPlugin.h
```diff
@@ -73,8 +73,31 @@ class FrameProcessorPlugin : public IFrameCallback, public OdinData::IVersionedO
     void remove_callback(const std::string& name);
     void remove_all_callbacks();
     void notify_end_of_acquisition();
+    static constexpr char METADATA_VERSION[] = "metadata_version";
```
JamesOHeaDLS: I suggest that the constant name be METADATA_VERSION_KEY to better represent its purpose.

cpp/frameProcessor/src/FrameProcessorController.cpp
```diff
@@ -321,6 +316,20 @@ void FrameProcessorController::provideStatus(OdinData::IpcMessage& reply, bool m
...
+    for (auto iter : plugins_) {
+        try {
+            latest_metadata_ver = std::max(
+                latest_metadata_ver,
+                reply.get_param<int64_t>(iter.second->get_name() + '/' + FrameProcessorPlugin::METADATA_VERSION)
+            );
+        } catch (...) {
+            // pass
+        }
+    }
```
ajgdls: When I test this it leads to an exception being thrown and no configuration request replies.  I assume this means that there is a plugin that has not replied with a timestamp for structural changes, but that should not be possible.
I'm pretty certain that we want the parent constructor for all plugins to create the first timestamp, and hopefully we can automatically fill the plugin response with this default timestamp if no further changes have taken place.

This would mean that all existing plugins would "just work" without any change required.

ajgdls (changes requested):
General comment, to make it clear to any future developers what is taking place could we update all methods / variables with "timestamp" in place of "version"?
We are using the timestamp as the hash, but I think the use of version might confuse developers as we also have source code versions for the plugins etc.

cpp/frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -494,6 +497,7 @@ void FileWriterPlugin::requestConfiguration(OdinData::IpcMessage& reply)
             }
         }
     }
+    reply.set_param(this->get_name() + '/' + FrameProcessorPlugin::METADATA_VERSION, get_config_metadata_version());
```
JamesOHeaDLS: Some inconsistency between the plugins as to whether they use this->get_name() or just get_name()

For now, maybe just make the additional line consistent with how it is used within the rest of the file

cpp/frameProcessor/src/FrameProcessorController.cpp
```diff
@@ -321,6 +316,20 @@ void FrameProcessorController::provideStatus(OdinData::IpcMessage& reply, bool m
...
+    reply.set_param(FrameProcessorController::METADATA_HASH, latest_metadata_ver);
```
JamesOHeaDLS: I think the key name should be consistent - eg use _VERSION(_KEY) rather than _HASH

cpp/frameProcessor/src/FrameProcessorController.cpp
```diff
@@ -321,6 +316,20 @@ void FrameProcessorController::provideStatus(OdinData::IpcMessage& reply, bool m
         std::vector<std::string> plugin_warnings = iter->second->get_warnings();
         warning_messages.insert(warning_messages.end(), plugin_warnings.begin(), plugin_warnings.end());
     }
+
+    int64_t latest_metadata_ver = -1;
```
JamesOHeaDLS: This loop is used here and also below.  This version is only different because of the try-catch clause but that doesn't seem to be used here anyway.  Should be made into a function

cpp/frameProcessor/src/KafkaProducerPlugin.cpp
```diff
@@ -126,6 +128,7 @@ void KafkaProducerPlugin::requestConfiguration(OdinData::IpcMessage& reply)
     reply.set_param(get_name() + '/' + KafkaProducerPlugin::CONFIG_PARTITION, this->partition_);
     reply.set_param(get_name() + '/' + KafkaProducerPlugin::CONFIG_DATASET, this->dataset_name_);
     reply.set_param(get_name() + '/' + KafkaProducerPlugin::CONFIG_INCLUDE_PARAMETERS, this->include_parameters_);
+    reply.set_param(this->get_name() + '/' + FrameProcessorPlugin::METADATA_VERSION, get_config_metadata_version());
```
JamesOHeaDLS: Again regarding consistency of this-> use

cpp/frameProcessor/src/KafkaProducerPlugin.cpp
```diff
@@ -143,6 +146,7 @@ void KafkaProducerPlugin::status(OdinData::IpcMessage& status)
     status.set_param(get_name() + '/' + STATUS_FRAMES_LOST, frames_lost_);
     /* Number of acknowledged frames */
     status.set_param(get_name() + '/' + STATUS_FRAMES_ACK, frames_ack_);
+    status.set_param(this->get_name() + '/' + FrameProcessorPlugin::METADATA_VERSION, get_status_metadata_version());
```
JamesOHeaDLS: Again regarding consistency of this-> use

cpp/frameProcessor/src/LiveViewPlugin.cpp
```diff
@@ -197,6 +198,7 @@ void LiveViewPlugin::requestConfiguration(OdinData::IpcMessage& reply)
     reply.set_param(get_name() + '/' + LiveViewPlugin::CONFIG_SOCKET_ADDR, image_view_socket_addr_);
     reply.set_param(get_name() + '/' + LiveViewPlugin::CONFIG_PER_SECOND, per_second_);
     reply.set_param(get_name() + '/' + LiveViewPlugin::CONFIG_DATASET_NAME, dataset_names_);
+    reply.set_param(this->get_name() + '/' + FrameProcessorPlugin::METADATA_VERSION, get_config_metadata_version());
```
JamesOHeaDLS: Again regarding consistency of this-> use

cpp/frameProcessor/src/OffsetAdjustmentPlugin.cpp
```diff
@@ -75,6 +76,7 @@ void OffsetAdjustmentPlugin::configure(OdinData::IpcMessage& config, OdinData::I
 void OffsetAdjustmentPlugin::requestConfiguration(OdinData::IpcMessage& reply)
 {
     reply.set_param(get_name() + '/' + OFFSET_ADJUSTMENT_CONFIG, offset_adjustment_.load());
+    reply.set_param(this->get_name() + '/' + FrameProcessorPlugin::METADATA_VERSION, get_config_metadata_version());
```
JamesOHeaDLS: Again regarding consistency of this-> use

JamesOHeaDLS (changes requested):
Are there tests to go with this new approach?

cpp/frameProcessor/src/FrameProcessorController.cpp
Ohisemega: You are right; the error was that I didn't wrap the `get_param<>()` request with a try-catch block. 
But the Parent Constructor that instantiates the first timestamp doesn't automatically make the implementation work. The child plugins have to append it to their `requestConfiguration()` and `status()` responses for the controller to be able to fetch it. An exception will still be thrown if the controller cannot find it in the `IpcMessage` response.

cpp/frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -168,6 +168,7 @@ FileWriterPlugin::FileWriterPlugin() :
     add_status_param_metadata(prefix + STATUS_LAST_CLOSE, PMDD::UINT_T, PMDA::READ_ONLY, 0, PMD::MAX_UNSET);
     add_status_param_metadata(prefix + STATUS_MAX_CLOSE, PMDD::UINT_T, PMDA::READ_ONLY, 0, PMD::MAX_UNSET);
     add_status_param_metadata(prefix + STATUS_MEAN_CLOSE, PMDD::UINT_T, PMDA::READ_ONLY, 0, PMD::MAX_UNSET);
+    update_status_metadata_version();
```
ajgdls: Have you called this here because you are worried about a race condition from constructor ordering?

cpp/frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -366,6 +367,7 @@ void FileWriterPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMess
                     this->configure_dataset(dataset_name, dsetConfig, reply);
                 }
             }
+            update_config_metadata_version();
```
ajgdls: This should also be added for the delete_datasets case as that can alter the structure.

ajgdls (changes requested):

cpp/frameProcessor/src/FileWriterPlugin.cpp
Ohisemega: It was a typo I didn't fix; I will modify it now

cpp/frameProcessor/src/FileWriterPlugin.cpp
Ohisemega: Done

cpp/frameProcessor/src/FileWriterPlugin.cpp
```diff
@@ -350,6 +350,7 @@ void FileWriterPlugin::configure(OdinData::IpcMessage& config, OdinData::IpcMess
                 LOG4CXX_INFO(logger_, "Dataset name " << dataset_name << " found, creating...");
                 // If we can retrieve a single string parameter then we are being asked to create
                 // a new dataset.  Only create it if it doesn't already exist.
+                std::cout << "Creating new dataset\n";
```
ajgdls: This line needs removing.

ajgdls (changes requested):

cpp/frameProcessor/src/FileWriterPlugin.cpp
Ohisemega: Oops, I'm always embarrassed when I leave logging info in a PR. :(

ajgdls (approved):

cpp/frameProcessor/test/FileWriterPluginTest.cpp
```diff
@@ -151,4 +151,18 @@ BOOST_AUTO_TEST_CASE(FileWriterPluginZeroFrames)
...
+{
+    FrameProcessor::FileWriterPlugin fwp;
+    int64_t first_ts = fwp.get_config_metadata_ts();
+    OdinData::IpcMessage cfg;
+    OdinData::IpcMessage reply;
+    fwp.set_name("hdf");
+    cfg.set_param<std::string>(FrameProcessor::FileWriterPlugin::CONFIG_DATASET, "new_dataset");
+    fwp.configure(cfg, reply);
+    int64_t second_ts = fwp.get_config_metadata_ts();
+    std::cout << "The first is: " << first_ts << " the second is: " << second_ts << '\n';
```
ajgdls: Oops sorry.  One final std::cout I noticed.

ajgdls (changes requested):

cpp/frameProcessor/test/FileWriterPluginTest.cpp
Ohisemega: Sorry, just fixed that as well.

cpp/frameProcessor/src/FrameProcessorController.cpp
```diff
@@ -43,7 +43,7 @@ const std::string FrameProcessorController::CONFIG_VALUE = "value";
 
 const std::string FrameProcessorController::COMMAND_KEY = "command";
 const std::string FrameProcessorController::SUPPORTED_KEY = "supported";
-const std::string FrameProcessorController::METADATA_HASH = "metadata_hash";
+const std::string FrameProcessorController::METADATA_TS_KEY = "metadata_ts";
```
ajgdls: I think Tim commented on this also.  The name should be updated as it is not just the metadata timestamp, but a structural change timestamp.

Maybe STRUCTURAL_TS_KEY = "structural_ts" although I'm not sure that is a great description.  @timcnicholls what did you suggest for this?

Ohisemega: Maybe call it  `LATEST_STRUCT_TS_KEY = "latest_struct_ts"`?

timcnicholls: I was thinking of individual keys for status and config, e.g. `status_ts`, `config_ts`. It's the same key for both but I would favour the clarity of being explicit over a more abstractly named key that is common to status and config.

I note also in the `provideStatus` and `requestConiguration` methods the calls are still filling a local variable `latest_metadata_ver`; maybe this should be `latest_ts` for clarity?

Ohisemega: OK, fixed as required in the latest commit.

timcnicholls (approved):

JamesOHeaDLS (approved):
Looks good, thanks @Ohisemega

ajgdls (approved):

PR #537, opened 2026-07-21
Build on Rockylinux:9 to match RHEL 9 Kernel. 

MahirAbbas:
This builds odin on `rockylinux:9` to match `RHEL 9` kernel. it also pins `odin-control` to `1.6.2`. It is likely we should be able to start compiling with gcc 15 once `hdf5filters` compilation bugs are fixed and upstreamed.

Reasoning: Low level performance analysis tools (such as `perf` or `eBPF`, which are kernel level tools used to inspect the performance of applications) are built against the kernel. Version mismatches lead either to degraded capabilities or to downright inability to run/collect metrics due to ABI differences. `rockylinux` is a FOSS linux distro with 100% RHEL binary compatibility.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,12 @@
-# Shared setup stage ###########################################################
-FROM ubuntu:24.04 AS common
+FROM rockylinux:9  AS common
 
 ENV PATH=/odin/bin:/odin/scripts:/venv/bin:$PATH
 
 # Get fundamental packages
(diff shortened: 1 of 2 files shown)


MahirAbbas:
Hdf5 now builds on x86 and Aarch64

Ohisemega:
We have figured out how to install `perf` within the `Ubuntu24` container image. So this PR will be closed.

PR #540, opened 2026-07-23
Mount Build Directory as Volume

MahirAbbas:
This creates build dependency issues when building on macOS due to differences in `APFS` and the Linux File System. Mounting as a volume allows Docker to manage the mount, resolving the build issues. Tested on both a Linux workstation and on Mac

### DIFF
--- a/.devcontainer/devcontainer-lock.json
+++ b/.devcontainer/devcontainer-lock.json
@@ -1,0 +1,9 @@
+{
+  "features": {
+    "ghcr.io/devcontainers/features/common-utils:2": {
+      "version": "2.5.9",
+      "resolved": "ghcr.io/devcontainers/features/common-utils@sha256:cb0c4d3c276f157eed17935747e364178d75fee17f55c4e129966f64633deb3a",
+      "integrity": "sha256:cb0c4d3c276f157eed17935747e364178d75fee17f55c4e129966f64633deb3a"
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


Ohisemega:
Has this been tested on a Linux host, @MahirAbbas?
 Our current runtime platform systems are all Linux, so we do need to ensure we aren't building for a specific platform that odin-data isn't running on.
 A Mac build is currently not yet in the pipeline (an AARCH-64 build is, however, in the roadmap to enable cross-compilation on Mac). But cross-compilation should be systematic, with extensive conditional checks in the build system for each supported platform.
 
 For future reference, any changes in the build system should be tested on a Linux host primarily; a change which doesn't meet the Linux OS host requirements would be difficult to integrate.

MahirAbbas:
> Has this been tested on a Linux host, @MahirAbbas? Our current runtime platform systems are all Linux, so we do need to ensure we aren't building for a specific platform that odin-data isn't running on. A Mac build is currently not yet in the pipeline (an AARCH-64 build is, however, in the roadmap to enable cross-compilation on Mac). But cross-compilation should be systematic, with extensive conditional checks in the build system for each supported platform.

Yes this has been tested and is working on an x86 Linux system

Ohisemega (approved):

Ohisemega:
> > Has this been tested on a Linux host, @MahirAbbas? Our current runtime platform systems are all Linux, so we do need to ensure we aren't building for a specific platform that odin-data isn't running on. A Mac build is currently not yet in the pipeline (an AARCH-64 build is, however, in the roadmap to enable cross-compilation on Mac). But cross-compilation should be systematic, with extensive conditional checks in the build system for each supported platform.
> 
> Yes this has been tested and is working on an x86 Linux system

OK, that's great; it seems the workspace is being mounted as a volume mount. This is OK for me.
cc: @JamesOHeaDLS, @ajgdls

PR #541, opened 2026-07-23
Add Optimization Flag

### LINKED ISSUE #542: Add Optimization Flag to the CMake Build System
Add the `-O2` flag to the CMAKE build system.
Compile odin-data and ensure the changes do not break unit tests or integration tests!

MahirAbbas:
~Creates a common_flags library which has compile options. The current flag is now -O2 and -g.~

Adds optimisation flag to `Dockerfile`

Fixes #542

### DIFF
--- a/.devcontainer/devcontainer-lock.json
+++ b/.devcontainer/devcontainer-lock.json
@@ -1,0 +1,9 @@
+{
+  "features": {
+    "ghcr.io/devcontainers/features/common-utils:2": {
+      "version": "2.5.9",
+      "resolved": "ghcr.io/devcontainers/features/common-utils@sha256:cb0c4d3c276f157eed17935747e364178d75fee17f55c4e129966f64633deb3a",
+      "integrity": "sha256:cb0c4d3c276f157eed17935747e364178d75fee17f55c4e129966f64633deb3a"
+    }
+  }
+}
--- a/.devcontainer/devcontainer-lock.json
+++ b/.devcontainer/devcontainer-lock.json
@@ -1,9 +1,0 @@
-{
-  "features": {
-    "ghcr.io/devcontainers/features/common-utils:2": {
-      "version": "2.5.9",
-      "resolved": "ghcr.io/devcontainers/features/common-utils@sha256:cb0c4d3c276f157eed17935747e364178d75fee17f55c4e129966f64633deb3a",
-      "integrity": "sha256:cb0c4d3c276f157eed17935747e364178d75fee17f55c4e129966f64633deb3a"
-    }
-  }
-}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -46,6 +46,7 @@
     "workspaceMount": "source=${localWorkspaceFolder}/..,target=/workspaces,type=bind",
     "mounts": [
         // Mount /dev/shm for odin-data applications to use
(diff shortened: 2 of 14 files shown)


Ohisemega:
This looks OK to me. Are you OK with adding the flag @ajgdls?
cc: @JamesOHeaDLS

Ohisemega (approved):

Ohisemega:
The response from the architecture team is to benchmark this PR in the profiling suite before committing the PR. So the proposed change is good; we would like justification by analysis and numbers.

MahirAbbas:
Initial data suggests performance improvements are between 6.7% and 18.2%, with Sum plugin showing a 46% decrease in processing time. @Ohisemega

MahirAbbas:
The current build is now with `-O2` and with `-g` to preserve debug symbols

Ohisemega:
Based on my comments in PR #544, it might be worth using this method to add the optimisation flag instead:
`set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O2")`
The concern is that `interface` doesn't seem to propagate the flags from the top `CMakeLists.txt` file!

Ohisemega (changes requested):
See last comment about changing from using `interface` to: 
`set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O2")`

MahirAbbas:
@Ohisemega Upon further research and thought, it makes more sense to use the compilation flag inside the Dockerfile. This also leads to the compilation flag being propagated down.

Ohisemega:
> @Ohisemega Upon further research and thought, it makes more sense to use the compilation flag inside the Dockerfile. This also leads to the compilation flag being propagated down.

Hi @MahirAbbas, the changes you made will affect the build container, which is good, but we also need those changes in the `developer` stage of the `Dockerfile`. Since the `Dockerfile`'s developer stage doesn't build the binary, we need to modify the `CMake` files as well.

MahirAbbas:
> > @Ohisemega Upon further research and thought, it makes more sense to use the compilation flag inside the Dockerfile. This also leads to the compilation flag being propagated down.
> 
> Hi @MahirAbbas, the changes you made will affect the build container, which is good, but we also need those changes in the `developer` stage of the `Dockerfile`. Since the `Dockerfile`'s developer stage doesn't build the binary, we need to modify the `CMake` files as well.

@Ohisemega Is the `developer` stage used for anything but development? The optimisation will increase compile time (which I was assuming would be quite annoying when developing). It made sense to me to ship an optimised image, and leave the `developer` experience unoptimised, but I may have been wrong.

Ohisemega:
> > > @Ohisemega Upon further research and thought, it makes more sense to use the compilation flag inside the Dockerfile. This also leads to the compilation flag being propagated down.
> > 
> > 
> > Hi @MahirAbbas, the changes you made will affect the build container, which is good, but we also need those changes in the `developer` stage of the `Dockerfile`. Since the `Dockerfile`'s developer stage doesn't build the binary, we need to modify the `CMake` files as well.
> 
> @Ohisemega Is the `developer` stage used for anything but development? The optimisation will increase compile time (which I was assuming would be quite annoying when developing). It made sense to me to ship an optimised image, and leave the `developer` experience unoptimised, but I may have been wrong.

It's a fair point you have raised. I'll approve it then!

Ohisemega (approved):

PR #545, opened 2026-08-07
Build System Cleanup

### LINKED ISSUE #532: Build System Clean up!
The `CMake` build system of Odin-data needs some cleanup. Some files which have likely been renamed or deleted still have references in the code comments and the build system. Search the build system for files referenced by the CMake instructions which no longer have references in the project and remove them. Also delete/update comments in source or header files that refer to those deprecated classes/files.
As an example, in `cpp/frameProcessor/src/CMakeLists.txt`, we have these lines:
```
file(GLOB APP_SOURCES FrameProcessorController.cpp
                      SharedMemoryController.cpp
                      SharedMemoryParser.cpp
                      SocketHandler.cpp )
```
However, the `SharedMemoryParser.cpp` and `SocketHandler.cpp` files no longer exist in the project. A global search for `SharedMemoryParser` brings up more references. These references should be removed and the relevant comments updated or removed.

MahirAbbas:
Replaces `file(GLOB..` with `target_sources(..`
GLOB fails silently if the files don’t exist, while `target_sources` fails with an error.

Fixes #532


Ohisemega (approved):

ajgdls (changes requested):
This is a good improvement, but I think to fully fix the issue #532 some extra references need to be removed from comments in the source files.

MahirAbbas:
> This is a good improvement, but I think to fully fix the issue #532 some extra references need to be removed from comments in the source files.

I've naively removed all references to `SharedMemoryParser`. Is this what you meant @ajgdls ?

Ohisemega:
~@MahirAbbas, there's a significant change in the codecoverage/project test. Could you look into this?
cc@ @ajgdls~
Edit: The reason is that a rogue `.coverage` file was added to the commit.

PR #560, opened 2026-09-04
Fix IpcTornadoClient Uncaught Exception Bug

### LINKED ISSUE #559: Fix IpcTornadoClient Bug
There is an unhandled case in the IpcTornadoClient class:
When no plugins are loaded in the FrameProcessor app, the class will throw an exception trace call because it tries to fetch the plugin names `plugin_names : list = params["plugins"]["names"]`.
Protect this method against cases where no plugins are loaded.

Ohisemega:
Protect `IpcTornadoClient`'s: `_update_status()` and `_update_configuration()` methods from triggering uncaught exceptions when no Plugins are loaded.

Fixes #559

### DIFF
--- a/python/src/odin_data/control/ipc_tornado_client.py
+++ b/python/src/odin_data/control/ipc_tornado_client.py
@@ -180,13 +180,13 @@ class IpcTornadoClient(object):
         :param config_msg: Incoming configuration message response
         """
         params = config_msg[self.PARAMS_KEY]
-        
-        plugin_names : list = params["plugins"]["names"]
         self._parameters[self.IPC_VAL_CONFIG] = {}
-        self._parameters[self.IPC_VAL_CONFIG][self.CONFIG_PARAMS_KEY]: dict = {}
-        for name in plugin_names:
-            self._parameters[self.IPC_VAL_CONFIG][self.CONFIG_PARAMS_KEY][name] = params[name]
-            params.pop(name, None)
+        if("plugins" in params and "names" in params["plugins"]):
+            plugin_names : list = params["plugins"]["names"]
+            self._parameters[self.IPC_VAL_CONFIG][self.CONFIG_PARAMS_KEY]: dict = {}
+            for name in plugin_names:
+                self._parameters[self.IPC_VAL_CONFIG][self.CONFIG_PARAMS_KEY][name] = params[name]
+                params.pop(name, None)
         params.pop("plugins", None)
 
         if(self.IPC_VAL_CONFIG_TS in params):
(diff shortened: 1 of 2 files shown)


JamesOHeaDLS:
Thanks @Ohisemega this looks good.

I think it would be worth having a test for _update_configuration where different config_msgs can be fed in to capture the different scenarios

Ohisemega:
> Thanks @Ohisemega this looks good.
> 
> I think it would be worth having a test for _update_configuration where different config_msgs can be fed in to capture the different scenarios

OK, just seeing this; will do!
Edit: @JamesOHeaDLS: it is ready for review/approval.

python/tests/test_ipc_tornado_client.py
```diff
@@ -57,33 +81,34 @@ def test_ipc_tornado_client(self, client, tornado_channel):
...
         client.send_request("request_commands")
         client._callback([reply_str])
         assert client.parameters["commands"] == {"test_command_item": "Test3"}
 
-        reply_str = '{"id": 4, "msg_type": "nack", "msg_val": "configure", "params": {"test_config_fail": "Test4", "error": "test error"}}'
+        reply_str = '{"id": 6, "msg_type": "nack", "msg_val": "configure", "params": {"test_config_fail": "Test4", "error": "test error"}}'
         client.send_configuration({"item1": "value1"})
         client._callback([reply_str])
         rejected = client.read_rejected_configs()
+        print(rejected)
```
ajgdls: Can you remove this print statement please.

ajgdls (changes requested):
Just one minor modification needed please.

python/tests/test_ipc_tornado_client.py
Ohisemega: Thank you, done!

python/src/odin_data/control/ipc_tornado_client.py
```diff
@@ -202,11 +202,12 @@ def _update_status(self, status_msg):
...
-        self._parameters[self.IPC_VAL_STATUS][self.STATUS_PARAMS_KEY]["plugins"] = params["plugins"]
-        for name in plugin_names:
-            self._parameters[self.IPC_VAL_STATUS][self.STATUS_PARAMS_KEY][name] = params[name]
-            params.pop(name, None)
+        if("plugins" in params and "names" in params["plugins"]):
+            plugin_names : list = params["plugins"]["names"]
+            self._parameters[self.IPC_VAL_STATUS][self.STATUS_PARAMS_KEY]["plugins"] = params["plugins"]
+            for name in plugin_names:
+                self._parameters[self.IPC_VAL_STATUS][self.STATUS_PARAMS_KEY][name] = params[name]
+        params.pop("plugins", None)
```
JamesOHeaDLS, line 211: This for loop has lost `params.pop(name, None)` - is that intentional?

Ohisemega: Yes, that's intentional. If we receive a message with none/empty `params`, we want to remove whatever was previously in the IpcTornadoClient's dictionary.

ajgdls (approved):

JamesOHeaDLS (approved):

Issue #517, opened 2026-06-10
Odin Data Roadmap


ajgdls:
- Release 1.13.0. DLS
- Benchmarking (currently ends 1st September 2026.  Possible 3-month extension). DLS
- For release 2.0.0 (merge all PRs onto dev2.0 branch):

> - Pull out the shared memory controller into a plugin in the FP C++ application. 3 months
> - Implement Frame Receiver plugin (re-use as much of the existing Frame Receiver codebase to allow current decoders to work without change)
> - Eventually remove Frame Receiver as a separate application.
> - Implement Frame Simulator as a plugin (for FP).

- For point relase (1.14+ or 2.0+):

> - Abstract all rapid JSON out of the existing FP classes (mostly wrapped in `IpcMessage` but have leaked out into other classes).
> - Once rapid JSON is abstracted, investigate SimdJSON as a replacement.
> - Thread pinning
> - Investigation into frame worker queue (once performance and benchmarking are complete)
> - Documentation improvements
> - Increase C++ testing coverage to 80%
> - Devcontainer image build for Arm 

> - Just some note about creating branches that align with git-flow for future development

PR #564, opened 2026-09-08
WIP: Add SharedMemoryPlugin Implementation

### LINKED ISSUE #565: Convert SharedMemoryController Module into a Plugin in the FP
We want to transform the `SharedMemoryController` module into a plugin which can be loaded/unloaded at runtime by developers or users.

Ohisemega:
Fixes #565 

This is in-view of odin-data release 2.0. See Issue #517 (Roadmap) for details.


cpp/frameProcessor/src/SharedMemoryPlugin.cpp
```diff
@@ -0,0 +1,144 @@
...
+}
+
+void SharedMemoryPlugin::status(OdinData::IpcMessage& reply)
+{
+    if (shmctrlr_handle_) {
+        reply.set_param(this->get_name() + '/' + SharedMemoryPlugin::STATUS_SHB_NAME, shmctrlr_handle_->getName());
+        reply.set_param(
+            this->get_name() + '/' + SharedMemoryPlugin::STATUS_SHB_CONFIGURED, shmctrlr_handle_->isConfigured()
+        );
+    } else {
```
Ohisemega, line 117: Hi @ajgdls,
I was thinking about this implementation, and I suggest that if there is no existing `SharedMemoryController` object in the plugin, then we should not return any status information.
There are 3 possible states for the controller object:

- There is no controller object.
- There is a controller, but it is not configured.
- There is a controller, and it is configured.

A status-response method would care only about states 2 & 3, and the if condition covers both.

Ohisemega: The `else-condition` will return `false` for the `shared_buf_configured` status parameter if there is no existing controller object. And that could be misleading; hence, I suggest we return nothing.

ajgdls: OK I see you point on this, I agree with your suggestion.

cpp/frameProcessor/include/SharedMemoryPlugin.h
```diff
@@ -0,0 +1,59 @@
...
+    /** IpcReactor thread */
+    boost::thread m_thread_;
+    /** The shared memory controller object */
+    boost::shared_ptr<SharedMemoryController> shmctrlr_handle_;
+    /** Pointer to logger */
+    LoggerPtr logger_;
+
+    static void dummy_timer()
+    {
+    }
```
ajgdls: I think we should consider renaming this as dummy timer suggests the implementation is not fully thought out.  We can discuss.

cpp/frameProcessor/src/FrameProcessorController.cpp
```diff
@@ -403,8 +394,9 @@ void FrameProcessorController::configure(OdinData::IpcMessage& config, OdinData:
     // Check for a request to inject an End Of Acquisition object
     if (config.has_param(FrameProcessorController::CONFIG_EOA)) {
         LOG4CXX_DEBUG_LEVEL(1, logger_, "Injecting End Of Acquisition object into plugin chain");
-        if (sharedMemController_) {
-            sharedMemController_->injectEOA();
+        std::string plugin_name = config.get_param<std::string>(FrameProcessorController::CONFIG_EOA);
+        if (plugins_.count(plugin_name)) {
```
ajgdls: This changes the client API, as injecting an EOA frame now requires the name of the plugin to inject into, where as before it was always injected into the shared memory controller.   This probably is not used by many detectors but worth noting.

ajgdls (commented):
Some points of discussion before we fully approve.  We should probably check through the source code documentation.

cpp/frameProcessor/include/SharedMemoryPlugin.h
timcnicholls: As discussed at meeting, timer is necessary to allow the reactor to tick and terminate cleanly, so this should be named more clearly.

cpp/frameProcessor/src/FrameProcessorController.cpp
timcnicholls: As discussed - let's look at putting EOA back in to core application, with it sent to all plugins that are head of chain.

cpp/frameProcessor/src/SharedMemoryPlugin.cpp
```diff
@@ -0,0 +1,145 @@
...
+ * which manage the receipt of frame ready notifications and construction of
+ * Frame objects from shared memory.
+ * Pointers to the two objects are kept by this class.
+ *
+ * \param[in] sharedMemName - Name of the shared memory block opened by the frame receiver.
+ * \param[in] frPublisherString - Endpoint for sending frame release notifications.
+ * \param[in] frSubscriberString - Endpoint for receiving frame ready notifications.
+ */
+void SharedMemoryPlugin::setupFrameReceiverInterface(
+    const std::string& frPublisherString,
```
timcnicholls: References to "publisher" and "subscriber" for IPC channels should be renamed to "release" and "ready" for consistency with the meaning of the channel, rather than the type.

cpp/frameProcessor/src/SharedMemoryPlugin.cpp
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * @param reply - Response IpcMessage.
+ */
+void SharedMemoryPlugin::requestConfiguration(OdinData::IpcMessage& reply)
+{
+    reply.set_param(this->get_name() + '/' + SharedMemoryPlugin::CONFIG_FR_READY, this->frReadyEndpoint_);
+    reply.set_param(this->get_name() + '/' + SharedMemoryPlugin::CONFIG_FR_RELEASE, this->frReleaseEndpoint_);
+}
+
+void SharedMemoryPlugin::status(OdinData::IpcMessage& reply)
```
timcnicholls: Be good to add some statistics reporting for e.g. frames received

timcnicholls (commented):
Looks good. Some comments as per discussion in meeting. Let's also target this PR at a dev_2.0 branch, as suggested in the 2.0 roadmap issue #517

